PR #1, opened 2019-12-04
Add the initial metrics API definition

vmarkovtsev:

### DIFF
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -1,0 +1,196 @@
+openapi: 3.0.0
+info:
+  version: 1.0.0
+  title: Athenian Dashboard API
+  license:
+    name: Proprietary until Eiso decides on licensing
+servers:
+  - url: https://api.athenian.co/v1
+paths:
+  /metrics:
+    post:
+      summary: Calculate metrics.
+      operationId: calcMetrics
+      requestBody:
+        description: Desired metric definitions.
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/MetricsRequest"
+      responses:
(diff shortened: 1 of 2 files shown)


spec/openapi.yaml
```diff
@@ -0,0 +1,108 @@
+openapi: 3.0.0
+info:
+  version: 1.0.0
+  title: Athenian Dashboard API
+  license:
+    name: CC-BY-4.0
```
dpordomingo: I think that `CC` is not a proper software license. On the other side, I know that @se7entyse7en already left this as something to be defined later, so he preferred to leave it empty (which use to mean: 'all rights reserved')

vmarkovtsev: This is the license of the API document (that's stated in the spec!), and it was the most popular according to https://matija.suklje.name/popular-licenses-in-openapi
So no worries.

vmarkovtsev:
@dpordomingo Regarding the endpoints: they are listed in this doc.
However, I need to add a few service ones to let you consume `/metrics`.

vmarkovtsev:
BTW: here is the link to the spec that guides me: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md

spec/openapi.yaml
vmarkovtsev: @eiso says: no licenses. So changed to proprietary (default).

se7entyse7en: Is this a real issue? I mean the license of the spec? I just don't know whether there having a CC-BY license for the spec is in conflict with being part of an unlicensed project. \cc @eiso

PR #2, opened 2019-12-05
Add the autogenerated server stub

vmarkovtsev:


vmarkovtsev:
@se7entyse7en As you see, I've chosen aiohttp as the backbone. It is the best option for me to deliver things quickly.

PR #4, opened 2019-12-11
Add the DB connection and CI

vmarkovtsev:
Sorry for a huge pile of code.


.gitignore
```diff
@@ -1,4 +1,7 @@
 .idea
+
+server/tests/db.sqlite
```
kuba--: Why do we need sqlite, if we can run postgres locally?

server/athenian/api/slogging.py
```diff
@@ -0,0 +1,281 @@
...
+    dt = get_datetime_now().astimezone()
+    tzstr = dt.strftime("%z")
+    tzstr = tzstr[:-2] + ":" + tzstr[-2:]
+    return dt.tzinfo, tzstr
+
+
+timezone, tzstr = get_timezone()
+_now = get_datetime_now()
+if _now.month == 12:
+    _fest = "🎅"
```
eiso: Priceless Easter eggs 😍

Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+    apt-get install -y --no-install-suggests --no-install-recommends \
+      apt-utils ca-certificates gnupg2 locales curl python3 python3-distutils && \
+    echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && \
+    locale-gen && \
+    curl -L https://bootstrap.pypa.io/get-pip.py | python3 && \
+    pip3 install --no-cache-dir cython && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Intel MKL
```
kuba--: Do we really need it, right now (Kernel Math Library)?
The same with OpenMP?

server/athenian/api/controllers/default_controller.py
```diff
@@ -1,21 +1,41 @@
...
     body = MetricsRequest.from_dict(body)
-    return web.Response(status=200)
+    github_repos_query = []
+    for for_set in body._for:
+        for repo in for_set.repositories:
+            if repo.startswith("github.com/"):
+                github_repos_query.append(github.Repository.name == repo[len("github.com/"):])
+    github_repos_query = select([github.Repository]).where(sql.or_(*github_repos_query))
+    repos = await request.db.fetch_all(query=github_repos_query)
+    print(repos)
```
kuba--: leftover?

server/athenian/api/controllers/default_controller.py
```diff
@@ -1,21 +1,41 @@
...
+    met.date_from = body.date_from
+    met.date_to = body.date_to
+    met.granularity = body.granularity
+    met.metrics = body.metrics
+    met.calculated = []
+    now = datetime.date.today()
+    for for_set in body._for:
+        met.calculated.append(CalculatedMetric(
+            _for=for_set,
+            values=[CalculatedMetricValues(date=now, values=[0.5] * len(met.metrics))] * 10))
```
kuba--: why 10?, why 0.5?

server/athenian/api/controllers/default_controller.py
```diff
@@ -1,21 +1,41 @@
...
-from athenian.api.models.invalid_request_error import InvalidRequestError
+from athenian.api.models.db import github
+# from athenian.api.models.invalid_request_error import InvalidRequestError
 from athenian.api.models.metrics_request import MetricsRequest
-from athenian.api.models.no_source_data_error import NoSourceDataError
-from athenian.api import util
+# from athenian.api.models.no_source_data_error import NoSourceDataError
 
 
 async def calc_metrics(request: web.Request, body) -> web.Response:
```
kuba--: lets split this function into sth. like `fetch repos`, `collect metrics`

server/athenian/api/models/db/__init__.py
```diff
@@ -0,0 +1,5 @@
+def hack_sqlite_arrays():
```
kuba--: maybe we don't need this if we get rid of sqlite

server/tests/conftest.py
```diff
@@ -1,19 +1,34 @@
...
-import connexion
+from athenian.api import AthenianApp
+from athenian.api.models.db import hack_sqlite_arrays
+from athenian.api.models.db.github import Base
+from .sample_db_data import fill_session
 
 
 @pytest.fixture
-def client(loop, aiohttp_client):
+def client(loop, aiohttp_client, sqlite_db):
```
kuba--: I wold make it possible to test against existing db (mainly postgres) as well.

server/athenian/api/models/db/github/__init__.py
```diff
@@ -0,0 +1,223 @@
...
+Base = declarative_base()
+
+
+class VersionedMixin:
+    versions = Column("versions", ARRAY(Integer()), index=True)
+
+
+class IssueComment(Base, VersionedMixin):
+    __tablename__ = "github_issue_comments_versioned"
+    sum256 = Column("sum256", String(64), primary_key=True)
```
dennwc: Noticed the `sum256` is used frequently. Maybe deserves a `Mixin` as well?

server/athenian/api/models/db/github/__init__.py
```diff
@@ -0,0 +1,223 @@
...
+
+Base = declarative_base()
+
+
+class VersionedMixin:
+    versions = Column("versions", ARRAY(Integer()), index=True)
+
+
+class IssueComment(Base, VersionedMixin):
+    __tablename__ = "github_issue_comments_versioned"
```
dennwc: Is this schema directly from the metadata-retrieval? Or it's something new?

server/athenian/api/slogging.py
dennwc: Log parsers would be very happy :grin:

server/athenian/api/controllers/default_controller.py
```diff
@@ -1,21 +1,41 @@
...
     :param body: Desired metric definitions.
     :type body: dict | bytes
-
     """
     body = MetricsRequest.from_dict(body)
-    return web.Response(status=200)
+    github_repos_query = []
+    for for_set in body._for:
+        for repo in for_set.repositories:
+            if repo.startswith("github.com/"):
```
dennwc: Should it also handle `http(s)://github.com/...`?

.gitignore
se7entyse7en: I agree that for local development we can use postgres, we can just use a simple docker-compose file. On the other hand, for testing it could make sense using sqlite imo.

.travis.yml
```diff
@@ -0,0 +1,21 @@
+language: python
+python: 3.7
+sudo: true
+dist: bionic
```
se7entyse7en: We should agree on this as well to have a default across all the projects. \cc @lwsanty

server/athenian/api/slogging.py
vmarkovtsev: Log parsers should always be happy because we will run structured logging in production :)

.gitignore
vmarkovtsev: This is for the local unit tests, so that they can run without docker and otherwise any running services.

server/athenian/api/slogging.py
```diff
@@ -0,0 +1,281 @@
...
+def get_timezone() -> Tuple[datetime.tzinfo, str]:
+    """Discover the current time zone and it's standard string representation."""
+    dt = get_datetime_now().astimezone()
+    tzstr = dt.strftime("%z")
+    tzstr = tzstr[:-2] + ":" + tzstr[-2:]
+    return dt.tzinfo, tzstr
+
+
+timezone, tzstr = get_timezone()
+_now = get_datetime_now()
```
se7entyse7en: `_fest` is broken if we do not restart for a month since it's set only once.

server/athenian/api/slogging.py
```diff
@@ -0,0 +1,281 @@
+import argparse
```
se7entyse7en: Could you briefly explain me the need for this as it seems a lot of work just for logging. I didn't dig depth into the code.

.travis.yml
```diff
@@ -0,0 +1,21 @@
+language: python
+python: 3.7
+sudo: true
```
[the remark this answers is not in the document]
vmarkovtsev: Travis has ignored `sudo` since the last ~12 months, and it is implicitly `true` even if you set it to `false`.

Proof: https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

As you see, they abandoned the container-based environments so that both `sudo: false` and `sudo: true` builds run with the same speed.

That being said, the current Travis config is not using sudo anyway, so I can turn it off, but that will lead to `pip install --user` in the `install` section. I don't think it's worth it.

.travis.yml
vmarkovtsev: I have chosen the most up-to-date environment here.

Dockerfile
```diff
@@ -0,0 +1,69 @@
+FROM ubuntu:19.10
+
+ENV BROWSER=/browser \
+    LC_ALL=en_US.UTF-8
+
+RUN echo '#!/bin/bash\n\
```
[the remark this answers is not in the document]
vmarkovtsev: Nope, this works fine. This is the special syntax for multi-line strings in Dockerfile: `\n\`. The ending quote is on line 10.

Dockerfile
vmarkovtsev: We are going to depend on pandas, pandas depends on numpy, and I custom-compile numpy to work faster with MKL.
So things like calculating stats should work faster.

There is a pypi package that does not force to do these things: https://pypi.org/project/intel-numpy/
But they don't provide a binary build for 3.7, unfortunately.

.gitignore
kuba--: yeah, I know sounds fine. the only question is how will populate this db

Dockerfile
vmarkovtsev: I can remove these whistles and blows if everybody thinks it's not worth it...

server/athenian/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""Python SDK and supported machine learning based analyzers for lookout."""
```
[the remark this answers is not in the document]
vmarkovtsev: Haha, good catch of copy-paste

server/athenian/api/controllers/default_controller.py
kuba--: but what does it mean?

server/athenian/api/controllers/default_controller.py
vmarkovtsev: I totally agree. This is currently a dummy shitty code to demo that everything works together. Let me start working on real logic in the next PR.

server/athenian/api/models/db/github/__init__.py
kuba--: yeah, looks similar, but we have to find a way to easily update it

server/athenian/api/controllers/default_controller.py
vmarkovtsev: Sorry, I should have left a comment: this is just a stupid demo, no real logic is hidden.

.travis.yml
kuba--: we don't want to run ubuntu (desktop distro) on the server, right?
So maybe it makes sense to have the same distro on travis and in Docker file.

server/athenian/api/models/db/github/__init__.py
vmarkovtsev: Yes, I generated it with `sqlacodegen` from the real Postgres schema initialized by metadata-retrieval.

server/athenian/api/models/db/github/__init__.py
vmarkovtsev: +1

server/tests/conftest.py
vmarkovtsev: +1

vmarkovtsev: https://github.com/athenianco/athenian-api/issues/5

server/athenian/api/models/db/github/__init__.py
vmarkovtsev: https://github.com/athenianco/athenian-api/issues/6

Dockerfile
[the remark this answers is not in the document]
dennwc: Isn't that `'` on line 10 escaped with `\`?

server/tests/conftest.py
se7entyse7en: And this re-opens the issue of opening tickets on JIRA vs GH :p

Dockerfile
se7entyse7en: Since it's already here and it will make numpy go faster I'd just keep it.

.gitignore
vmarkovtsev: I plan to dump from the staging environment once we have it and ORMize it in `sample_db_data.py`

.travis.yml
vmarkovtsev: Python does not play well with any musl-like. It is very hard to install packages, and those that install sometimes don't function well. Ubuntu is a golden standard here.

Regarding `ubuntu (desktop distro)`: c'mon, an Ubuntu docker image is `apt-get` with `coreutils`, there is nothing else.
It's called "core-cloud" nowadays, https://partner-images.canonical.com/core/bionic/current/
As long as you don't `apt-get install` much, you don't install much of bloated crap.

lwsanty: let's use the latest(afaik in travis it is bionic) across all the projects, the dafault one which is xenial seems to be too outdated

please correct me if I'm wrong

Dockerfile
```diff
@@ -0,0 +1,69 @@
+FROM ubuntu:19.10
```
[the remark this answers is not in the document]
lwsanty: btw, do we need to carry whole fat `ubuntu` image as a FROM one?
are there any more light-weight options in our case?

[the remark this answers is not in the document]
vmarkovtsev: `bionic` runs with Python 3.6, unfortunately. I don't want to support two different Python environment versions.

We could use smth similar to pyenv... Or better, pydockenv in Travis.

Dockerfile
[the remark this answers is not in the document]
vmarkovtsev: LOL, indeed I should remove the last backslash.

server/athenian/api/controllers/default_controller.py
vmarkovtsev: Yep, these are just random numbers for now.

The meaning is to return something as the metric values.

server/athenian/api/slogging.py
vmarkovtsev: Added the module docstring.

server/athenian/api/slogging.py
vmarkovtsev: The added module docstring should resolve this question.

server/tests/conftest.py
vmarkovtsev: I heard that they are automatically mirrored, no?

.travis.yml
vmarkovtsev: Yep. Proof: https://docs.travis-ci.com/user/reference/overview/#linux

Dockerfile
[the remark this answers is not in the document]
vmarkovtsev: Compared to 400MB of Intel MKL in one of the layers, 73MB of the root image doesn't play a significant role. The overall image size is 1.8GB atm.

Anyway, this is the most light-weight ubuntu option, yes.

server/athenian/api/__init__.py
```diff
@@ -1,20 +1,115 @@
...
+            db = databases.Database(db_conn, **(db_options or {}))
+            await db.connect()
+            logging.getLogger(__package__).info("Connected to the DB")
+            self.db = db
+            self._db_connected_event.set()
+
+        # Schedule the DB connection
+        asyncio.get_event_loop().call_soon(asyncio.ensure_future, connect_to_db())
+
+    def create_app(self) -> aiohttp.web.Application:
```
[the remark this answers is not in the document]
vmarkovtsev: The default is

```python
return aiohttp.web.Application()
```

server/athenian/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""Athenians are the best."""
```
dennwc: Good one! :smile: No bias. Not at all :)

server/tests/conftest.py
se7entyse7en: Oh if that's the case then that's great, I didn't know whether it was already working. \cc @marnovo

.travis.yml
se7entyse7en: Yup, agree with @lwsanty

Dockerfile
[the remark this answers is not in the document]
se7entyse7en: > Or better, pydockenv in Travis.

hahahah

> I don't want to support two different Python environment versions.

I agree, and it also doesn't make sense for an application. And I'd avoid installing python on bionic on the Dockerfile, unless we encounter a problem later.

server/athenian/api/slogging.py
se7entyse7en: Makes sense, I'll avoid running locally the api for more than a month then :trollface:

PR #8, opened 2019-12-13
add cd

lwsanty:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,21 @@ language: python
 python: 3.7
 sudo: true
 dist: bionic
+
+env:
+  global:
+    - GOOGLE_DOCKER_IMAGE=gcr.io/athenian-1/api
+    - IMAGE=${GOOGLE_DOCKER_IMAGE}:${TRAVIS_BRANCH}-${TRAVIS_COMMIT:0:7}
+
 cache:
-  pip: true
   directories:
-    - $HOME/docker
+    - "$HOME/.cache/pip"
+    - "$HOME/google-cloud-sdk/"
+    - "$HOME/docker"
+
 before_cache:
   - chown -R travis:travis $HOME/.cache/pip
-  - docker save -o $HOME/docker/images.tar athenian/api:latest
+  - docker save -o $HOME/docker/images.tar ${GOOGLE_DOCKER_IMAGE}:latest
 
 matrix:
   fast_finish: true
@@ -22,13 +30,20 @@ matrix:
       script:
         - flake8
         - pytest
-    - before_install:
+    - if: branch != master
+      before_install:
         - docker load -i $HOME/docker/images.tar || true
         - docker image list
       install:
-        - docker build --cache-from athenian/api -t athenian/api .
+        - docker build --cache-from ${GOOGLE_DOCKER_IMAGE} -t ${GOOGLE_DOCKER_IMAGE} .
       script:
-        - docker run --rm athenian/api --help
+        - docker run --rm ${GOOGLE_DOCKER_IMAGE} --help
+    - stage: deploy
+      if: branch = master AND type != pull_request
+      script:
+        - docker load -i $HOME/docker/images.tar || true
+        - docker tag ${GOOGLE_DOCKER_IMAGE} ${IMAGE}
+        - make docker-push
 
 notifications:
   email: false
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,7 @@
+CI_REPOSITORY ?= https://github.com/athenianco/ci
+CI_BRANCH ?= master
+CI_PATH ?= .ci
+MAKEFILE := $(CI_PATH)/Makefile.main
+$(MAKEFILE):
+	git clone --quiet --depth 1 -b $(CI_BRANCH) $(CI_REPOSITORY) $(CI_PATH);
+-include $(MAKEFILE)


Makefile
```diff
@@ -0,0 +1,37 @@
+# TODO: this Makefile will probably be reusable, we should consider moving it somewhere
```
se7entyse7en: Heavy +1

se7entyse7en: I think that we could move it to infra, at least for now.

.travis.yml
```diff
@@ -2,20 +2,44 @@ language: python
...
   - pip install -r test-requirements.txt
   - pip install -e .
 
-script:
-  - flake8
-  - pytest
+stages:
+  - name: test
+  - name: deploy
+    if: branch = add-cd AND type != pull_request
```
se7entyse7en: Shouldn't this be an OR?

Makefile
lwsanty: @se7entyse7en infra is private, create a CI repo maybe?

.travis.yml
lwsanty: it's WIP, here should be `if: branch = add-cd AND type != pull_request`
should be and because it will trigger deploy in the case of PR to pool request, please correct me if I'm wrong

lwsanty:
blocked by https://github.com/athenianco/athenian-api/issues/10

Makefile
se7entyse7en: oh, right. Yup, a CI repo makes sense IMO.

.travis.yml
se7entyse7en: ah ok, I got it now. thanks!

Makefile
lwsanty: Done

.travis.yml
```diff
@@ -2,33 +2,44 @@ language: python
...
 cache:
-  pip: true
   directories:
-    - $HOME/docker
+    - "$HOME/.cache/pip"
+    - "$HOME/google-cloud-sdk/"
+
 before_cache:
   - chown -R travis:travis $HOME/.cache/pip
-  - docker save -o $HOME/docker/images.tar athenian/api:latest
```
vmarkovtsev: An unclean rebase overwrote my caching trick.

.travis.yml
```diff
@@ -2,33 +2,44 @@ language: python
...
       script:
         - flake8
         - pytest
-    - before_install:
-        - docker load -i $HOME/docker/images.tar || true
-        - docker image list
-      install:
-        - docker build --cache-from athenian/api -t athenian/api .
+    - stage: deploy
+      name: 'Deploy API'
```
vmarkovtsev:
```suggestion
      name: 'Deploy'
```

.travis.yml
lwsanty: the problem is we use makefile from https://github.com/athenianco/ci and I did not know how to propagate this caching in convenient way... if you have some good idea how to achieve this I can fix it...

lwsanty: @vmarkovtsev Done

.travis.yml
```diff
@@ -2,33 +2,53 @@ language: python
...
-        - pip install -r requirements.txt
-        - pip install -r lint-requirements.txt
-        - pip install -r test-requirements.txt
-        - pip install -e .
+    - stage: test
+      name: 'Test'
       script:
         - flake8
         - pytest
-    - before_install:
```
vmarkovtsev: There should be two parallel testing stage steps: the one that runs flake8 and pytest and the one that builds the image.

The deployment step should simply `make docker-push`.

Rationale: it has been common to screw up the docker build and the failure will show up only after the merge.

.travis.yml
```diff
@@ -2,33 +2,53 @@ language: python
 python: 3.7
 sudo: true
 dist: bionic
+
+env:
+  global:
+    - DOCKER_CACHE_IMAGE=athenian/api
+    - GOOGLE_DOCKER_IMAGE=gcr.io/athenian-1/api
```
vmarkovtsev: I don't feel strongly about the original tag name. How about having the same everywhere so that there is no need for retagging.

.travis.yml
vmarkovtsev: I think I get the problem: we will run the docker build twice after merge. We could restrict the testing step that builds docker only for PRs.

.travis.yml
lwsanty: we push the image with tag `gcr.io/athenian-1/api:branch-commit(7 symbols)` so we will retag anyway

.travis.yml
lwsanty: just to make sure that I understood you correctly:
- for a regular branch commit we run only tests
- for a PR we run tests and docker build/docker run in parallel
- for master we run run tests and docker build/docker run in parallel + deploy at the end, am I right?

vmarkovtsev: Almost. Change:
* for master we run run tests and deploy at the end (no docker build/run in the testing stage).

.travis.yml
vmarkovtsev: Anyway, there will be less code and simpler scripts.

.travis.yml
lwsanty: Done

.travis.yml
lwsanty: Done

PR #12, opened 2019-12-16
Fix the illegal instruction crash

vmarkovtsev:


se7entyse7en:
I trust you here @vmarkovtsev 😂

PR #20, opened 2019-12-19
Add authentication with Auth0

dennwc:
For now the Auth is optional and is enabled by setting `AUTH0_DOMAIN` and `AUTH0_AUDIENCE` as mentioned in the `DEPLOYMENT`.

---
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/20)

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -16,8 +16,10 @@ docker run -it --rm --entrypoint python3 athenian/api -m athenian.api.models.sta
 
 The server requires:
 
-- `SENTRY_KEY` and `SENTRY_PROJECT` environment variables.
-- Accessible PostgreSQL endpoint.
+- (optional) `SENTRY_KEY` and `SENTRY_PROJECT` environment variables to enable error logging.
+- (optional) `AUTH0_DOMAIN` and `AUTH0_AUDIENCE` environment variables to enable authorization.
+- Accessible PostgreSQL endpoint with the metadata.
+- Accessible PostgreSQL endpoint with the server state.
 - Exposing the configured HTTP port outside.
 
 ### Configuration
--- a/server/athenian/api/__init__.py
+++ b/server/athenian/api/__init__.py
@@ -11,6 +11,7 @@ import databases
 import sentry_sdk
 from sentry_sdk.integrations.aiohttp import AioHttpIntegration
 
+from athenian.api.auth import Auth0, auth_is_configured
 from athenian.api.controllers.status import setup_status
 from athenian.api.metadata import __description__, __package__, __version__
 from athenian.api.serialization import FriendlyJson
@@ -74,6 +75,7 @@ class AthenianApp(connexion.AioHttpApp):
         self.mdb = self.sdb = None  # type: Optional[databases.Database]
         self._mdb_connected_event = asyncio.Event()
         self._sdb_connected_event = asyncio.Event()
+        self._auth0 = None
 
         async def connect_to_mdb():
             db = databases.Database(mdb_conn, **(mdb_options or {}))
@@ -93,9 +95,26 @@ class AthenianApp(connexion.AioHttpApp):
         asyncio.get_event_loop().call_soon(asyncio.ensure_future, connect_to_mdb())
         asyncio.get_event_loop().call_soon(asyncio.ensure_future, connect_to_sdb())
 
+    async def close(self):
+        """Free resources associated with the object."""
+        if self._auth0 is not None:
+            await self._auth0.close()
+
     def create_app(self) -> aiohttp.web.Application:
         """Override the base class method to inject middleware."""
-        return aiohttp.web.Application(middlewares=[self.with_db])
+        middlewares = [self.with_db]
+        if not auth_is_configured():
+            # TODO: always require auth, or add a flag to assert that it's enabled
+            logging.getLogger(__package__).warning(
+                "API authentication is disabled; set AUTH0_DOMAIN and AUTH0_AUDIENCE")
+        else:
+            self._auth0 = Auth0(whitelist=[
+                r"/v1/openapi.json",
+                r"/v1/ui*",
+                r"/status*",
+            ])
+            middlewares.insert(0, self._auth0.middleware)
+        return aiohttp.web.Application(middlewares=middlewares)
 
     @aiohttp.web.middleware
     async def with_db(self, request, handler):
@@ -141,4 +160,7 @@ def main():
     log.info("Version %s", __version__)
     setup_sentry()
     app = AthenianApp(mdb_conn=args.metadata_db, sdb_conn=args.state_db, ui=args.ui)
-    app.run(host=args.host, port=args.port, use_default_access_log=True)
+    try:
+        app.run(host=args.host, port=args.port, use_default_access_log=True)
+    finally:
+        asyncio.get_event_loop().run_until_complete(app.close())
--- a/server/athenian/api/auth.py
+++ b/server/athenian/api/auth.py
@@ -1,0 +1,140 @@
+import logging
+import os
+import re
+
+import aiohttp.web
+import dateutil.parser
+from jose import jwt
+
+
+class AuthError(Exception):
+    """Auth error with an HTTP status code."""
+
+    def __init__(self, error, status_code):
+        """Create a new auth error."""
+        self.error = error
+        self.status_code = status_code
+
+
+class User():
+    """User profile information."""
+
+    def __init__(self, nickname: str, name: str, picture: str, updated_at: str, **kwargs):
+        """Create a new User object."""
+        self.username = nickname
+        self.fullname = name
+        self.picture = picture
+        self.updated = dateutil.parser.parse(updated_at)
+
+
+AUTH0_DOMAIN = os.getenv("AUTH0_DOMAIN")
+AUTH0_AUDIENCE = os.getenv("AUTH0_AUDIENCE")
+
+
+def auth_is_configured():
+    """Return true if auth is configured."""
+    return AUTH0_DOMAIN and AUTH0_AUDIENCE
+
+
+class Auth0:
+    """Class for Auth0 middleware compatible with aiohttp."""
+
+    def __init__(self, domain=AUTH0_DOMAIN, audience=AUTH0_AUDIENCE, whitelist=tuple()):
+        """Create a new Auth0 middleware."""
+        self._domain = domain
+        self._audience = audience
+        self._whitelist = whitelist
+        self._session = aiohttp.ClientSession()
+
+    async def close(self):
+        """Free resources associated with the object."""
+        await self._session.close()
+
+    def _is_whitelisted(self, request: aiohttp.web.Request) -> bool:
+        for pattern in self._whitelist:
+            if re.match(pattern, request.path):
+                return True
+        return False
+
+    def _get_token_auth_header(self, request: aiohttp.web.Request) -> str:
+        """Obtain the access token from the Authorization Header."""
+        try:
+            auth = request.headers["Authorization"]
+        except KeyError:
+            raise AuthError("Authorization header is expected", 401) from None
+
+        parts = auth.split()
+
+        if parts[0].lower() != "bearer":
+            raise AuthError("Authorization header must start with Bearer", 401)
+        elif len(parts) == 1:
+            raise AuthError("Token not found", 401)
+        elif len(parts) > 2:
+            raise AuthError('Authorization header must be "Bearer <token>"', 401)
+
+        token = parts[1]
+        return token
+
+    async def _get_user_info(self, token):
+        # TODO: cache based on decoded claims
+        resp = await self._session.get("https://%s/userinfo" % self._domain,
+                                       headers={"Authorization": "Bearer " + token})
+        user = await resp.json()
+        return User(**user)
+
+    @aiohttp.web.middleware
+    async def middleware(self, request, handler):
+        """Middleware function compatible with aiohttp."""
+        if self._is_whitelisted(request):
+            return await handler(request)
+
+        try:
+            token = self._get_token_auth_header(request)
+            unverified_header = jwt.get_unverified_header(token)
+        except AuthError as e:
+            return aiohttp.web.Response(body=e.error, status=e.status_code)
+        except jwt.JWTError:
+            return aiohttp.web.Response(body="Invalid header."
+                                        " Use an RS256 signed JWT Access Token.", status=401)
+        if unverified_header["alg"] != "RS256":
+            return aiohttp.web.Response(body="Invalid header."
+                                        " Use an RS256 signed JWT Access Token.", status=401)
+
+        # TODO: update periodically instead of pulling it on each request
+        jwks_req = await self._session.get("https://%s/.well-known/jwks.json" % self._domain)
+        jwks = await jwks_req.json()
+
+        # There might be multiple keys, find the one used to sign this request
+        rsa_key = None
+        for key in jwks["keys"]:
+            if key["kid"] == unverified_header["kid"]:
+                rsa_key = {k: key[k] for k in ("kty", "kid", "use", "n", "e")}
+
+        if rsa_key is None:
+            return aiohttp.web.Response(body="Unable to find an appropriate key", status=401)
+
+        try:
+            jwt.decode(
+                token,
+                rsa_key,
+                algorithms=["RS256"],
+                audience=self._audience,
+                issuer="https://%s/" % self._domain,
+            )
+        except jwt.ExpiredSignatureError:
+            return aiohttp.web.Response(body="token expired", status=401)
+        except jwt.JWTClaimsError:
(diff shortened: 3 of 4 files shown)


DEPLOYMENT.md
```diff
@@ -20,6 +20,9 @@ The server requires:
 - Accessible PostgreSQL endpoint.
 - Exposing the configured HTTP port outside.
 
+Optionally:
```
vmarkovtsev: Let's have it this way then:

The server requires:

- (optional) `SENTRY_KEY` and `SENTRY_PROJECT` environment variables to enable error logging.
- (optional) `AUTH0_DOMAIN` and `AUTH0_AUDIENCE` environment variables to enable authorization.
- Accessible PostgreSQL endpoint with the metadata.
- Accessible PostgreSQL endpoint with the server state.
- Exposing the configured HTTP port outside.

server/athenian/api/__init__.py
```diff
@@ -95,7 +96,18 @@ def __init__(self, mdb_conn: str, sdb_conn: str, ui: bool,
...
     def create_app(self) -> aiohttp.web.Application:
         """Override the base class method to inject middleware."""
-        return aiohttp.web.Application(middlewares=[self.with_db])
+        middlewares=[self.with_db]
+        if not AuthConfigured():
+            # TODO: always require auth, or add a flag to assert that it's enabled
+            logging.getLogger(__package__).warning("API authentication is disabled; set AUTH0_DOMAIN and AUTH0_AUDIENCE")
+        else:
+            auth0 = Auth0(whitelist=[
+                r'/v1/openapi.json',
```
vmarkovtsev:
```suggestion
                r'"v1/openapi.json",
```
etc.

I will add these to the OpenAPI definition once this is merged. I dunno how, but it should be possible.

server/athenian/api/__init__.py
```diff
@@ -95,7 +96,18 @@ def __init__(self, mdb_conn: str, sdb_conn: str, ui: bool,
...
+                r'/v1/ui*',
+                r'/status*',
```
vmarkovtsev: `/status` is polled by the Google Cloud health checker, we cannot wall it...

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+from jose import jwt
+
+# Format error response and append status code.
+class AuthError(Exception):
+    def __init__(self, error, status_code):
+        self.error = error
+        self.status_code = status_code
+
+class User():
+    def __init__(self, info):
```
vmarkovtsev: How about:

```python
import dateutil.parser

def __init__(self, nickname: str, name: str, picture: str, updated_at: str, **kwargs):
    self.username = nickname
    self.fullname = name
    self.picture = picture
    self.updated = dateutil.parser.parse(updated_at)
```

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+            raise AuthError("Authorization header must be Bearer token", 401)
+
+        token = parts[1]
+        return token
+
+    def _get_user_info(self, token):
+        # TODO: cache based on decoded claims
+        jsonurl = urlopen(HTTPRequest("https://"+self._domain+"/userinfo", headers={"Authorization": "Bearer "+token}))
+        resp = json.loads(jsonurl.read())
+        return User(resp)
```
vmarkovtsev: ...so here it changes to
```suggestion
        return User(**resp)
```

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+        self.picture = info["picture"]
+        self.updated = info["updated_at"]
+
+AUTH0_DOMAIN = os.environ.get("AUTH0_DOMAIN")
+AUTH0_AUDIENCE = os.environ.get("AUTH0_AUDIENCE")
+
+def AuthConfigured():
+    return AUTH0_DOMAIN and AUTH0_AUDIENCE
+
+class Auth0():
```
vmarkovtsev:
```suggestion
class Auth0:
```

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+    def __init__(self, info):
+        self.username = info["nickname"]
+        self.fullname = info["name"]
+        self.picture = info["picture"]
+        self.updated = info["updated_at"]
+
+AUTH0_DOMAIN = os.environ.get("AUTH0_DOMAIN")
+AUTH0_AUDIENCE = os.environ.get("AUTH0_AUDIENCE")
+
+def AuthConfigured():
```
vmarkovtsev:
```suggestion
def auth_is_configured():
```
(Python function naming style)

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+        self.status_code = status_code
+
+class User():
+    def __init__(self, info):
+        self.username = info["nickname"]
+        self.fullname = info["name"]
+        self.picture = info["picture"]
+        self.updated = info["updated_at"]
+
+AUTH0_DOMAIN = os.environ.get("AUTH0_DOMAIN")
```
vmarkovtsev:
```suggestion
AUTH0_DOMAIN = os.getenv("AUTH0_DOMAIN")
```
(same but shorter)

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+        self._algos = ["RS256"]
+        self._whitelist = whitelist
+
+    def _check_request(self, request):
+        for pattern in self._whitelist:
+            if re.match(pattern, request.path):
+                return True
+        return False
+
+    def _get_token_auth_header(self, request):
```
vmarkovtsev:
```suggestion
    def _get_token_auth_header(self, request: aiohttp.web.Request) -> str:
```
(nicer IDE experience)

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+    return AUTH0_DOMAIN and AUTH0_AUDIENCE
+
+class Auth0():
+    def __init__(self, domain=AUTH0_DOMAIN, audience=AUTH0_AUDIENCE, whitelist=tuple()):
+        self._domain = domain
+        self._audience = audience
+        self._algos = ["RS256"]
+        self._whitelist = whitelist
+
+    def _check_request(self, request):
```
vmarkovtsev:
```suggestion
    def _check_request(self, request: aiohttp.web.Request) -> bool:
```

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+        for pattern in self._whitelist:
+            if re.match(pattern, request.path):
+                return True
+        return False
+
+    def _get_token_auth_header(self, request):
+        """Obtains the access token from the Authorization Header
+        """
+        auth = request.headers.get("Authorization", None)
+        if not auth:
```
vmarkovtsev: More Pythonic:
```python
try:
    auth = request.headers["Authorization"]
except KeyError:
    raise AuthError("Authorization header is expected", 401) from None
```

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+            raise AuthError("Authorization header is expected", 401)
+
+        parts = auth.split()
+
+        if parts[0].lower() != "bearer":
+            raise AuthError("Authorization header must start with Bearer", 401)
+        elif len(parts) == 1:
+            raise AuthError("Token not found", 401)
+        elif len(parts) > 2:
+            raise AuthError("Authorization header must be Bearer token", 401)
```
vmarkovtsev:
```suggestion
            raise AuthError('Authorization header must be "Bearer Token"', 401)
```

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+        resp = json.loads(jsonurl.read())
+        return User(resp)
+
+    @aiohttp.web.middleware
+    async def middleware(self, request, handler):
+        if self._check_request(request):
+            return await handler(request)
+
+        # TODO: update periodically instead of pulling it on each request
+        jsonurl = urlopen("https://"+self._domain+"/.well-known/jwks.json")
```
vmarkovtsev: We need to change this to [aiohttp Client](https://docs.aiohttp.org/en/stable/client_quickstart.html#make-a-request). The reason is that Python does not those hipster goroutines, so the synchronous call to `urlopen` stops the whole server thread.

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+        jwks = json.loads(jsonurl.read())
+
+        try:
+            token = self._get_token_auth_header(request)
+            unverified_header = jwt.get_unverified_header(token)
+        except AuthError as e:
+            return aiohttp.web.Response(body=e.error, status=e.status_code)
+        except jwt.JWTError:
+            return aiohttp.web.Response(body="Invalid header. Use an RS256 signed JWT Access Token.", status=401)
+        if unverified_header["alg"] != self._algos[0]:
```
vmarkovtsev:
```suggestion
        if unverified_header["alg"] != "RS256":
```
I mean, we already hardcode RS256 in the error message, so...

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+            return aiohttp.web.Response(body="Invalid header. Use an RS256 signed JWT Access Token.", status=401)
+
+        # There might be multiple keys, find the one used to sign this request
+        rsa_key = None
+        for key in jwks["keys"]:
+            if key["kid"] == unverified_header["kid"]:
+                rsa_key = {
```
vmarkovtsev:
```python
rsa_key = {k: key[k] for k in ("kty", "kid", "use", "n", "e")}
```

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+                    "kty": key["kty"],
+                    "kid": key["kid"],
+                    "use": key["use"],
+                    "n": key["n"],
+                    "e": key["e"]
+                }
+
+        if rsa_key is None:
+            return aiohttp.web.Response(body="Unable to find appropriate key", status=401)
```
vmarkovtsev:
```suggestion
            return aiohttp.web.Response(body="Unable to find an appropriate key", status=401)
```

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+
+        try:
+            payload = jwt.decode(
+                token,
+                rsa_key,
+                algorithms=self._algos,
+                audience=self._audience,
+                issuer="https://"+self._domain+"/"
```
vmarkovtsev:
```suggestion
                issuer="https://%s/" % self._domain
```
(more Pythonic)

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+            )
+        except jwt.ExpiredSignatureError:
+            return aiohttp.web.Response(body="token expired", status=401)
+        except jwt.JWTClaimsError:
+            return aiohttp.web.Response(body="incorrect claims, please check the audience and issuer", status=401)
+        except Exception:
+            return aiohttp.web.Response(body="Unable to parse authentication token.", status=401)
```
vmarkovtsev:
```suggestion
            return aiohttp.web.Response(body="Unable to parse the authentication token.", status=401)
```

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+
+        user = self._get_user_info(token)
+        logging.getLogger(__package__).info("User %s", user.username)
```
vmarkovtsev:
```suggestion
        logging.getLogger("auth").info("User %s", user.username)
```
So that we can silence the auth logs or make them more important.

vmarkovtsev:
The error handling is impressive @dennwc, you went quite far with security :smile:

server/athenian/api/__init__.py
dennwc: This is a whitelist (stuff not walled by auth).

server/athenian/api/auth.py
dennwc: Is there a way to attach the name to `aiohttp` logs, btw?

server/athenian/api/auth.py
dennwc: Ouch. Will do.

server/athenian/api/__init__.py
vmarkovtsev: Oops sorry, I confused the lists! Then it should be fine as-as.

server/athenian/api/auth.py
vmarkovtsev: The only not-too-magical way I see is to wrap the logging calls so that they always require setting the user info.

server/athenian/api/auth.py
```diff
@@ -0,0 +1,119 @@
...
+        request.userinfo = user
```
vmarkovtsev:
```suggestion
        request.user = user
```
how about this? This would correspond to Django (right @se7entyse7en?)

dennwc:
Resolved most comments, the client change pending.

dennwc:
@vmarkovtsev Changed the HTTP client library. Getting this then the app is killed:
```
ERROR:bd0d:asyncio:Unclosed client session
```
So it's clear I need to close session when the app is going to close. But is the right way to do it? `session.close()` is async, so looks like I cannot call it in `__del__`...

server/athenian/api/auth.py
```diff
@@ -0,0 +1,118 @@
...
+    async def _get_user_info(self, token):
+        # TODO: cache based on decoded claims
+        resp = await self._session.get("https://%s/userinfo" % self._domain,
+                headers={"Authorization": "Bearer "+token})
+        user = await resp.json()
+        return User(**user)
+
+    @aiohttp.web.middleware
+    async def middleware(self, request, handler):
+        if self._check_request(request):
```
se7entyse7en:
```suggestion
        if self._is_whitelisted(request):
```
This seems clearer.

server/athenian/api/auth.py
```diff
@@ -0,0 +1,118 @@
...
+            return await handler(request)
+
+        try:
+            token = self._get_token_auth_header(request)
+            unverified_header = jwt.get_unverified_header(token)
+        except AuthError as e:
+            return aiohttp.web.Response(body=e.error, status=e.status_code)
+        except jwt.JWTError:
+            return aiohttp.web.Response(body="Invalid header. Use an RS256 signed JWT Access Token.", status=401)
+        if unverified_header["alg"] != "RS256":
```
se7entyse7en: When this could happen?

server/athenian/api/auth.py
```diff
@@ -0,0 +1,118 @@
...
+
+def auth_is_configured():
+    return AUTH0_DOMAIN and AUTH0_AUDIENCE
+
+class Auth0:
+    def __init__(self, domain=AUTH0_DOMAIN, audience=AUTH0_AUDIENCE, whitelist=tuple()):
+        self._domain = domain
+        self._audience = audience
+        self._whitelist = whitelist
+        self._session = aiohttp.ClientSession()
```
vmarkovtsev: So IMHO we should do the following for a clean exit:

1. Add `async def close(self)` to `Auth0`
2. Save `Auth0` instance inside `AthenianApp`
3. Add `async def close(self)` to `AthenianApp` which `await`-s `Auth0.close()`
4. In `def main()`:

```python
try:
    app.run(...)
finally:
    asyncio.get_event_loop().run_until_complete(app.close())
```

server/athenian/api/auth.py
```diff
@@ -0,0 +1,118 @@
...
+            return aiohttp.web.Response(body="Unable to parse the authentication token.", status=401)
+
+        try:
+            user = await self._get_user_info(token)
+            logging.getLogger("auth").info("User %s", user.username)
+        except Exception:
+            return aiohttp.web.Response(body="The token is likely revoked.", status=401)
+
+        request.user = user
+        return await handler(request)
```
vmarkovtsev:
```suggestion
        return await handler(request)

```

server/athenian/api/auth.py
dennwc: I think only is we misconfigure Auth0.

server/athenian/api/auth.py
dennwc: Done, getting new errors now:
```
RuntimeError: Event loop is closed
sys:1: RuntimeWarning: coroutine 'AthenianApp.close' was never awaited
ERROR:7224:asyncio:Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7fa60e01afd0>
ERROR:7224:asyncio:Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x7fa5ee86ba70>, 403408.321548599)]']
connector: <aiohttp.connector.TCPConnector object at 0x7fa5f4ea2b70>
```

vmarkovtsev: OK let's leave it like this. I will fix it after merging, this requires some thought. I know why it happened but the fix requires some experimentation.

PR #29, opened 2020-01-05
Repository sets

vmarkovtsev:
https://athenianco.atlassian.net/browse/ENG-106

To be rebased on #28


server/athenian/api/controllers/reposet_controller.py
```diff
@@ -0,0 +1,81 @@
...
+
+from athenian.api.controllers.response import response, ResponseError
+from athenian.api.models.state.models import RepositorySet
+from athenian.api.models.web import CreatedIdentifier, ForbiddenError, NotFoundError
+
+
+async def create_reposet(request: web.Request, body: List[str]) -> web.Response:
+    """Create a repository set.
+
+    :param id: Numeric identifier of the repository set to list.
```
se7entyse7en:
```suggestion
```

server/athenian/api/models/state/models.py
```diff
@@ -0,0 +1,69 @@
...
+    """Mixin to invoke default() and onupdate() on all the columns."""
+
+    class Context:
+        """Pretend to be a fully-featured SQLAlchemy execution context."""
+
+        def __init__(self, parameters: dict):
+            """init"""
+            self.current_parameters = parameters
+
+    def create_defaults(self):
```
se7entyse7en: Isn't this done automatically? I mean when creating an instance of `RepositorySet`, doesn't the unset fields have the default value? AFAIS sqlalchemy already handles a `default` kwarg. Are you overriding its behavior?

BTW it's not the top that we need to right these behaviors our own :(

server/athenian/api/controllers/reposet_controller.py
```diff
@@ -0,0 +1,81 @@
...
+    rs = RepositorySet(owner=request.user.username, items=body)
+    rs.create_defaults()
+    rid = await request.sdb.execute(insert(RepositorySet).values(rs.explode()))
+    return response(CreatedIdentifier(rid))
+
+
+async def delete_reposet(request: web.Request, id: int) -> web.Response:
+    """Delete a repository set.
+
+    :param id: Numeric identifier of the repository set to list.
```
se7entyse7en:
```suggestion
    :param id: Numeric identifier of the repository set to delete.
```

server/athenian/api/controllers/reposet_controller.py
```diff
@@ -0,0 +1,81 @@
...
+        return ResponseError(ForbiddenError(
+            detail="User %s is not allowed to access %d" % (request.user.username, id))).response
+    await request.sdb.execute(delete(RepositorySet).where(RepositorySet.id == id))
+    return web.Response(status=200)
+
+
+async def get_reposet(request: web.Request, id: int) -> web.Response:
+    """List a repository set.
+
+    :param id: Numeric identifier of the repository set to list.
```
se7entyse7en:
```suggestion
    :param id: Numeric identifier of the repository set to get.
```

server/athenian/api/controllers/reposet_controller.py
```diff
@@ -0,0 +1,81 @@
...
+        return ResponseError(ForbiddenError(
+            detail="User %s is not allowed to access %d" % (request.user.username, id))).response
+    # "items" collides with dict.items() so we have to access the list via []
+    return web.json_response(rs["items"], status=200)
+
+
+async def update_reposet(request: web.Request, id: int, body: List[str]) -> web.Response:
+    """Update a repository set.
+
+    :param id: Numeric identifier of the repository set to list.
```
se7entyse7en:
```suggestion
    :param id: Numeric identifier of the repository set to update.
```

server/athenian/api/controllers/status_controller.py
```diff
@@ -28,26 +28,37 @@
...
+        registry=registry,
     )
     app["REQUEST_IN_PROGRESS"] = prometheus_client.Gauge(
         "requests_in_progress_total", "Requests in progress",
         ["app_name", "endpoint", "method"],
+        registry=registry,
     )
     app.middlewares.insert(0, instrument)
-    app.router.add_get("/status", render_status)
+    app.router.add_get("/status", StatusRenderer(registry))
```
se7entyse7en: Are you trying this with Prometheus running locally?

server/athenian/api/models/state/versions/e7afe365ab0e_create_repository_sets_table.py
```diff
@@ -0,0 +1,42 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "e7afe365ab0e"
+down_revision = "34eafe9e7cd9"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    def count_items(ctx):
```
se7entyse7en: So adding an inner `Context` is a standard that is handled by sqlalchemy and alembic theirselves?

DEPLOYMENT.md
```diff
@@ -17,7 +17,7 @@ docker run -it --rm --entrypoint python3 athenian/api -m athenian.api.models.sta
 The server requires:
 
 - (optional) `SENTRY_KEY` and `SENTRY_PROJECT` environment variables to enable error logging.
-- (optional) `AUTH0_DOMAIN` and `AUTH0_AUDIENCE` environment variables to enable authorization.
+- `AUTH0_DOMAIN` and `AUTH0_AUDIENCE` environment variables to enable authorization.
```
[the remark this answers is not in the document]
vmarkovtsev: It is impossible to have CRUD without auth. Instead of having two different server execution modes, I hardcoded the user for now.

server/athenian/api/auth.py
```diff
@@ -16,29 +16,24 @@ def __init__(self, error, status_code):
...
 
-class User():
+class User:
     """User profile information."""
 
-    def __init__(self, nickname: str, name: str, picture: str, updated_at: str, **kwargs):
+    def __init__(self, profile: str, name: str, picture: str, updated_at: str, **kwargs):
         """Create a new User object."""
-        self.username = nickname
+        self.username = profile.split("://", 1)[1]
```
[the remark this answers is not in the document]
vmarkovtsev: This is taken from Auth0 docs: https://auth0.com/docs/api/authentication#get-user-info

server/athenian/api/models/state/models.py
vmarkovtsev: This indeed sucks. I left a comment about this on line 7.

SQLAlchemy has two layers, and those fancy defaults and onupdate-s belong to the ORM. However, `databases`, the asyncio DB engine we are using, does not support the ORM layer.

server/athenian/api/controllers/status_controller.py
vmarkovtsev: The good thing about Prometheus is that you don't have to run it at all :smile: So I never looked at the actual plots.

server/athenian/api/models/state/models.py
se7entyse7en: Ah ok, it's clear now. It was weird to me when I read it, but now I understand that it's specifically for the asyncio DB engine. That would be great if you could specify it 👍

server/athenian/api/models/state/versions/e7afe365ab0e_create_repository_sets_table.py
vmarkovtsev: Yes, this is how the ORM handles them. Strictly speaking, alembic does not care about defaults and onupdate-s, so I should remove them from here.

server/athenian/api/auth.py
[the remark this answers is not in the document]
se7entyse7en: okay, also we'll iterate on this according to the user flow.

DEPLOYMENT.md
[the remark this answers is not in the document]
se7entyse7en: > Instead of having two different server execution modes, I hardcoded the user for now.

This totally makes sense IMO for now. We'll then need to think about how to tackle having a public dashboard for dogfooding using the same SaaS instance.

PR #42, opened 2020-01-13
Fix handling empty SQL resultsets

vmarkovtsev:
Fixes ENG-126

### DIFF
--- a/server/athenian/api/models/web/for_set.py
+++ b/server/athenian/api/models/web/for_set.py
@@ -26,7 +26,7 @@ class ForSet(Model):
         }
 
         self._repositories = repositories
-        self._developers = developers
+        self._developers = developers or []
 
     @classmethod
     def from_dict(cls, dikt: dict) -> "ForSet":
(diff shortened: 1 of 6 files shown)


server/athenian/api/models/web/for_set.py
```diff
@@ -26,7 +26,7 @@ def __init__(
         }
 
         self._repositories = repositories
```
se7entyse7en: Shall we for `repositories` as well? The constructor accepts `None`, but typing is not `Optional`.

vmarkovtsev: Repositories may not be empty by the API contract. Those constructor defaults must allow the engine to initialize an object without any kwargs. This is confusing. If I were designing the logic, I would remove all the arguments from `__init__`, but we have to live with what we have.

vmarkovtsev: In other words, if `repositories` is None, then the model is not initialized yet. However, `developers` has a default value which is an empty list.

PR #45, opened 2020-01-14
Add OFFLINE migrations

vmarkovtsev:
Addresses ENG-118

### DIFF
--- a/server/athenian/api/models/state/README.md
+++ b/server/athenian/api/models/state/README.md
@@ -11,6 +11,10 @@ python3 -m athenian.api.models.state postgres://user:password@host:port/database
 Replace `postgres://user:password@host:port/database` with the actual DB connection string in
 [SQLAlchemy format](https://docs.sqlalchemy.org/en/13/core/engines.html).
 
+It is possible to define `OFFLINE` environment variable (to any non-empty value) so that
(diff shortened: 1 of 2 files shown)


vmarkovtsev:
@lwsanty Does this work for you or there is smth extra/different that you need?

server/athenian/api/models/state/README.md
```diff
@@ -11,6 +11,10 @@ python3 -m athenian.api.models.state postgres://user:password@host:port/database
 Replace `postgres://user:password@host:port/database` with the actual DB connection string in
 [SQLAlchemy format](https://docs.sqlalchemy.org/en/13/core/engines.html).
 
+It is possible to define `OFFLINE` environment variable (to any non-empty value) so that
+the SQL operations dump is written to the standard output while the actual DB is stays untouched.
```
lwsanty: maybe I have some problems with English today, but got a nitpick:
> while the actual DB is stays untouched.

`while the actual DB stays untouched.` or `while the actual DB is staying untouched.`

lwsanty:
@vmarkovtsev let's try it with this format first :thumbsup:

server/athenian/api/models/state/README.md
vmarkovtsev: Oops

PR #46, opened 2020-01-14
Add /v1/user and change the user object

vmarkovtsev:

### DIFF
--- a/server/athenian/api/auth.py
+++ b/server/athenian/api/auth.py
@@ -19,10 +19,11 @@ class AuthError(Exception):
 class User:
     """User profile information."""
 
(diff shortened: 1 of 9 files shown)


server/athenian/api/auth.py
```diff
@@ -19,10 +19,11 @@ def __init__(self, error, status_code):
 class User:
     """User profile information."""
 
-    def __init__(self, profile: str, name: str, picture: str, updated_at: str, **kwargs):
+    def __init__(self, sub: str, email: str, name: str, picture: str, updated_at: str, **kwargs):
```
dpordomingo: I'm a newbie with Python, but... could this change cause https://athenianco.atlassian.net/browse/ENG-138

PR #56, opened 2020-01-20
User accounts

vmarkovtsev:


server/athenian/api/controllers/reposet.py
```diff
@@ -50,7 +55,16 @@
     rs = await db.fetch_one(select(columns).where(RepositorySet.id == id))
     if rs is None or len(rs) == 0:
         raise ResponseError(NotFoundError(detail="Repository set %d does not exist" % id))
-    if rs[RepositorySet.owner.key] != user.id:
+    if account is None:
+        # if there is no account, suppose the reposet owner
+        account = rs[RepositorySet.owner.key]
```
se7entyse7en: Shouldn't it be the `account` the `user` is associated with? So something like:
```
if account is None:
    account = <get account from user>

if rs[RepositorySet.owner.key] != account:
   ...
```

server/athenian/api/models/state/models.py
```diff
@@ -67,3 +68,24 @@ def count_items(ctx):
...
     count_items = staticmethod(count_items)
+
+
+class UserAccount(Base):
+    """User<>account many-to-many relations."""
+
+    __tablename__ = "user_accounts"
+
+    user_id = Column("user_id", String(256), primary_key=True)
+    account_id = Column("account_id", Integer(), ForeignKey("accounts.id", name="fk_user_account"),
```
se7entyse7en: I'd add a unique constraint on the tuple `(user_id, account_id)`.

server/athenian/api/controllers/reposet.py
```diff
@@ -50,7 +55,16 @@
...
+    elif rs[RepositorySet.owner.key] != account:
         raise ResponseError(ForbiddenError(
             detail="User %s is not allowed to access %d" % (user.id, id)))
-    return RepositorySet(**rs)
+    # finally check that the user belongs to the account
+    status = await db.fetch_one(select([UserAccount.is_admin]).where(
+        and_(UserAccount.user_id == user.id, UserAccount.account_id == account)))
+    if status is None:
+        raise ResponseError(ForbiddenError(
+            detail="User %s does not belong to account %d" % (user.id, account)))
```
se7entyse7en: I'd split this logic into a separate `is_admin` function since it's also used in other parts.  For example, when modifying a reposet to me makes more sense to check first whether the client is an admin.

server/athenian/api/controllers/reposet_controller.py
```diff
@@ -1,67 +1,86 @@
...
-async def delete_reposet(request: web.Request, id: int) -> web.Response:
+async def delete_reposet(request: AthenianWebRequest, id: int) -> web.Response:
     """Delete a repository set.
 
     :param id: Numeric identifier of the repository set to delete.
     :type id: int
     """
     try:
-        await fetch_reposet(id, [], request.sdb, request.user)
+        _, is_admin = await fetch_reposet(id, [], request.sdb, request.user, None)
```
se7entyse7en: As said in another comment I'd just check if is admin then delete.

server/athenian/api/controllers/user_controller.py
```diff
@@ -1,8 +1,34 @@
...
-async def get_user(request: web.Request) -> web.Response:
+async def get_user(request: AthenianWebRequest) -> web.Response:
     """Return details about the current user."""
-    return web.json_response(vars(request.user), status=200, dumps=FriendlyJson.dumps)
+    await request.user.load_accounts(request.sdb)
+    return response(request.user)
+
+
+async def get_account(request: AthenianWebRequest, id: int) -> web.Response:
+    """Return details about the current user."""
```
se7entyse7en:
```suggestion
    """Return details about the current account."""
```

server/athenian/api/controllers/reposet.py
vmarkovtsev: This is equivalent, I think. The code below is

```python
status = await db.fetch_one(select([UserAccount.is_admin]).where(
        and_(UserAccount.user_id == user.id, UserAccount.account_id == account)))
```

So we check the user's access to the reposet in SQL. Taking an external `account` is a convenience for the metrics endpoint which ensures that the user does not mix reposets from different accounts together.

I should remove the `account` arg completely and check the reposet mixture on the calling side.

server/athenian/api/models/state/models.py
vmarkovtsev: They are both `primary_key=True` so according to the docs they should be unique together (composite primary key).

server/athenian/api/controllers/reposet.py
vmarkovtsev: Cool!

PR #57, opened 2020-01-21
Invites

vmarkovtsev:
To be rebased on the merged #56


vmarkovtsev:
@lwsanty This adds yet another secret environment variable, `ATHENIAN_INVITATION_KEY`. You should assign it to a good password of your choice.

server/athenian/api/controllers/invitation_controller.py
```diff
@@ -0,0 +1,160 @@
...
+            select([Invitation.account_id, Invitation.accepted, Invitation.is_active])
+            .where(and_(Invitation.id == iid, Invitation.salt == salt)))
+        if inv is None:
+            return ResponseError(NotFoundError(
+                detail="Invitation was not found.")).response
+        if not inv[Invitation.is_active.key]:
+            return ResponseError(ForbiddenError(
+                detail="This invitation is disabled.")).response
+        acc = inv[Invitation.account_id.key]
+        is_admin = acc == admin_backdoor
```
se7entyse7en: I'd rename `acc` -> `acc_id` just for clarity.

server/athenian/api/controllers/invitation_controller.py
```diff
@@ -0,0 +1,160 @@
...
+        if is_admin:
+            # create a new account for the admin user
+            acc = await conn.execute(insert(Account).values(Account().create_defaults().explode()))
+            if acc == admin_backdoor:
```
se7entyse7en: I think that this should be:

```suggestion
            if acc[Invitation.account_id.key] == admin_backdoor:
```

server/README.md
```diff
@@ -45,6 +45,14 @@ pytest
 
 Prometheus monitoring: `http://localhost:8080/status`.
 
+Generating admin invitations:
+
+```
+ATHENIAN_INVITATION_KEY=secret python3 -m athenian.api.invite_admin sqlite://
+```
+
+Replace `sqlite://` with the actual DB endpoint and `secret` with the actual passphrase.
```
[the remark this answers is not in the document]
vmarkovtsev: So true.

server/athenian/api/controllers/invitation_controller.py
vmarkovtsev: This one no, the insertion result is an integer primary key. IDK how all the underlying magic works but that's a tested fact...

PR #63, opened 2020-01-23
Gracefully handle Auth0 mgmt API rate limit

vmarkovtsev:
Fixes https://sentry.io/organizations/athenianco/issues/1466357502

### DIFF
--- a/server/athenian/api/auth.py
+++ b/server/athenian/api/auth.py
@@ -1,9 +1,11 @@
 import asyncio
 from datetime import timedelta
+from http import HTTPStatus
 import logging
 import os
+from random import random
 import re
-from typing import Iterable, List, Sequence
+from typing import List, Optional, Sequence
 
 import aiohttp.web
 from aiohttp.web_runner import GracefulExit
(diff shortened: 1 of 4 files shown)


server/athenian/api/auth.py
```diff
@@ -117,16 +118,30 @@ def ensure_static_configuration(cls):
...
-            try:
-                resp = await self._session.get("https://%s/api/v2/users/%s" % (self._domain, user),
-                                               headers={"Authorization": "Bearer " + token})
-            except RuntimeError:
-                # our loop is closed and we are doomed
+        assert len(users) >= 0  # we need __len__
+
+        async def get_user(user: str) -> Optional[User]:
+            retries = 1
+            while retries <= 10:
```
se7entyse7en: You could eventually use a `for-else` block to avoid handling counter and checking again the status:

```
for i in range(1, 11):
    try:
        ...
        if resp.status == HTTPStatus.TOO_MANY_REQUESTS:
            ...
        else:
            break
    except:
        return None
else:
    return None
```

vmarkovtsev: Much better this way!

PR #59, opened 2020-01-23
Update the schema to match `metadata`

vmarkovtsev:

PR #65, opened 2020-01-27
Progress endpoint

vmarkovtsev:
To be rebased on #59


server/athenian/api/openapi/openapi.yaml
```diff
@@ -140,6 +140,40 @@ paths:
         be an admin of the specified account.
       x-openapi-router-controller: athenian.api.controllers.invitation_controller
       x-codegen-request-body-name: body
+  /invite/progress/{id}:
```
se7entyse7en: Do we have an endpoint to get  the account id given the current request?

vmarkovtsev: `/invite/accept` returns it, yes.

vmarkovtsev: I suggest saving it to the local storage so that you don't have to hack "the first element of /user.accounts".

se7entyse7en: Exactly that was my concern. I mean AFAIK `/invite/accept` is supposed to be called once. So, saving to local storage would work, but would be limited. The cases not covered are for example deletion of the local storage, or using another browser, or going incognito mode.

vmarkovtsev: The API is designed that a user can be a member of multiple accounts, so there is not much the server can help with. As long as we have the internal contract that the mapping is 1-1 user-account, it is safe to query `/user` and take the only (first) record in `.accounts`.

se7entyse7en: Okay, that makes sense, also because at least for now in the webapp there's no way to handle being part of multiple accounts.

PR #66, opened 2020-01-28
Help development with clean command and docs

dpordomingo:
What do you think about adding instructions about how to clean data and use your own image when developing?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,9 +26,15 @@ $(ENV_FILE):
 
 $(IO_DIR)/%.sqlite: $(ENV_FILE)
 	docker run --rm -e DB_DIR=/io -v$(IO_DIR):/io --env-file $(ENV_FILE) --entrypoint python3 $(IMAGE) /server/tests/gen_sqlite_db.py
-	@touch $@
+	if [ ! -f $@ ]; then @touch $@; fi
 
 .PHONY: run-api
-
 run-api: $(IO_DIR)/mdb.sqlite $(IO_DIR)/sdb.sqlite
 	docker run --rm -p 8080:8080 -v$(IO_DIR):/io --env-file $(ENV_FILE) $(IMAGE) --ui --metadata-db=sqlite:///io/mdb.sqlite --state-db=sqlite:///io/sdb.sqlite
+
+.PHONY: clean
+clean: fixtures-clean
+
+.PHONY: fixtures-clean
+fixtures-clean:
+	rm -rf $(IO_DIR)/mdb.sqlite $(IO_DIR)/sdb.sqlite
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -26,9 +26,15 @@ $(ENV_FILE):
 
 $(IO_DIR)/%.sqlite: $(ENV_FILE)
 	docker run --rm -e DB_DIR=/io -v$(IO_DIR):/io --env-file $(ENV_FILE) --entrypoint python3 $(IMAGE) /server/tests/gen_sqlite_db.py
-	@touch $@
+	if [ ! -f $@ ]; then @touch $@; fi
```
dpordomingo: This avoids an error thrown when the database already exists. It happens because the data is created from the inside of the container with root privileges, so it's not possible to `touch` it without sudoing.

dpordomingo:
I don't understand how is this failing in Travis
https://travis-ci.com/athenianco/athenian-api/jobs/280906460

Makefile
vmarkovtsev: Does it run every time now? We should probably `docker --user $$(id -u)` instead

server/README.md
```diff
@@ -101,6 +101,18 @@ make run-api
 
 And open http://localhost:8080/v1/ui
 
+If you want to run your own API image, use instead:
+```
+# Run the API container
+IMAGE=your_api_image:tag make run-api
+```
+
+You can clean API data, created from `make run-api`, by running:
```
vmarkovtsev:
```suggestion
You can erase the API data fixtures created by `make run-api` with:
```

Makefile
dpordomingo: After reviewing this... what's the purpose of the `touch` here?
Aren't `mdb.sqlite` and `sdb.sqlite` fixtures generated by the `docker run ... gen_sqlite_db.py` above? If so, why is it needed to `touch` them?

vmarkovtsev: Otherwise `make` thinks that it needs to re-run the rule: the modification timestamps become out of order.

dpordomingo: to run again `$(IO_DIR)/%.sqlite`?
Why? if the passed file already exists, it is not needed to rerun the target, is it?
I locally tried after deleting the `touch`, and it didn't run the `docker run ... gen_sqlite_db.py` again after creating `mdb.sqlite` and `sdb.sqlite` fixtures for the first time.

vmarkovtsev: The databases used to be recreated over and over for me without `touch` - this is why I had to add it, apparently. Let me test.

vmarkovtsev: Yep, this is no longer needed now. Good! Then can you please remove this line completely.

vmarkovtsev:
@dpordomingo It fails in Travis because the build relies on the secrets that are not accessible to the forks. You should push to a branch inside this repo and then create a new PR.

PR #67, opened 2020-01-28
Fix the state fixture generation

vmarkovtsev:
Fixes ENG-156

### DIFF
--- a/server/athenian/api/models/state/__main__.py
+++ b/server/athenian/api/models/state/__main__.py
@@ -1,10 +1,11 @@
 import os
+import subprocess
 import sys
 
 import jinja2
 
 
-def main():
+def main(url=None, exec=True):
     """
     Initialize the server state DB.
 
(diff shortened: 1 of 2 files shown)


server/athenian/api/models/state/__main__.py
```diff
@@ -23,8 +23,11 @@ def main():
     args = ["alembic", "alembic", "upgrade", "head"]
     if os.getenv("OFFLINE"):
         args.append("--sql")
-    os.execlp(*args)
+    if exec:
+        os.execlp(*args)
+    else:
+        subprocess.run(" ".join(args[1:]), check=True, shell=True)
```
se7entyse7en: Not a big deal, but do we need this? I'd change this as:
```
subprocess.run(args, check=True, capture_output=True)
```
as no need also to add slicing if we remove the first value.

vmarkovtsev: `capture_output=True` will not automatically echo the command's output, unfortunately - this is what I wanted here.

PR #70, opened 2020-01-28
Gods

vmarkovtsev:
ENG-175

To be rebased on #59

### DIFF
--- a/server/athenian/api/auth.py
+++ b/server/athenian/api/auth.py
@@ -13,7 +13,9 @@ from connexion.exceptions import OAuthProblem
 from connexion.lifecycle import ConnexionRequest
 from connexion.operations import secure
 from jose import jwt
+from sqlalchemy import select
 
+from athenian.api.models.state.models import God
 from athenian.api.models.web.user import User
 
 
@@ -340,3 +342,11 @@ class Auth0:
             raise OAuthProblem(description="invalid claims, please check the audience and issuer")
         except jwt.JWTError:
             raise OAuthProblem(description="Unable to parse the authentication token.")
+
+        god = await request.sdb.fetch_one(select([God.mapped_id]).where(God.user_id == user.id))
+        if god is not None:
+            mapped_id = god[God.mapped_id.key]
+            if mapped_id is not None:
+                user.god_id = user.id
+                user.id = (await self.get_users([mapped_id]))[0]
+                self.log.info("God mode: %s became %s", user.god_id, user.id)
--- a/server/athenian/api/auth.py
+++ b/server/athenian/api/auth.py
@@ -347,6 +347,7 @@ class Auth0:
         if god is not None:
             mapped_id = god[God.mapped_id.key]
             if mapped_id is not None:
-                user.god_id = user.id
-                user.id = (await self.get_users([mapped_id]))[0]
-                self.log.info("God mode: %s became %s", user.god_id, user.id)
+                god_id = user.id
+                user = (await self.get_users([mapped_id]))[0]
+                user.god_id = god_id
+                self.log.info("God mode: %s became %s", god_id, mapped_id)
--- a/server/athenian/api/auth.py
+++ b/server/athenian/api/auth.py
@@ -1,5 +1,6 @@
 import asyncio
 from datetime import timedelta
+import functools
 from http import HTTPStatus
 import logging
 import os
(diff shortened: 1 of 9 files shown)


server/athenian/api/auth.py
```diff
@@ -273,4 +275,11 @@ def _get_token_auth_header(self, request: aiohttp.web.Request) -> str:
         except Exception:
             return aiohttp.web.Response(body="Your auth token is likely revoked.", status=401)
 
+        god = await request.sdb.fetch_one(select([God.mapped_id]).where(God.user_id == user.id))
+        if god is not None:
+            mapped_id = god[God.mapped_id.key]
+            if mapped_id is not None:
+                user.god_id = user.id
+                user.id = (await self.get_users([mapped_id]))[0]
```
se7entyse7en: Why not simply as follows?
```suggestion
                user.id = mapped_id
```

server/athenian/api/auth.py
```diff
@@ -273,4 +275,11 @@ def _get_token_auth_header(self, request: aiohttp.web.Request) -> str:
         except Exception:
             return aiohttp.web.Response(body="Your auth token is likely revoked.", status=401)
 
+        god = await request.sdb.fetch_one(select([God.mapped_id]).where(God.user_id == user.id))
+        if god is not None:
+            mapped_id = god[God.mapped_id.key]
+            if mapped_id is not None:
+                user.god_id = user.id
```
se7entyse7en: 🤔I don't understand how this works. I mean, shouldn't this be done independently from whether the god is currently mapped already? Otherwise [this](https://github.com/athenianco/athenian-api/pull/70/commits/c8484d150ba50ad54ad7f2d803b5fc60428440ce?file-filters%5B%5D=.py#diff-7640db81d011b0cff79abcd6886d5be6R40) would be always `None`, doesn't it?

server/athenian/api/controllers/user_controller.py
```diff
@@ -32,3 +32,23 @@
...
+async def become_user(request: AthenianWebRequest, id: str) -> web.Response:
+    """God mode ability to turn into any user. The current user must be marked internally as \
+    a super admin."""
+    user_id = getattr(request.user, "god_id", None)
+    if user_id is None:
+        return ResponseError(ForbiddenError(
+            detail="User %s is not allowed to mutate" % user_id)).response
+    async with request.sdb.connection() as conn:
+        god = await conn.fetch_one(select([God]).where(God.user_id == user_id))
+        if id and (await conn.fetch_one(
```
se7entyse7en: We can swap these two lines to avoid the query if the user doesn't exist.

server/athenian/api/auth.py
vmarkovtsev: You found a bug haha. It should be like this:
```python
god_id = user.id
user = (await self.get_users([mapped_id]))[0]
user.god_id = god_id
```

server/athenian/api/controllers/user_controller.py
vmarkovtsev: Yes, we can. But I don't want to expose an endpoint that will allow everybody to count how many users we have registered by applying binary search.

vmarkovtsev: :+1:

server/athenian/api/auth.py
vmarkovtsev: This should work like this. Let's suppose there is a super admin `vadim@athenian.co` and a regular user `marvin@athenian.co`

1. `vadim@athenian.co` logs in as usual.
2. Call `/v1/become/auth0|whatever-belongs-to-marvin@athenian.co`
3. A new record in the DB appears that maps `vadim@athenian.co` (`God.user_id`) to `marvin@athenian.co` (`God.mapped_id`).
4. Any subsequent request from `vadim@athenian.co` is first handled as normal, so Auth0 checks whether the user is `vadim@athenian.co`.
5. However, in the end, we check whether `vadim@athenian.co` is a god. If he is, we look up the mapped ID in the DB.
6. We query the mgmt Auth0 API to fetch the full profile of the mapped user - `marvin@athenian.co`.
7. We overwrite the `user` field of `request` and additionally set the extra attribute `god_id` to indicate that the user is a mapped god.
8. API handlers think that the user is `marvin@athenian.co`.
9. But `/v1/become` checks `user.god_id` and if it exists, it is used in the DB god check instead of the regular `user.id`. Thus we don't lose the ability to turn into any other user, including the empty string (`None`, the initial default unmapped state).

se7entyse7en: Got it! Thanks for the detailed explanation!

vmarkovtsev: I should include it in the DD once merged.

PR #72, opened 2020-01-30
Add bearerAuth in openapi spec for autheniticated path

se7entyse7en:
This change is required for generating the client for making it able to call with bearer token.

I tried generating the python files and on changes occurred.

Reference: https://swagger.io/docs/specification/authentication/bearer-authentication/

### DIFF
--- a/server/athenian/api/controllers/security_controller.py
+++ b/server/athenian/api/controllers/security_controller.py
@@ -1,0 +1,20 @@
+from typing import Dict
+
+from jose import jwt
+
+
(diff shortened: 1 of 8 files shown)


vmarkovtsev:
I don't think that we need Python generation here - the models will not change.

se7entyse7en:
@vmarkovtsev feel free to merge this yourself 👍

vmarkovtsev:
Yep, I need to pass the CI first 😉

PR #73, opened 2020-01-30
Revamp auth

vmarkovtsev:


vmarkovtsev:
@se7entyse7en This PR burns the bridges: either there is no `Authorization` header and we assign `gkwillie`, or there is a bearer token and it gets properly checked. No more hardcoded users.

vmarkovtsev:
cc @dpordomingo ^

PR #75, opened 2020-02-02
Admin backdoor index

se7entyse7en:
Changes default admin backdoor index to 0.

The reason I did this is that I wasn't able to test userflow handshake webapp <> api due to an error when calling `/v1/accept`. The error was a `LockedError`. The problem is that when generating the admin link with the default admin backdoor account with the "special" id, since the `id` is an autoincrement by default, the api was returning an error because was trying to create an new account with `id = admin_backdoor + 1`. Changing it to `0` (that IMHO is still special enough), fixed the problem.

Alternatives that I considered:
- adding a field in `Account` like `is_backdoor`, but it seemed to be an overkill for just this case,
- resetting the auto increment when creating the special backdoor account, but it's backend specific.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,8 @@ $(MAKEFILE):
 	git clone --quiet --depth 1 -b $(CI_BRANCH) $(CI_REPOSITORY) $(CI_PATH);
 -include $(MAKEFILE)
 
+DOCKER_RUN_EXTRA_ARGS ?= -it
+
 IO_DIR ?= $(PWD)/server/tests
 ENV_FILE ?= .env
 
@@ -29,7 +31,8 @@ $(IO_DIR)/%.sqlite: $(ENV_FILE)
 
 .PHONY: run-api
 run-api: $(IO_DIR)/mdb.sqlite $(IO_DIR)/sdb.sqlite
-	docker run --rm -p 8080:8080 -v$(IO_DIR):/io --env-file $(ENV_FILE) $(IMAGE) --ui --metadata-db=sqlite:///io/mdb.sqlite --state-db=sqlite:///io/sdb.sqlite
+	docker run $(DOCKER_RUN_EXTRA_ARGS) --rm -p 8080:8080 -v$(IO_DIR):/io --env-file $(ENV_FILE) $(IMAGE) --ui --metadata-db=sqlite:///io/mdb.sqlite --state-db=sqlite:///io/sdb.sqlite
+
 .PHONY: invitation-link
 invitation-link: $(IO_DIR)/mdb.sqlite $(IO_DIR)/sdb.sqlite
 	docker run --rm -e DB_DIR=/io -v$(IO_DIR):/io --env-file $(ENV_FILE) --entrypoint python3 $(IMAGE) -m athenian.api.invite_admin sqlite:///io/sdb.sqlite
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,9 @@ $(IO_DIR)/%.sqlite: $(ENV_FILE)
 .PHONY: run-api
 run-api: $(IO_DIR)/mdb.sqlite $(IO_DIR)/sdb.sqlite
 	docker run --rm -p 8080:8080 -v$(IO_DIR):/io --env-file $(ENV_FILE) $(IMAGE) --ui --metadata-db=sqlite:///io/mdb.sqlite --state-db=sqlite:///io/sdb.sqlite
+.PHONY: invitation-link
+invitation-link: $(IO_DIR)/mdb.sqlite $(IO_DIR)/sdb.sqlite
+	docker run --rm -e DB_DIR=/io -v$(IO_DIR):/io --env-file $(ENV_FILE) --entrypoint python3 $(IMAGE) -m athenian.api.invite_admin sqlite:///io/sdb.sqlite
 
 .PHONY: clean
 clean: fixtures-clean
--- a/server/athenian/api/__init__.py
+++ b/server/athenian/api/__init__.py
@@ -43,7 +43,9 @@ def parse_args() -> argparse.Namespace:
   AUTH0_CLIENT_ID          Client ID of the Auth0 Machine-to-Machine Application
   AUTH0_CLIENT_SECRET      Client Secret of the Auth0 Machine-to-Machine Application
   ATHENIAN_INVITATION_KEY  Passphrase to encrypt the invitation links
-  """,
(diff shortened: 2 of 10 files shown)


server/athenian/api/controllers/invitation_controller.py
```diff
@@ -106,15 +106,8 @@ def bad_req():
...
             # create a new account for the admin user
-            acc_id = await conn.execute(
-                insert(Account).values(Account().create_defaults().explode()))
-            if acc_id >= admin_backdoor:
-                await conn.execute(delete(Account).where(Account.id == acc_id))
-                return ResponseError(GenericError(
-                    type="/errors/LockedError",
-                    title=HTTPStatus.LOCKED.phrase,
-                    status=HTTPStatus.LOCKED,
-                    detail="Invitation was not found.")).response
```
se7entyse7en: I don't know exactly the reason behind this, otherwise we can re-introduce this as `if acc_id >= max_account_id`.

se7entyse7en:
@vmarkovtsev before fixing tests I'd like to know first it this makes sense to you. The code in this PR made the auth handshake with the webapp work (https://github.com/athenianco/athenian-webapp/pull/50).

vmarkovtsev:
The real bug is that there should not be a DB record with `admin_backdoor` in the first place. I will add a check to completely restrict it, but I wonder how it appeared.

Makefile
```diff
@@ -29,7 +36,8 @@ $(IO_DIR)/%.sqlite: $(ENV_FILE)
 
 .PHONY: run-api
 run-api: $(IO_DIR)/mdb.sqlite $(IO_DIR)/sdb.sqlite
-	docker run --rm -p 8080:8080 -v$(IO_DIR):/io --env-file $(ENV_FILE) $(IMAGE) --ui --metadata-db=sqlite:///io/mdb.sqlite --state-db=sqlite:///io/sdb.sqlite
+	docker run $(DOCKER_RUN_EXTRA_ARGS) --rm -p 8080:8080 -v$(IO_DIR):/io --env-file $(ENV_FILE) $(IMAGE) --ui --metadata-db=sqlite:///io/mdb.sqlite --state-db=sqlite:///io/sdb.sqlite
```
vmarkovtsev: `run-api` was supposed to execute for exclusively development debugging. Any deployment requires a separate config because the logic has many shortcuts for development. Why do we need this?

server/tests/gen_sqlite_db.py
```diff
@@ -17,22 +19,25 @@ def fixture(*args, **kwargs):
 from tests.controllers.conftest import db_dir, metadata_db
 from tests.sample_db_data import fill_state_session
 
+parser = argparse.ArgumentParser(description='Create databases.')
+parser.add_argument('--fill-state-session', action='store_true',
```
vmarkovtsev: This should false by default because many places depend on the old behavior.

vmarkovtsev:
Now I remember: this is a result of FOREIGN_KEY constraint. OK, now I know why and need to think about how to fix...

Makefile
se7entyse7en: Yup, this is for development only. I used this for debugging after putting some `pdb.set_trace()` in the code. Even if it required to rebuild the docker image, the time for rebuilding is really small.

server/tests/gen_sqlite_db.py
se7entyse7en: Since I didn't find it being used elsewhere other than the `Makefile` I thought it was ok to change it, but maybe I missed some other places. Gonna change it.

se7entyse7en:
> Now I remember: this is a result of FOREIGN_KEY constraint.
You mean primary key? Because AFAIS the issue is the to the `id` being a primary key making it auto increment. 
 
> OK, now I know why and need to think about how to fix...
👍 
Feel free to close this, the main purpose was to have a working version of the api to make the webapp work.

server/athenian/api/__init__.py
```diff
@@ -43,7 +43,9 @@ class Formatter(argparse.ArgumentDefaultsHelpFormatter, argparse.RawTextHelpForm
   AUTH0_CLIENT_ID          Client ID of the Auth0 Machine-to-Machine Application
   AUTH0_CLIENT_SECRET      Client Secret of the Auth0 Machine-to-Machine Application
   ATHENIAN_INVITATION_KEY  Passphrase to encrypt the invitation links
-  """,
+  ATHENIAN_INVITATION_URL_PREFIX
+                           String with which any invitation URL starts, e.g. https://app.athenian.co/i/
+  """,  # noqa
```
vmarkovtsev: This is starting to look like a missed configuration file. We should consider putting these to a file in the next version.

server/athenian/api/invite_admin.py
```diff
@@ -15,14 +15,30 @@ def main():
...
         session.add(Account(id=invitation_controller.admin_backdoor))
         session.commit()
+        max_id = session.query(func.max(Account.id)).filter(Account.id < admin_backdoor).first()
+        if max_id is None or max_id[0] is None:
+            max_id = 0
+        else:
+            max_id = max_id[0]
+        max_id += 1
+        if engine.url.drivername == "postgres":
+            engine.execute("ALTER SEQUENCE accounts_id_seq RESTART WITH %d;" % max_id)
```
vmarkovtsev: @se7entyse7en This is the fix - I tested with both sqlite and postgres, and I learned something new about AUTOINCREMENT :D SQLite has primary keys autoincremented by default, but the counter values are calculated dynamically as `MAX(id)`. This is not what we want, so I had to force SQLite `AUTOINCREMENT` keyword feature, hence `sqlite_autoincrement=True`. Postgres does not have this quirk so it was easier to deal with.

server/athenian/api/__init__.py
se7entyse7en: I agree with you, reading this configurations from a file will make things easier also from infra pov. Regarding the value, for now it should be `https://owl.athenian.co/i/`, for local development by default it should be `http://localhost:3000/i/`.

server/athenian/api/invite_admin.py
se7entyse7en: Thanks!

server/athenian/api/__init__.py
se7entyse7en: nvm let's read it from env var.

se7entyse7en:
Okay, I saw you added `postgresql` as well. Feel free to merge 👍, thanks!

PR #79, opened 2020-02-04
Add /filter/repositories and /filter/contributors

vmarkovtsev:


server/athenian/api/openapi/openapi.yaml
```diff
@@ -978,3 +1033,29 @@ components:
...
       - values
+    FilterItemsRequest:
+      example:
+        account: 1
+        date_from: 2000-01-23
+        date_to: 2020-05-23
+        in:
+        - '{1}'
+      properties:
+        account:
```
dpordomingo: Since our tool is analyzing only one `RepositorySet` at the same time, why filtering by `account`, &mdash;which can contain many `RepositorySets`&mdash;, instead of filtering by only one `RepositorySet`?

vmarkovtsev: The request may contain arbitrary reposets and individual repositories, and we need to check whether they are (1) make sense together, that is, we don't mix reposets from different accounts together and (2) they are accessible from the user's account, e.g. when somebody specifies src-d/go-git in athenian dog fooding or a random reposet ID belonging to the account where the user is not included.

dpordomingo: But, is that a feature?
>_The request may contain arbitrary reposets and individual repositories_

Because I don't know which use case requires it.

Also, would the current implementation avoid a user getting the contributors of `go-git` (not being that repo of any reposet of any of their accounts), using this payload?
```
{
  account: 1, // the user belongs to this account
  in: ["go-git"]
}
```

Because with the proposed implementation, it would not leak contributors from random repos.
```
{
  reposet: 33, // the user can access this reposet
  in: ["go-git"] // it will avoid getting info from go-git because it does not belong to reposet#33
}
```

vmarkovtsev: > I don't know which use case requires it.
Teams.

> avoid a user getting the contributors

The current implementation avoids what you wrote, but that specific check is not implemented yet :smile: Specifying the account provides the room for implementing that check. However, the reposets are already validated properly.

Anyway, you should expect that all API filtered calls require to specify the account. This is our designed arch of having accounts.

PR #86, opened 2020-02-06
Remove the hardcoded gkwillie

vmarkovtsev:


se7entyse7en:
I guess that this requires setting the value on infra side \cc @lwsanty

se7entyse7en:
But why it's mandatory?

vmarkovtsev:
Olek has already configured the deployments. This is just a matter of cleaner configuration.

PR #90, opened 2020-02-06
Fix rebuilding the DBs if .env gets updated

vmarkovtsev:


Makefile
```diff
@@ -30,6 +30,7 @@ $(ENV_FILE):
 
 $(IO_DIR)/%.sqlite: $(ENV_FILE)
 	docker run --rm -e DB_DIR=/io -v$(IO_DIR):/io --env-file $(ENV_FILE) --entrypoint python3 $(IMAGE) /server/tests/gen_sqlite_db.py
+	@touch $@
```
dpordomingo: The database is created wit `read-only` attributes, so `touch` fails.

vmarkovtsev: WDYM? It works fine for me.

vmarkovtsev: Is it about the owner user?

vmarkovtsev: How about now?

dpordomingo: It did! Many thanks!

PR #91, opened 2020-02-07
Include useful tags in Sentry context

vmarkovtsev:


dpordomingo:
Imo we could also now remove the obligation of bumping the `__version__` on Evey commit.

vmarkovtsev:
Knowing myself, I will stop bumping the version at all then.

PR #106, opened 2020-02-17
ci -> change condition on deploy

lwsanty:
related to https://github.com/athenianco/ci/pull/4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ branches:
 env:
   global:
     - GOOGLE_DOCKER_IMAGE=gcr.io/athenian-1/api
-    - IMAGE_BRANCH_COMMIT=${GOOGLE_DOCKER_IMAGE}:${TRAVIS_BRANCH}-${TRAVIS_COMMIT:0:7}
+    - IMAGE_BRANCH_COMMIT=${GOOGLE_DOCKER_IMAGE}:${TRAVIS_COMMIT:0:7}
 
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -63,7 +63,7 @@ matrix:
         - set +e
       script: skip
     - stage: deploy
-      if: branch = master AND type != pull_request AND commit_message =~ /^Bump version/
+      if: branch IN (master, production) AND type != pull_request AND commit_message =~ /^Bump version/
```
vmarkovtsev: IDK if it will work as you expect though. The version bump commit is created in the master branch automatically. Perhaps you should have two OR-ed conditions that nstead.

lwsanty: I probably did not get the problem with this condition and a bump commit

just in case, docs quote
https://docs.travis-ci.com/user/conditions-v1
>Attribute in lists:
>
>branch IN (one, other)

se7entyse7en: Here are the assumptions that I'm making:
- We want to deploy in staging/production/whatever only things that correspond to a bump,
- We want to deploy in production only things that have been already deployed in staging,
- `production` is always an ancestor of `master`.

This means that any docker image that we want to ship on production, is actually already built as it has been already previously built for staging (`master` branch), and this should be true for all projects actually.

But currently we have the image name that depends on the branch being built: `IMAGE_BRANCH_COMMIT=${GOOGLE_DOCKER_IMAGE}:${TRAVIS_BRANCH}-${TRAVIS_COMMIT:0:7}`. Hence, if on `production` we don't rebuild the docker image, that image is not found.

If we make the `IMAGE_BRANCH_COMMIT` target branch agnostic, so for example `IMAGE_BRANCH_COMMIT=${GOOGLE_DOCKER_IMAGE}:${TRAVIS_COMMIT:0:7}`, we can actually just build the image only when on `master`, and on `production` only do the pubsub-publish.

I don't mind doing these in second step, for now we could just build always.

se7entyse7en: So, if we make `production` point to any forward commit correspdondoing to a bump, this should work as expected. It will just rebuild the image even if it's unnecessary.

lwsanty: moreover, we have a build cache, so in this case theoretically we don't even need to skip building because it will be done from the cache, so theoretically we can just change
```
IMAGE_BRANCH_COMMIT=${GOOGLE_DOCKER_IMAGE}:${TRAVIS_BRANCH}-${TRAVIS_COMMIT:0:7}
```
to
```
IMAGE_BRANCH_COMMIT=${GOOGLE_DOCKER_IMAGE}:master-${TRAVIS_COMMIT:0:7}
```
and it will cause `image up to date` on push

@se7entyse7en @vmarkovtsev wdyt?

se7entyse7en: Makes sense to me, let' wait for @vmarkovtsev as well.

se7entyse7en: I'd prefer to have the version number itself, but it's so ok for now IMO.

vmarkovtsev: Yeah, let's have the version number because it is easier to interpret now that we've got automatic version bumps.

```
version_file=server/athenian/api/metadata.py
current_version=$(grep __version__ $version_file | cut -d\" -f2)
```

se7entyse7en: LGTM 👍

.travis.yml
```diff
@@ -9,7 +9,7 @@ branches:
 env:
   global:
     - GOOGLE_DOCKER_IMAGE=gcr.io/athenian-1/api
-    - IMAGE_BRANCH_COMMIT=${GOOGLE_DOCKER_IMAGE}:${TRAVIS_BRANCH}-${TRAVIS_COMMIT:0:7}
+    - IMAGE_BRANCH_COMMIT=${GOOGLE_DOCKER_IMAGE}:master-${TRAVIS_COMMIT:0:7}
```
vmarkovtsev: Do we need to mention `master` at all, after all?

.travis.yml
```diff
@@ -64,15 +64,26 @@ matrix:
...
+        docker tag ${IMAGE} ${GOOGLE_DOCKER_IMAGE}
+        IMAGE=${GOOGLE_DOCKER_IMAGE} make docker-push
+    - stage: deploy-production
+      if: branch = production AND type != pull_request AND commit_message =~ /^Bump version/
+      script: |
+        make gcloud-login
+        VERSION_FILE=server/athenian/api/metadata.py
+        CURRENT_VERSION=$(grep __version__ $version_file | cut -d\" -f2)
+        IMAGE=${IMAGE_BRANCH_COMMIT}-${CURRENT_VERSION}
+        make pubsub-publish
```
vmarkovtsev: How is it different from the regular `deploy`? I mean, both do `make pubsub-publish`, so will this always be a no-op?

lwsanty: @vmarkovtsev do you propose to put them in one job?

lwsanty: Done

PR #129, opened 2020-02-25
Generate the new models and rename the old ones

vmarkovtsev:


vmarkovtsev:
@dpordomingo @se7entyse7en This is the upcoming API schema changes. Please review before I started implementing the endpoints.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1435,3 +1513,327 @@ components:
...
+              timestamp: 2020-02-25T14:27:21Z
+            committer:
+              login: github.com/gkwillie
+              name: Groundskeeper Willie
+              email: bot@athenian.co
+              timestamp: 2020-02-25T14:27:21Z
+              message: 'Bump version: 0.0.42 → 0.0.43'
+              size_added: 1
+              size_removed: 1
+              files_changed: 1
```
dpordomingo: Why are these fields under `committer` instead of being direct children of the `commit`.
imo it should be one indentation less

server/athenian/api/openapi/openapi.yaml
```diff
@@ -310,9 +344,51 @@ paths:
...
+      - bearerAuth: []
+      summary: Measure the amount of code that was pushed outside of pull requests.
+      tags:
+        - metrics
+        # {% if False %}
+        - default
+      # {% endif %}
+      x-openapi-router-controller: athenian.api.controllers.metrics_controller
+      x-codegen-request-body-name: body
+  /metrics/pr_linear:
```
dpordomingo: What's the meaning of this new name? Why not `/metrics/prs`?

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1435,3 +1513,327 @@ components:
...
+            $ref: '#/components/schemas/CommitGroup'
+          type: array
+        include:
+          $ref: '#/components/schemas/IncludedNativeUsers'
+      required:
+      - data
+      - include
+      type: object
+    CommitGroup:
+      description: List of commits that were committed starting from a specific time
```
dpordomingo: In order to make it more flexible, I'd not mention the interval not the granularity, which is something about the query using this entity, which can use an interval, or not.
```suggestion
      description: List of commits
```

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1435,3 +1513,327 @@ components:
...
+            name: Groundskeeper Willie
+            email: bot@athenian.co
+            timestamp: 2020-02-25T14:27:21Z
+            message: 'Bump version: 0.0.42 → 0.0.43'
+            size_added: 1
+            size_removed: 1
+            files_changed: 1
+      properties:
+        date:
+          description: Commits were pushed beginning with this date. They were not
```
dpordomingo: _ditto_

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1435,3 +1513,327 @@ components:
...
+          timestamp: 2020-02-25T14:27:21Z
+        committer:
+          login: github.com/gkwillie
+          name: Groundskeeper Willie
+          email: bot@athenian.co
+          timestamp: 2020-02-25T14:27:21Z
+          message: 'Bump version: 0.0.42 → 0.0.43'
+          size_added: 1
+          size_removed: 1
+          files_changed: 1
```
dpordomingo: this should be one indentation less

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1435,3 +1513,327 @@ components:
...
+            timestamp: 2020-02-25T14:27:21Z
+          committer:
+            login: github.com/gkwillie
+            name: Groundskeeper Willie
+            email: bot@athenian.co
+            timestamp: 2020-02-25T14:27:21Z
+            message: 'Bump version: 0.0.42 → 0.0.43'
+            size_added: 1
+            size_removed: 1
+            files_changed: 1
```
dpordomingo: imo it should be one indentation less

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1357,7 +1434,8 @@ components:
           description: Number of commits in this PR.
           type: integer
         review_requested:
-          description: Value indicating whether the author of this PR requested a review.
+          description: Value indicating whether the author of this PR requested a
```
dpordomingo: Only the author?
What if any other contributor requested a review? is it not considered for the `review_requested`?
Imo it should not depend on the authorship of the request.
(it could be made automatically by a bot, assigning a reviewer if the PR meets some requirements)

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1435,3 +1513,327 @@ components:
...
+            hash: 79f0bccb856b35ef08ec93da2ff5585da4d2e50e
+            author:
+              login: github.com/gkwillie
+              name: Groundskeeper Willie
+              email: bot@athenian.co
+              timestamp: 2020-02-25T14:27:21Z
+            committer:
+              login: github.com/gkwillie
+              name: Groundskeeper Willie
+              email: bot@athenian.co
```
dpordomingo: Why not add this repeated fields into `included` fields as done with the user `avatar` at #107?

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1435,3 +1513,327 @@ components:
...
+        bypassed_commits: 2
+        bypassed_lines: 30
+        total_commits: 3
+        total_lines: 60
+      items:
+        $ref: '#/components/schemas/CodeBypassingPRsMeasurement'
+      type: array
+    CodeBypassingPRsMeasurement:
+      description: Statistics about code pushed outside of pull requests in a
+        certain time interval.
```
dpordomingo: _ditto_
it is not needed to mention the time interval since it depends on the query building it

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1435,3 +1513,327 @@ components:
...
+      example:
+        date: 2020-01-23
+        bypassed_commits: 1
+        bypassed_lines: 20
+        total_commits: 2
+        total_lines: 50
+      properties:
+        date:
+          description: Commits were pushed beginning with this date. They were not
+            pushed later than +granularity.
```
dpordomingo: _ditto_
it is not needed to mention the granularity, because it depends on the query

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1435,3 +1513,327 @@ components:
...
+          property:
+            description: Main trait of the commits - the core of the filter.
+            enum:
+            - bypassing_prs
+            type: string
+        required:
+        - property
+        type: object
+    CommitsGroupedByTime:
+      description: Lists of commits for each time interval.
```
dpordomingo: _ditto_
```suggestion
      description: Lists of commits grouped by date
```

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1435,3 +1513,327 @@ components:
...
+        timestamp:
+          description: When the corresponding action happened.
+          format: date-time
+          type: string
+      required:
+      - email
+      - name
+      - timestamp
+      type: object
+    CommitFilter:
```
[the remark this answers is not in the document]
vmarkovtsev: It has some unique props. I will refactor the common filter core later.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: F*cking copy-paste in PyCharm

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: Well, the original idea was to provide the "linear" metric values. At this stage, it has mutated into something PR-specific, so yes, we can make it `/metrics/prs`.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: Hehe

vmarkovtsev: Let's keep it laser-focused for now and abstract later. It is hard to foresee what will be the future endpoints now. Better very specific now than refactoring crappy abstractions later.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: OK, I forgot about it :+1:

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: Some commits do not have mapped usernames. I can do what you propose if you are fine with all the three props being optional.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: As soon as we have a query that re-uses this model and this description becomes irrelevant, I will remove it.

server/athenian/api/openapi/openapi.yaml
dpordomingo: I already to be aware of a user not having an avatar, so the same about name or email.
Anyhow, what do you mean with "three props"? I'm only talking about `name` and `email`, since `login` would be the key to get the included user, and `timestamp` is unique per commit.

server/athenian/api/openapi/openapi.yaml
dpordomingo: I don't see the benefit of keeping some docs that are not strictly true (because in the context of just `CommitGroup`, where this description is shown, the date interval and granularity has no meaning at all), and it can be easily outdated in the future as soon as other endpoint starts serving `CommitGroup`.

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
dpordomingo: Could you at least use 
```yml
    MetricsRequest:
      properties:
        for:
            $ref: '#/components/schemas/ForSet'
```
Instead of:
```yml
    MetricsRequest:
      properties:
        in:
        with_author:
          $ref: '#/components/schemas/DeveloperSet'
        with_committer:
          $ref: '#/components/schemas/DeveloperSet'
```
It would be consistent with the request sent by `/metrics/prs`

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: There are commits outside of the GitHub userbase.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: The key word is as soon as. It is strictly true right now, and *as soon as* starts serving, I will address this issue. Don't invent problems and KISS. I can rename the model to be CommitGroupFor FilterCommitsRequest just to stop the intimidating, endless arguments with you :sweat:

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
vmarkovtsev: MetricsRequest is not filtered by commit author and commit committer, so I cannot do it. As I said, I plan to extract the common core for both filter models later, don't worry.

server/athenian/api/openapi/openapi.yaml
dpordomingo: Would that mean that in some cases there would be users with only `login` and under included there would be its `name` and `email`, and for other committers, there will be no `login` but only `name` and `email`, and nothing in the included?
If that's what you mean, lgtm

server/athenian/api/openapi/openapi.yaml
dpordomingo: I have a strong opinion about this, but not strong enough to be considered as blocker this time. So yes, I do believe in your capacity for handling this in the best way. Let's continue as you're proposing.

dpordomingo: I mean, it's not needed to rename `FilterCommitRequest`, which is actually ok.

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
dpordomingo: That's my point `MetricsRequest` could have been filtered by `commit_author`, `commit_commiter`, `comment_author`, `reviewer`, `merger`, `creator`, but we decided to filter only by `developers` (consider them as "participants"), and let the API do an `OR` with all the different kind of participation in the PR.

In this case, I'm proposing the same. Since there is no necessity on filtering by commit authorship, let's avoid overcomplicating the facade of the API, offering different ways to filter based on the kind of commit authorship.

se7entyse7en:
@vmarkovtsev what's the state of this? Given that @dennwc is working in providing the views, is this still relevant in the short-term? AFAIU we can minimize the changes by doing it two steps:
1. use views emulating previous schema,
2. migrate api to use new schema after v1.

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
vmarkovtsev: `MetricsRequest` is for pull requests, it is not related to `commit_author` or `commit_commiter`. I will rename it to `PullRequestMetricsRequest` to avoid confusion.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: Yes, exactly, you got it :+1:

vmarkovtsev:
Yes, I also think that it makes sense to merge the schema changes now and implement the endpoints later. I am addressing the last David's concerns and merging.

cc/ @se7entyse7en

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: Renamed to avoid confusion.

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
dpordomingo: we could also filter PRs by kind of participations, PRs are really related to commits and comments.

server/athenian/api/openapi/openapi.yaml
dpordomingo: did you finally added `name` and `email` into de included fields?

vmarkovtsev: Now everything is optional, as we 🤝.

dpordomingo: Optional is one thing. Moving  `name` and `email` into de included fields another. This last one is a good way to save a lot of bandwidth as done with user avatars.

vmarkovtsev: Once we need it, I will surely list them, yes.

PR #138, opened 2020-03-04
Test JS API Client autogeneration

dpordomingo:
This PR will introduce into Travis build a test to ensure that the autogenerated JS API client won't fail, which can happen as demoed by https://github.com/dpordomingo/athenian-api/pull/4; if that case happens, it will be raised a red flag in Travis [like this one](https://travis-ci.com/dpordomingo/athenian-api/jobs/293797859#L488).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,13 @@ matrix:
         - set +e
     - if: commit_message !~ /Bump version/
       stage: main
+      name: Auto generated js client tests
+      script:
+        - docker run --rm -v `pwd`:/local --user `id -u`:`id -g` openapitools/openapi-generator-cli generate --input-spec=/local/spec/openapi.yaml --generator-name=javascript --output=/local/js-client
+        - npm --prefix js-client install
+        - npm --prefix js-client test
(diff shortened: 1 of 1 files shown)


server/athenian/api/openapi/openapi.yaml
```diff
@@ -1314,9 +1314,7 @@ components:
         values:
           description: Calculated metric values in the same order as `metrics`.
           items:
-# {% if False %}
-            allOf:
-# {% else %}{{ "\n" }}            oneOf:{% endif %}
+            oneOf:
```
vmarkovtsev: Is this fix no longer required?

dpordomingo: It is **required**.

I just created the PR from the wrong branch, so you saw here what should be at https://github.com/dpordomingo/athenian-api/pull/4 in order to provide an example of the failing case. Sorry for letting you see an incomplete PR.

dpordomingo:
superseded by https://github.com/athenianco/athenian-api/pull/139

PR #157, opened 2020-03-17
ci -> tweaks

lwsanty:
1) restrict deploy on push to master branch
2) remove commit hash from docker image tag

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ branches:
 env:
   global:
     - GOOGLE_DOCKER_IMAGE=gcr.io/athenian-1/api
-    - IMAGE_BRANCH_COMMIT=${GOOGLE_DOCKER_IMAGE}:${TRAVIS_COMMIT:0:7}
 
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -91,16 +90,15 @@ matrix:
         - git remote add gkwillie https://gkwillie:${GKWILLIE_TOKEN}@github.com/$TRAVIS_REPO_SLUG
         - git push --tags gkwillie HEAD:master
         - set +e
-    - if: branch IN (master, production) AND type != pull_request AND commit_message =~ /^Bump version/
+    - if: branch = master AND type != pull_request AND commit_message =~ /^Bump version/
```
vmarkovtsev: So there is no more `production` branch, correct?

lwsanty: from [[ENG-275]] we will have pinned services versions on prod, so since that issue is done we will deploy on prod only the images that are mentioned in specific config files, that means that trying to update the service on production from production branch is still possible but not recommended.

PR #158, opened 2020-03-17
Releases

vmarkovtsev:


server/athenian/api/openapi/openapi.yaml
```diff
@@ -1587,8 +1585,17 @@ components:
           description: Number of review comments this PR received.
           type: integer
         merged:
-          description: Value indicating whether this PR was merged.
-          type: boolean
+          description: When this PR was merged.
+          format: date-time
+          type: string
```
dpordomingo: What's the value for a not merged PR? an empty string?

vmarkovtsev: This field will be absent from the response.

PR #171, opened 2020-03-24
Implement /metrics/developers

vmarkovtsev:


eiso:
Loving this ❤️! Crazy speed you had putting this together

dpordomingo:
Shouldn't this replace `/filter/contributors`? You can get in that other endpoint a subset of the info you can get from this new one. Then, I'd mark the older one as `deprecated`

vmarkovtsev:
Nope. The purpose is different (filter vs metrics), the implementation is very different and the performance is also very different (`contributors` is much faster). Don't get dilusioned: numbers that are returned in contributors are returned just because they are calculated anyway, they are a very limited subset of metrics and are needed to sort people in the list, nothing more.

PR #176, opened 2020-03-25
Fix first account creation

se7entyse7en:
I bumped into this by using an empty `state` db and using the admin invitation link.


vmarkovtsev:
@se7entyse7en Can you please quickly adda test for this in `test_invitation_controller.py`?

```python
from sqlalchemy import delete

async def test_empty_db_account_creation(sdb):
    await sdb.execute(delete([Account]))
    ...
```

se7entyse7en:
@vmarkovtsev test added 👍

PR #192, opened 2020-04-01
Release match settings

vmarkovtsev:


dpordomingo:
Since I found no description I had to do an extra effort to understand the logic of this PR; let's see If I did it properly:

Releases returned by the API will be defined by a `ReleaseMatchSetting` per account and repository, stored in the DB; If there is no `ReleaseMatchSetting` for any repo, it will be used a default. I guess that the default will be _"releases will be defined by tag"_, won't it?.

Webapp can fetch `ReleaseMatchSettings` by calling `GET::/settings/release_match/{account_id}`, getting the explicitelly defined `ReleaseMatchSetting` of each repo of that account. Defaults won't be returned.

Then, webapp can modify the stored `ReleaseMatchSettings` by calling `PUT::/settings/release_match` sending a `ReleaseMatchRequest`, containing:
- `repositories`: `RepositorySet` of repositories to be affected,
- `match`: `ReleaseMatchStrategy` that can be either `branch` or `tag`, whether releases are read from branches or tags,
- `branches` and `tags`, defining the regular expression that will capture the releases based on the match strategy.

What it's not clear for me is if there is any kind of constraint for `ReleaseMatchStrategy`, `branches` and `tags`, e.g. can you define the following?
```
{ // this should fail imo
  match: branch,
  tag: v.*
}
```
or
```
{ // only 'v.*' tags that are reachable from 'production' branch
  match: tag,
  tag: v.*
  branch: production
}
```

Then... why not using a single git reference, instead of the combination of `ReleaseMatchStrategy plus `branches` plus `tags`, ie:
- `match`: `refs/tags/.*`, or `refs/heads/.*`, eg: 
  ```
  { // only tags following semantic versioning
    match: refs/tags/v\d+\.\d+\.\d+$
  }
  ```
  or  
  ```
  { // merges into 'production' branch
    match: refs/heads/production$
  }
  ```

vmarkovtsev:
All is correct David!

> Defaults won't be returned.

Whatever you prefer here, I can return them or not.

```
{ // this should fail imo
  match: branch,
  tags: v.*
}
```
This fails, yes. It is hard to declare in YAML so the check is done in the code.

```
{ // only 'v.*' tags that are reachable from 'production' branch
  match: tag,
  tags: v.*
  branches: production
}
```
This will work and the `tags` will be written in the DB even though it won't be used.

BTW do you prefer `tags` and `branches` or `tag` and `branch`  as property names?

> Then... why not using a single git reference

The underlying machinery does not deal with Git directly. There is a special, unique table for tags and a special, unique table for branches. The matching algorithm itself is complex as hell, it is way different for tags and branches, and has nothing to do with how Git references work. So such a generic way will never work with GitHub API.

vmarkovtsev:
The current behavior is to return the defaults.

vmarkovtsev:
@rpau This is https://athenianco.atlassian.net/browse/ENG-425

dpordomingo:
I'm ok with not returning defaults from `GET::/settings/release_match/{account_id}` because otherwise, the account admin wouldn't know that it was used a default. I think it's better to let the admin know that some repos are using a custom `ReleaseMatchSetting`, and list the others that are using current default (by tag), so the admin will be able to add new `ReleaseMatchSetting` if needed.

About the `ReleaseMatchSetting`, please remove the `match ReleaseMatchStrategy` property. Imo it should be enough (and way less error-prone) with:
```json
{
 tags string
 branches string 
}
```

About using singular or plurals, both lgtm.

About using git references, it's not a strong requirement, but it could be used even not using git. We can just use a regexp to guess if the user is targeting tags, or branches. But up to you.

vmarkovtsev:
@dpordomingo 

```
{
 tags string
 branches string 
}
```

confuses between `branch` and `tag_or_branch` and between `tag` and `tag_or_branch`. We've got three strategies now so have to leave it. We'll have a fourth one in the future (web hooks).

dpordomingo:
`tag_or_branch` could be the default if none is set. Idk about how web hooks will play here, but I wonder why not just doing:
```go
{
 tags string
 branches string 
 webhook string
}
```

vmarkovtsev:
explicit is always better than implicit. Merging this.

dpordomingo:
Default values are going to be always there, and are always implicit by definition.

vmarkovtsev:
Sorry, no time for arguing, I need to ship a solution before tomorrow night.

PR #195, opened 2020-04-03
Provide release settings to the releases matcher

vmarkovtsev:


server/tests/test_auth.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, whitelist, cache=None):
         "picture": "https://s.gravatar.com/avatar/dfe23533b671f82d2932e713b0477c75?s=480&r=pg&d=https%3A%2F%2Fcdn.auth0.com%2Favatars%2Fei.png",  # noqa
```
eiso, line 33: I hope that picture does not get updated ;)

vmarkovtsev: Probably it will - it is sometimes taken directly from Auth0. Let's see :)

PR #290, opened 2020-04-28
Remove the missed bits of PR stages in /filter

vmarkovtsev:
ENG-720


vmarkovtsev:
@se7entyse7en I cannot merge it until the webapp is updated to send `properties` instead of `stages` in `/filter/pull_requests`.

PR #301, opened 2020-04-29
Fix imports in setup.py

se7entyse7en:
I wasn't able to build the cython extension without this imports ordering. See: https://stackoverflow.com/questions/21594925/error-each-element-of-ext-modules-option-must-be-an-extension-instance-or-2-t

I built it by running:
```
python setup.py build_ext --inplace
```

as explained [here](http://docs.cython.org/en/latest/src/quickstart/build.html#building-a-cython-module-using-setuptools).

### DIFF
--- a/server/setup.py
+++ b/server/setup.py
@@ -1,8 +1,11 @@
 from importlib.machinery import SourceFileLoader
 import os
 
-from Cython.Build import cythonize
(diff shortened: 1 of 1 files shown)


vmarkovtsev:
Whoa, I did not know this.

PR #311, opened 2020-05-03
Running api with dump of real data

se7entyse7en:
Closes [ENG-389](https://athenianco.atlassian.net/browse/ENG-389).

This PR aims to provide an easy way to run the api with a dump of the real data (`staging` or `production`). This is done with docker compose with 4 containers:
- api
- Memcached (for caching)
- postgres (for `state`, `precomputed` and `metadata` dabases)
- cloud SQL proxy (for connecting to the live environment and dump the databases)

The requirements are:
- docker compose,
- a service account with the corresponding json credentials file to connect to the CloudSQL database,
- the credentials for the CloudSQL databases.

See [here](https://github.com/athenianco/athenian-api/compare/local-development?expand=1#diff-7d62b4cde47a08800e3b7212ca2dd397R145) for more details.

## Video example

https://streamable.com/fu7abs

## Notes

Running the api in the container seems much more slower, at least on OSX, but I didn't investigate. Also, api can still be run locally natively and use only the services offered by docker compose.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,54 @@
+version: "3.7"
+services:
+  api:
+    image: ${API_IMAGE:-gcr.io/athenian-1/api:latest}
+    restart: unless-stopped
+    entrypoint: ['/bin/sh']
+    command: >
+      -c "python3 -m athenian.api.models.state \
+            postgresql://${POSTGRES_USER:-api}:${POSTGRES_PASSWORD:-api}@postgres:5432/state &&
+          python3 -m athenian.api \
+            --log-level=${API_LOG_LEVEL:-INFO} \
+            --log-structured \
+            --host=0.0.0.0 \
+            --port=8080 \
+            --metadata-db=postgresql://${POSTGRES_USER:-api}:${POSTGRES_PASSWORD:-api}@postgres:5432/metadata \
+            --state-db=postgresql://${POSTGRES_USER:-api}:${POSTGRES_PASSWORD:-api}@postgres:5432/state \
+            --precomputed-db=postgresql://${POSTGRES_USER:-api}:${POSTGRES_PASSWORD:-api}@postgres:5432/precomputed \
+            --ui"
+    env_file: .env
+    ports:
+      - ${API_HOST_PORT:-8080}:8080
+    depends_on:
+      - postgres
+      - memcached
+
+  postgres:
+    image: postgres:9.6-alpine
+    restart: unless-stopped
+    environment:
+      POSTGRES_USER: ${POSTGRES_USER:-api}
+      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-api}
+    volumes:
+      - postgres:/var/lib/postgresql/data
+      - db_dumps:/db_dumps
+
+  memcached:
+    image: launcher.gcr.io/google/memcached1
+    command: ['memcached', '-m', '64']
+
+  cloud_sql_proxy:
+    image: gcr.io/cloudsql-docker/gce-proxy:1.14
+    command: [
+      '/cloud_sql_proxy',
+      '-instances=${CLOUD_SQL_STAGING_INSTANCE}=tcp:0.0.0.0:5432,${CLOUD_SQL_PRODUCTION_INSTANCE}=tcp:0.0.0.0:5433',
+      '-credential_file=/secrets/cloudsql/credentials.json'
+    ]
+    volumes:
+      - ${CLOUD_SQL_PROXY_CREDENTIALS_FILE:-./credentials.json}:/secrets/cloudsql/credentials.json
+
+volumes:
+  postgres:
+    external: false
+  db_dumps:
+    external: false
--- a/server/README.md
+++ b/server/README.md
@@ -95,7 +95,7 @@ Generate sample SQLite metadata and state databases:
 
 ```
 docker run --rm -e DB_DIR=/io -v$(pwd):/io --entrypoint python3 athenian/api /server/tests/gen_sqlite_db.py
-``` 
+```
 
 You should have two SQLite files in `$(pwd)`: `mdb.sqlite` and `sdb.sqlite`. The third DB `--precomputed-db` can be set to any (already existing or not) `.sqlite` file or `sqlite://`.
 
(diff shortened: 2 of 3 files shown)


se7entyse7en:
> a service account with the corresponding json credentials file to connect to the CloudSQL database,

@lwsanty regarding this I used my own personal service account, maybe we could have a service account for this use case specifically to be used by everyone?

> the password for the CloudSQL databases.

Also for this, now the script uses `{env_name}-cloud-sql` user for dumping the databases. Maybe we could have another user with read-only rights.

docker-compose.yaml
```diff
@@ -0,0 +1,56 @@
+version: "3.7"
+services:
+  api:
+    image: ${API_IMAGE:-gcr.io/athenian-1/api:latest}
+    restart: unless-stopped
+    entrypoint: ['/bin/sh']
+    command: >
+      -c "python3 -m athenian.api.models.state \
+            postgresql://${POSTGRES_USER:-api}:${POSTGRES_PASSWORD:-api}@postgres:5432/state &&
+          python3 -m athenian.api.invite_admin \
```
vmarkovtsev, line 10: Why do we make another admin invitation if we are using the staging dumps?

load_postgres_data.py
```diff
@@ -0,0 +1,74 @@
+import os
```
vmarkovtsev, line 1: We should put this somewhere better than the root. I suggest `server/athenian/api/tests`

load_postgres_data.py
```diff
@@ -0,0 +1,74 @@
...
+import sys
+import time
+import subprocess
+import signal
+import tempfile
+import getpass
+
+conf = {
+    'staging': {
```
vmarkovtsev, line 10: This will not pass flake8. BTW as soon as we put it under `server/athenian/api` you'll notice this :)

load_postgres_data.py
```diff
@@ -0,0 +1,74 @@
...
+            "-Fc --dbname={db} > {out_path}"
+        ).format(port=cloud_sql_proxy_port, user=cloud_sql_db_user,
+                 db=db, out_path=dump_path)
+        print("Dumping database: {cmd}".format(cmd=guest_cmd))
+        run_pg_cmd_on_guest(guest_cmd, cloud_sql_db_password)
+        print("Done!")
+
+        guest_cmd = "createdb -U {user} {db}".format(
+            user=api_db_docker_user, db=db)
+        print("Creating database: {cmd}".format(cmd=guest_cmd))
```
vmarkovtsev, line 45: We don't use `format()` in our Python projects. Alternatives: f-strings or `%`.

load_postgres_data.py
```diff
@@ -0,0 +1,74 @@
...
+
+        guest_cmd = "pg_restore -x -Fc -O -U {user} -d {db} {in_path}".format(
+            user=api_db_docker_user, db=db, in_path=dump_path)
+        print("Loading dump into container: {cmd}".format(cmd=guest_cmd))
+        run_pg_cmd_on_guest(guest_cmd, cloud_sql_db_password)
+        print("Done!")
+
+
+if __name__ == '__main__':
+    if len(sys.argv) < 2:
```
vmarkovtsev, line 56: Can you please move all the code below to another function and then `exit(main())`.

load_postgres_data.py
```diff
@@ -0,0 +1,74 @@
...
+        print("Missing environment name")
+        exit(1)
+
+    environment = sys.argv[1]
+    if environment not in conf:
```
vmarkovtsev, line 61: I think that a trivial `parse_args` is better at this point. E.g. https://github.com/athenianco/athenian-api/blob/master/server/athenian/api/hacks/heat_cache.py#L17

load_postgres_data.py
```diff
@@ -0,0 +1,74 @@
...
+        print("Unknown environment")
+        exit(2)
```
vmarkovtsev, line 63: Thus we will have `return 2` in the function

server/README.md
```diff
@@ -142,6 +142,65 @@ You can erase the API data fixtures created by `make run-api` with:
 make clean
 ```
 
+## Running the API server locally with real data
+
+You can also run the api server and all the other services with real data using docker compose. For this setup you'll need the following:
+- docker compose,
+- a service account with the corresponding json credentials file to connect to the CloudSQL database,
```
vmarkovtsev, line 149:
```suggestion
- a Google Cloud service account with the corresponding json credentials file to connect to the CloudSQL database,
```

server/README.md
```diff
@@ -142,6 +142,65 @@ You can erase the API data fixtures created by `make run-api` with:
...
+- the credentials for the CloudSQL databases.
+
+Spin up the container with postgres and the container with cloud sql proxy:
+```
+$ docker-compose up cloud_sql_proxy postgres
+```
+
+This step requires the credentials of the service account as a json file. The default path is `./credentials.json`, otherwise it can be set using the `CLOUD_SQL_PROXY_CREDENTIALS_FILE` env var.
```
vmarkovtsev, line 157:
```suggestion
This step requires the credentials of the service account in a json file. The default path is `./credentials.json`, otherwise it can be set using the `CLOUD_SQL_PROXY_CREDENTIALS_FILE` env var.
```

server/README.md
```diff
@@ -142,6 +142,65 @@ You can erase the API data fixtures created by `make run-api` with:
...
+
+Once the dabase is ready to accept connection, run `load_postgre_data.py`. This script accepts a single argument that is the environment name. The accepted values are: `staging` and `production`:
```
vmarkovtsev, line 159:
```suggestion
Once the database is ready to accept connections, run `server/athenian/api/tests/load_postgres_data.py`. This script accepts a single argument that is the environment name. The accepted values are: `staging` and `production`:
```

server/README.md
```diff
@@ -142,6 +142,65 @@ You can erase the API data fixtures created by `make run-api` with:
...
+```
+$ CLOUD_SQL_POSTGRES_USER=<name of the pg user> python3 load_postgre_data.py staging
+```
+
+This will ask to input the password for the CloudSQL instance as follows:
+```
+Password for CloudSQL instance of 'staging' environment of user 'staging-cloud-sql':
+```
+
+Then it will dump `state`, `precomputed` and `metadata` databases from the live environment and restores it into the `postgres` container. Here's the example log:
```
vmarkovtsev, line 169:
```suggestion
Then it will dump `state`, `precomputed` and `metadata` databases from the live environment and restores it into the `postgres` container. Here's an example log:
```

server/README.md
```diff
@@ -142,6 +142,65 @@ You can erase the API data fixtures created by `make run-api` with:
...
+DETAIL:  Proceeding with relation creation anyway.
+Done!
+```
+
+Now you can finally run the `api`:
+```
+$ docker-compose up api
+```
+
+By default it runs the latest image, that is `gcr.io/athenian-1/api:latest`. It can be set using the `API_IMAGE` env var.
```
vmarkovtsev, line 200:
```suggestion
By default it runs the latest image, that is `gcr.io/athenian-1/api:latest`. It can be changed using the `API_IMAGE` env var.
```

docker-compose.yaml
se7entyse7en: true, I was initially willing to dump only `metadata`, but yup better to dump everything and remove this 👍

load_postgres_data.py
se7entyse7en: Wherever you think is better sounds good to me 👍

load_postgres_data.py
se7entyse7en: The reason is because I wanted this script to be as much compatible as possible so that whoever wanna run this doesn't need to run a specific version of python3, that's why I avoided using `f-strings`, but I'm ok with `%` 👍

lwsanty:
> @lwsanty regarding this I used my own personal service account, maybe we could have a service account for this use case specifically to be used by everyone?

Nope, each user should have his/her own list of permissions, so we can control access permissions

docker-compose.yaml
```diff
@@ -0,0 +1,56 @@
+version: "3.7"
```
lwsanty, line 1: please do what we discussed in slack ;)

se7entyse7en:
@vmarkovtsev @lwsanty all issues addressed and rebased on `master`.

server/tests/load_postgres_data.py
```diff
@@ -0,0 +1,93 @@
...
+        "docker-compose exec -e PGPASSWORD=%s postgres bash -c '%s'" % (
+            password, guest_cmd),
+    ], shell=True, check=True)
+
+
+def parse_args():
+    """Parse arguments for the posgres data loader command."""
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument("environment", default="staging",
```
vmarkovtsev, line 28:
```suggestion
    parser.add_argument("environment", default="staging", choices=conf,
```
and you'll not have to `if parsed_args.environment not in conf` below.

server/tests/load_postgres_data.py
```diff
@@ -0,0 +1,93 @@
...
+
+    return parsed_args
+
+
+def main():
+    """Run the postgres data loader."""
+    try:
+        args = parse_args()
+    except ValueError:
+        return 1
```
vmarkovtsev, line 56:
```suggestion
    except ValueError as e:
        print(e, file=sys.stderr)
        return 1
```

Otherwise, the user will not see the error message.

se7entyse7en:
@vmarkovtsev thanks! done

PR #325, opened 2020-05-06
Add GitHubPullRequestTimes model ENG-735

vmarkovtsev:


vmarkovtsev:
https://athenianco.atlassian.net/browse/ENG-787

server/athenian/api/models/precomputed/models.py
```diff
@@ -0,0 +1,28 @@
...
+    __tablename__ = "github_pull_request_times"
+
+    pr_node_id = Column(CHAR(32), primary_key=True)
+    release_match = Column(Text(), primary_key=True)
+    repository_full_name = Column(String(64 + 1 + 100), nullable=False)
+    pr_created_at = Column(TIMESTAMP(timezone=True), nullable=False)
+    pr_done_at = Column(TIMESTAMP(timezone=True))
+    developers = Column(HSTORE(), nullable=False)
+    format_version = Column(Integer(), nullable=False, default=1, server_default="1")
+    data = Column(LargeBinary())
```
se7entyse7en, line 24: What will be the content of this?

server/athenian/api/models/precomputed/models.py
```diff
@@ -0,0 +1,28 @@
...
+
+    __tablename__ = "github_pull_request_times"
+
+    pr_node_id = Column(CHAR(32), primary_key=True)
+    release_match = Column(Text(), primary_key=True)
+    repository_full_name = Column(String(64 + 1 + 100), nullable=False)
+    pr_created_at = Column(TIMESTAMP(timezone=True), nullable=False)
+    pr_done_at = Column(TIMESTAMP(timezone=True))
+    developers = Column(HSTORE(), nullable=False)
+    format_version = Column(Integer(), nullable=False, default=1, server_default="1")
```
se7entyse7en, line 23: Is this the version of `data`?

server/athenian/api/models/precomputed/models.py
```diff
@@ -0,0 +1,28 @@
...
+    """A group of repositories identified by an integer."""
+
+    __tablename__ = "github_pull_request_times"
+
+    pr_node_id = Column(CHAR(32), primary_key=True)
+    release_match = Column(Text(), primary_key=True)
+    repository_full_name = Column(String(64 + 1 + 100), nullable=False)
+    pr_created_at = Column(TIMESTAMP(timezone=True), nullable=False)
+    pr_done_at = Column(TIMESTAMP(timezone=True))
+    developers = Column(HSTORE(), nullable=False)
```
se7entyse7en, line 22: 🚀

server/athenian/api/models/precomputed/models.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class GitHubPullRequestTimes(Base):
+    """A group of repositories identified by an integer."""
+
+    __tablename__ = "github_pull_request_times"
+
+    pr_node_id = Column(CHAR(32), primary_key=True)
+    release_match = Column(Text(), primary_key=True)
+    repository_full_name = Column(String(64 + 1 + 100), nullable=False)
```
se7entyse7en, line 19: Why not 165?

server/athenian/api/models/precomputed/models.py
```diff
@@ -0,0 +1,28 @@
...
+from sqlalchemy.dialects.postgresql import HSTORE
+
+from athenian.api.models import create_base
+
+
+Base = create_base()
+
+
+class GitHubPullRequestTimes(Base):
+    """A group of repositories identified by an integer."""
```
se7entyse7en, line 13: Is this docstring correct?

server/athenian/api/models/precomputed/models.py
vmarkovtsev: I plan to `pickle.dumps` our `PullRequestTimes` there. We can add more data in the future, including the precomputed metrics - but not with the current arch.

server/athenian/api/models/precomputed/models.py
vmarkovtsev: Yes, the format version to avoid crashing. I have hit it several times with our 30-day-cached-but-not-really objects.

server/athenian/api/models/precomputed/models.py
vmarkovtsev: This is to indicate the current GitHub limits:
- 64 chars for the user name
- 100 chars for the repo name

server/athenian/api/models/precomputed/models.py
vmarkovtsev: wow, no :D thanks

server/athenian/api/models/precomputed/models.py
se7entyse7en: nit: Couldn't it be like a key itself inside `data`? I mean `data` could be something like:
```
{
    meta: {
        version: 1,
        // whatever
    },
    data: {
        // whatever
    }
}
```
This is just an example of a non-binary serialized version as json.

This would make the content of `data` self-contained. Now given `data` only you won't be able to de-serialize it unless you know what's the version that is saved elsewhere.

vmarkovtsev: I thought about it, but then we will not have an easy way to run some SQL to check if there are some older-versioned records left. Also, we will have to invent a binary meta-format for our format with `struct.unpack("!I", data.read(4))`. It is not worth it...

vmarkovtsev: I want to use `pickle` because it is the fastest deserialization except `marshal` for our data (I checked). My profiling proved that we have to care about these things.

se7entyse7en: ok fair enough 👍 Thanks for the explanation

PR #334, opened 2020-05-11
Add spec for team endpoints

se7entyse7en:

### DIFF
--- a/server/athenian/api/controllers/contributors_controller.py
+++ b/server/athenian/api/controllers/contributors_controller.py
@@ -1,0 +1,10 @@
+from aiohttp import web
+
+
+async def get_contributors(request: web.Request, id: int) -> web.Response:
+    """List all the contributors belonging to the specified account.
+
+    :param id: Numeric identifier of the account.
+
+    """
+    return web.Response(status=200)
--- a/server/athenian/api/controllers/team_controller.py
+++ b/server/athenian/api/controllers/team_controller.py
@@ -1,0 +1,56 @@
+from typing import Union
+
+from aiohttp import web
+
+from athenian.api.models.web.team_create_request import TeamCreateRequest
+from athenian.api.models.web.team_update_request import TeamUpdateRequest
+
+
+async def create_team(request: web.Request,
+                      body: Union[dict, bytes] = None) -> web.Response:
+    """Create a team.
+
+    :param body: Team creation request body.
+
+    """
+    body = TeamCreateRequest.from_dict(body)
+    return web.Response(status=200)
+
+
+async def delete_team(request: web.Request, id: int) -> web.Response:
+    """Delete a team.
+
+    :param id: Numeric identifier of the team to delete.
+
+    """
+    return web.Response(status=200)
+
+
+async def get_team(request: web.Request, id: int) -> web.Response:
+    """List the team's members. The user must belong to the account that owns the team.
+
+    :param id: Numeric identifier of the team to list.
+
+    """
+    return web.Response(status=200)
+
+
+async def list_teams(request: web.Request, id: int) -> web.Response:
+    """List the teams belonging to the current user.
+
+    :param id: Numeric identifier of the account.
+
+    """
+    return web.Response(status=200)
+
+
+async def update_team(request: web.Request, id: int,
+                      body: Union[dict, bytes] = None) -> web.Response:
+    """Update a team.
+
+    :param id: Numeric identifier of the team to update.
+    :param body: Team update request body.
+
+    """
+    body = TeamUpdateRequest.from_dict(body)
+    return web.Response(status=200)
--- a/server/athenian/api/models/web/contributor.py
+++ b/server/athenian/api/models/web/contributor.py
@@ -1,0 +1,130 @@
+from typing import Optional
+
+from athenian.api.models.web.base_model_ import Model
+
+
+class Contributor(Model):
+    """Details about a developer who contributed to some repositories owned by the account."""
+
+    def __init__(
+            self,
+            login: Optional[str] = None,
+            name: Optional[str] = None,
+            email: Optional[str] = None,
+            picture: Optional[str] = None,
+    ):
+        """Contributor - a model defined in OpenAPI
+
+        :param login: The login of this Contributor.
+        :param name: The name of this Contributor.
+        :param email: The email of this Contributor.
+        :param picture: The picture of this Contributor.
+        """
+        self.openapi_types = {"login": str, "name": str, "email": str, "picture": str}
+
+        self.attribute_map = {
+            "login": "login",
+            "name": "name",
+            "email": "email",
+            "picture": "picture",
+        }
+
+        self._login = login
+        self._name = name
+        self._email = email
+        self._picture = picture
+
+    @property
+    def login(self) -> str:
+        """Gets the login of this Contributor.
+
+        User name which uniquely identifies any developer on any service provider.
+        The format matches the profile URL without the protocol part.
+
+        :return: The login of this Contributor.
+        """
+        return self._login
+
+    @login.setter
+    def login(self, login: str):
+        """Sets the login of this Contributor.
+
+        User name which uniquely identifies any developer on any service provider.
+        The format matches the profile URL without the protocol part.
+
+        :param login: The login of this Contributor.
+        """
+        if login is None:
+            raise ValueError("Invalid value for `login`, must not be `None`")
+
+        self._login = login
+
+    @property
+    def name(self) -> str:
+        """Gets the name of this Contributor.
+
+        Full name of the contributor.
+
+        :return: The name of this Contributor.
+        """
+        return self._name
+
+    @name.setter
+    def name(self, name: str):
+        """Sets the name of this Contributor.
+
+        Full name of the contributor.
+
+        :param name: The name of this Contributor.
+        """
+        if name is None:
+            raise ValueError("Invalid value for `name`, must not be `None`")
+
+        self._name = name
+
+    @property
+    def email(self) -> str:
+        """Gets the email of this Contributor.
+
+        Email of the conributor.
+
+        :return: The email of this Contributor.
+        :rtype: str
+        """
+        return self._email
+
+    @email.setter
+    def email(self, email: str):
+        """Sets the email of this Contributor.
+
+        Email of the conributor.
+
+        :param email: The email of this Contributor.
+        """
+        if email is None:
+            raise ValueError("Invalid value for `email`, must not be `None`")
+
+        self._email = email
+
+    @property
+    def picture(self) -> str:
+        """Gets the picture of this Contributor.
+
+        Avatar URL of the contributor.
+
+        :return: The picture of this Contributor.
+        """
+        return self._picture
+
+    @picture.setter
+    def picture(self, picture: str):
+        """Sets the picture of this Contributor.
+
+        Avatar URL of the contributor.
+
+        :param picture: The picture of this Contributor.
+        """
+        if picture is None:
+            raise ValueError("Invalid value for `picture`, must not be `None`")
+
+        self._picture = picture
--- a/server/athenian/api/models/web/team.py
+++ b/server/athenian/api/models/web/team.py
@@ -1,0 +1,97 @@
+from typing import List, Optional
+
+from athenian.api.models.web.base_model_ import Model
+from athenian.api.models.web.contributor import Contributor
+
+
+class Team(Model):
+    """List of team members."""
+
+    def __init__(
+            self,
+            id: Optional[int] = None,
+            name: Optional[str] = None,
+            members: Optional[List[Contributor]] = None,
+    ):
+        """Team - a model defined in OpenAPI
+
+        :param id: The id of this Team.
+        :param name: The name of this Team.
+        :param members: The members of this Team.
+        """
+        self.openapi_types = {"id": int, "name": str, "members": List[Contributor]}
+
+        self.attribute_map = {"id": "id", "name": "name", "members": "members"}
+
+        self._id = id
+        self._name = name
+        self._members = members
+
+    @property
+    def id(self) -> int:
+        """Gets the id of this Team.
+
+        Team identifier.
+
+        :return: The id of this Team.
+        """
+        return self._id
+
+    @id.setter
+    def id(self, id: int):
+        """Sets the id of this Team.
+
+        Team identifier.
+
+        :param id: The id of this Team.
+        """
+        if id is None:
+            raise ValueError("Invalid value for `id`, must not be `None`")
+
+        self._id = id
+
+    @property
+    def name(self) -> str:
+        """Gets the name of this Team.
+
+        Name of the team.
+
+        :return: The name of this Team.
+        """
+        return self._name
+
+    @name.setter
+    def name(self, name: str):
+        """Sets the name of this Team.
+
+        Name of the team.
+
+        :param name: The name of this Team.
+        """
+        if name is None:
+            raise ValueError("Invalid value for `name`, must not be `None`")
+
+        self._name = name
+
+    @property
+    def members(self) -> List[Contributor]:
+        """Gets the members of this Team.
+
+        List of contributors.
+
+        :return: The members of this Team.
+        """
+        return self._members
+
+    @members.setter
+    def members(self, members: List[Contributor]):
+        """Sets the members of this Team.
+
+        List of contributors.
+
+        :param members: The members of this Team.
+        """
+        if members is None:
+            raise ValueError("Invalid value for `members`, must not be `None`")
+
+        self._members = members
--- a/server/athenian/api/models/web/team_create_request.py
+++ b/server/athenian/api/models/web/team_create_request.py
@@ -1,0 +1,102 @@
+from typing import List, Optional
+
+from athenian.api.models.web.base_model_ import Model
+
+
+class TeamCreateRequest(Model):
+    """Team creation request."""
+
+    def __init__(
+            self,
+            account: Optional[int] = None,
+            name: Optional[str] = None,
+            members: Optional[List[str]] = None,
+    ):
+        """TeamCreateRequest - a model defined in OpenAPI
+
+        :param account: The account of this TeamCreateRequest.
+        :param name: The name of this TeamCreateRequest.
+        :param members: The members of this TeamCreateRequest.
+        """
+        self.openapi_types = {"account": int, "name": str, "members": List[str]}
+
+        self.attribute_map = {
+            "account": "account",
+            "name": "name",
+            "members": "members",
+        }
+
+        self._account = account
+        self._name = name
+        self._members = members
+
+    @property
+    def account(self) -> int:
+        """Gets the account of this TeamCreateRequest.
+
+        Account identifier. That account will own the created team.
+
+        :return: The account of this TeamCreateRequest.
+        """
+        return self._account
+
+    @account.setter
+    def account(self, account: int):
+        """Sets the account of this TeamCreateRequest.
+
+        Account identifier. That account will own the created team.
+
+        :param account: The account of this TeamCreateRequest.
+        """
+        if account is None:
+            raise ValueError("Invalid value for `account`, must not be `None`")
+
+        self._account = account
+
+    @property
+    def name(self) -> str:
+        """Gets the name of this TeamCreateRequest.
+
+        Name of the team.
+
+        :return: The name of this TeamCreateRequest.
+        """
+        return self._name
+
+    @name.setter
+    def name(self, name: str):
+        """Sets the name of this TeamCreateRequest.
+
+        Name of the team.
+
+        :param name: The name of this TeamCreateRequest.
+        """
+        if name is None:
+            raise ValueError("Invalid value for `name`, must not be `None`")
+
+        self._name = name
+
+    @property
+    def members(self) -> List[str]:
+        """Gets the members of this TeamCreateRequest.
+
+        A set of developers. An empty list disables the filter and includes everybody.
+        Duplicates are automatically ignored.
+
+        :return: The members of this TeamCreateRequest.
+        """
+        return self._members
+
+    @members.setter
+    def members(self, members: List[str]):
+        """Sets the members of this TeamCreateRequest.
+
+        A set of developers. An empty list disables the filter and includes everybody.
+        Duplicates are automatically ignored.
+
+        :param members: The members of this TeamCreateRequest.
+        """
+        if members is None:
+            raise ValueError("Invalid value for `members`, must not be `None`")
+
+        self._members = members
--- a/server/athenian/api/models/web/team_update_request.py
+++ b/server/athenian/api/models/web/team_update_request.py
@@ -1,0 +1,98 @@
+from typing import List, Optional
+
+from athenian.api.models.web.base_model_ import Model
+
+
+class TeamUpdateRequest(Model):
+    """Team update request."""
+
+    def __init__(
+            self,
+            id: Optional[int] = None,
+            name: Optional[str] = None,
+            members: Optional[List[str]] = None,
+    ):
+        """TeamUpdateRequest - a model defined in OpenAPI
+
+        :param id: The id of this TeamUpdateRequest.
+        :param name: The name of this TeamUpdateRequest.
+        :param members: The members of this TeamUpdateRequest.
+        """
+        self.openapi_types = {"id": int, "name": str, "members": List[str]}
+
+        self.attribute_map = {"id": "id", "name": "name", "members": "members"}
+
+        self._id = id
+        self._name = name
+        self._members = members
+
+    @property
+    def id(self) -> int:
+        """Gets the id of this TeamUpdateRequest.
+
+        Team identifier.
+
+        :return: The id of this TeamUpdateRequest.
+        """
+        return self._id
+
+    @id.setter
+    def id(self, id: int):
+        """Sets the id of this TeamUpdateRequest.
+
+        Team identifier.
+
+        :param id: The id of this TeamUpdateRequest.
+        """
+        if id is None:
+            raise ValueError("Invalid value for `id`, must not be `None`")
+
+        self._id = id
+
+    @property
+    def name(self) -> str:
+        """Gets the name of this TeamUpdateRequest.
+
+        Name of the team.
+
+        :return: The name of this TeamUpdateRequest.
+        """
+        return self._name
+
+    @name.setter
+    def name(self, name: str):
+        """Sets the name of this TeamUpdateRequest.
+
+        Name of the team.
+
+        :param name: The name of this TeamUpdateRequest.
+        """
+        if name is None:
+            raise ValueError("Invalid value for `name`, must not be `None`")
+
+        self._name = name
+
+    @property
+    def members(self) -> List[str]:
+        """Gets the members of this TeamUpdateRequest.
+
+        A set of developers. An empty list disables the filter and includes everybody.
+        Duplicates are automatically ignored.
+
+        :return: The members of this TeamUpdateRequest.
+        """
+        return self._members
+
+    @members.setter
+    def members(self, members: List[str]):
+        """Sets the members of this TeamUpdateRequest.
+
+        A set of developers. An empty list disables the filter and includes everybody.
+        Duplicates are automatically ignored.
+
+        :param members: The members of this TeamUpdateRequest.
+        """
+        if members is None:
+            raise ValueError("Invalid value for `members`, must not be `None`")
+
+        self._members = members
--- a/server/athenian/api/openapi/openapi.yaml
+++ b/server/athenian/api/openapi/openapi.yaml
@@ -665,7 +665,7 @@ paths:
     delete:
       operationId: delete_reposet
       parameters:
-      - description: Numeric identifier of the repository set to list.
+      - description: Numeric identifier of the repository set to delete.
         explode: false
         in: path
         name: id
@@ -747,7 +747,7 @@ paths:
     put:
       operationId: update_reposet
       parameters:
-      - description: Numeric identifier of the repository set to list.
+      - description: Numeric identifier of the repository set to update.
         explode: false
         in: path
         name: id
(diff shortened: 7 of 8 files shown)


se7entyse7en:
I'll continue development starting from this and eventually fix according to issues you might raise.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,183 @@ paths:
...
+      x-openapi-router-controller: athenian.api.controllers.team_controller
+      x-codegen-request-body-name: body
+    put:
+      operationId: update_team
+      parameters:
+      - description: Numeric identifier of the team to update, empty on create.
+        explode: false
+        in: path
+        name: id
+        # Not actually required, but cannot set optional path parameter:
```
se7entyse7en, line 930: @vmarkovtsev wdyt about this? Do you have a better approach?

warenlg:
Wow, I'm curious to see how we label this PR...

warenlg:
Review Required, not bad!

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: You set it as `required: true` so we will not be able to call `PUT /team` to create a new one. 

I had a similar problem with `/become` and I solved it with a URL parameter. However, we've got a CRUD here...

So maybe I added `CREATE /reposet` because of this issue.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,183 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GenericError'
+          description: The specified team was not found.
+        409:
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GenericError'
+          description: There is an existing team with the same contents.
```
vmarkovtsev, line 966: Yes, this is a very good check to avoid duplicate teams in our precious DB :+1:

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1083,6 +1260,90 @@ components:
       items:
         $ref: '#/components/schemas/DeveloperID'
       type: array
+    TeamMember:
```
vmarkovtsev, line 1263: I wonder if we can merge `TeamMember` with `User` which are reported in `/account`...

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1083,6 +1260,91 @@ components:
...
+      description: A team member.
+      example:
+        login: github.com/gkwillie
+        id: github|60340680
+        native_id: 60340680
```
vmarkovtsev, line 1268: Let's not add `native_id` to the response, I cannot imagine why it would needed, and you'd better not know [how we get it](https://github.com/athenianco/athenian-api/blob/master/server/athenian/api/auth.py#L350) :smile:

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1083,6 +1260,91 @@ components:
...
+        name: Groundskeeper Willie
+        email: bot@athenian.co
+        picture: https://avatars0.githubusercontent.com/u/60340680?v=4
+        updated: 2020-01-23T12:00:00Z
+      allOf:
+        - $ref: '#/components/schemas/User'
+        - properties:
+            login:
```
vmarkovtsev, line 1276: How about adding this to the `User` model? Thus we can delete this `allOf`.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1083,6 +1260,91 @@ components:
...
+              $ref: '#/components/schemas/DeveloperID'
+          required:
+            - login
+    Team:
+      description: A list of members part of the same team.
```
vmarkovtsev, line 1281:
```suggestion
      description: List of team members.
```

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1083,6 +1260,91 @@ components:
...
+          updated: 2020-01-23T12:00:00Z
+        - login: github.com/se7entyse7en
+          id: github|5599208
+          native_id: 5599208
+          name: Lou Marvin Caraig
+          email: loumarvincaraig@gmail.com
+          picture: https://avatars1.githubusercontent.com/u/5599208?v=4
+          updated: 2020-05-05T18:19:05Z
+      properties:
+        id:
```
vmarkovtsev, line 1301:
```suggestion
        id:
          description: Team identifier.
```

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1083,6 +1260,91 @@ components:
...
+          type: integer
+        name:
+          type: string
```
vmarkovtsev, line 1304:
```suggestion
          description: Name of the team.
          type: string
```

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,183 @@ paths:
...
+        schema:
+          type: integer
+        style: simple
+      responses:
+        200:
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Team'
+          description: List of members in the specified team.
```
vmarkovtsev, line 899:
```suggestion
          description: List of the members in the specified team.
```

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,183 @@ paths:
...
+          description: The user is forbidden to list the specified team.
+        404:
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GenericError'
+          description: The specified team was not found.
+      security:
+      - bearerAuth: []
+      summary: List a team's members. The user must be in the account that owns the
```
vmarkovtsev, line 914:
```suggestion
      summary: List the team's members. The user must belong to the account that owns the
```

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,183 @@ paths:
...
+      - team
+# {% if False %}
+      - default
+# {% endif %}
+      x-openapi-router-controller: athenian.api.controllers.team_controller
+      x-codegen-request-body-name: body
+    put:
+      operationId: update_team
+      parameters:
+      - description: Numeric identifier of the team to update, empty on create.
```
vmarkovtsev, line 926:
```suggestion
      - description: Numeric identifier of the team to update, empty on creation.
```

server/athenian/api/openapi/openapi.yaml
se7entyse7en: Got it, it makes sense. I'll then add an endpoint similarly to `/reposet`

server/athenian/api/openapi/openapi.yaml
se7entyse7en: Indeed, probably that would the best thing. 👍

server/athenian/api/openapi/openapi.yaml
se7entyse7en: hahaha 😂 

Anyway, the only reason I put is to reuse `User` and it's a required property in `User`,

vmarkovtsev: ok, then let's not touch it.

se7entyse7en:
@vmarkovtsev all issues should have been addressed, and I added the endpoint for team creation

se7entyse7en:
But I guess this cannot be merged until I do not generate the models and add the new `login` attribute in `User`.

server/athenian/api/models/web/team.py
```diff
@@ -0,0 +1,86 @@
...
+
+from typing import List, Dict, Type
+
+from athenian.api.models.base_model_ import Model
+from athenian.api.models.user import User
+from athenian.api import util
+
+
+class Team(Model):
+    """NOTE: This class is auto generated by OpenAPI Generator (https://openapi-generator.tech).
```
vmarkovtsev, line 13: Can you please copy-paste the model description here

server/athenian/api/models/web/team.py
```diff
@@ -0,0 +1,86 @@
+# coding: utf-8
```
vmarkovtsev, line 1:
```suggestion
```
(we don't need it in py3)

server/athenian/api/models/web/team.py
```diff
@@ -0,0 +1,86 @@
...
+from athenian.api import util
+
+
+class Team(Model):
+    """NOTE: This class is auto generated by OpenAPI Generator (https://openapi-generator.tech).
+
+    Do not edit the class manually.
+    """
+
+    def __init__(self, name: str = None, members: List[User] = None):
```
vmarkovtsev, line 18:
```suggestion
    def __init__(self, name: Optional[str] = None, members: Optional[List[User]] = None):
```

server/athenian/api/models/web/team.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @classmethod
+    def from_dict(cls, dikt: dict) -> "Team":
+        """Returns the dict as a model
+
+        :param dikt: A dict.
+        :return: The Team of this Team.
+        """
+        return util.deserialize_model(dikt, cls)
+
```
vmarkovtsev, line 39:
```suggestion
```

server/athenian/api/models/web/team.py
```diff
@@ -0,0 +1,86 @@
...
+    @property
+    def name(self):
```
vmarkovtsev, line 41:
```suggestion
    def name(self) -> str:
```

server/athenian/api/models/web/team.py
```diff
@@ -0,0 +1,86 @@
...
+        """Gets the name of this Team.
+
+        Name of the team.
+
+        :return: The name of this Team.
+        :rtype: str
```
vmarkovtsev, line 47:
```suggestion
```

server/athenian/api/models/web/team.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        return self._name
+
+    @name.setter
+    def name(self, name):
```
vmarkovtsev, line 52:
```suggestion
    def name(self, name: str):
```
... and so on. The default generated code is not ideal.

server/athenian/api/models/web/user.py
```diff
@@ -131,7 +136,30 @@ def id(self, id: str):
         self._id = id
 
     @property
-    def native_id(self) -> str:
+    def login(self):
```
vmarkovtsev, line 139:
```suggestion
    def login(self) -> str:
```

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1083,6 +1296,139 @@ components:
...
+        team:
+          $ref: '#/components/schemas/Team'
+      required:
+      - account
+      - team
+      type: object
+    TeamID:
+      description: Team identifier.
+      example: Engineering
+      type: string
```
vmarkovtsev, line 1333:
```suggestion
```
(not used)

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,219 @@ paths:
...
+  /team/create:
+    post:
+      operationId: create_team
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/TeamCreateRequest'
+        x-body-name: body
+      responses:
```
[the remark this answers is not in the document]
vmarkovtsev: How about 422 Unprocessable Entity. We raise 400 on JSON syntax errors, so if we can find a more specific code for this error then it will be distinguishable.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,219 @@ paths:
...
+      summary: Create a team.
+      tags:
+      - team
+# {% if False %}
+      - default
+# {% endif %}
+      x-openapi-router-controller: athenian.api.controllers.team_controller
+      x-codegen-request-body-name: body
+  /team/{id}:
+    delete:
```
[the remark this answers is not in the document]
vmarkovtsev: We are likely to use ARRAY the same way as in reposets to migrate less tables, though IDK exactly how Marvin sees it.

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
dpordomingo: `422` lgtm

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,219 @@ paths:
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/TeamCreateRequest'
+        x-body-name: body
+      responses:
+        200:
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CreatedIdentifier'
```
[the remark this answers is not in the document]
se7entyse7en: yup totally makes sense to me, I just almost blindly copied what was being sent during the `PUT`, gonna change it since all those info are available to the api 👍

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
dpordomingo: A problem I see if we embed the users in the teams is that they can be outdated easily if we do not implement other update processes.
Also, if at some point we need to change their schema, the migration process would be also more complex.

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
se7entyse7en: good point, I'd just go for 400 with different error message. I think that distinguishing these two different cases by different status code will be important when and if we're going to have clients different from the webapp.

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
se7entyse7en: Yup, I'd rather user an ARRAY as we do for `reposets` 👍

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,219 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GenericError'
+          description: The user does not belong to the specified account.
+        409:
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GenericError'
+          description: Concurrent requests that are finishing the account registration.
```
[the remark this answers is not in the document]
se7entyse7en: mhmmm tricky. tbh I added this to make as much similar to `/reposets/{id}`, but probably in this case even `200` with empty list would fit as well. I think that `409` in case of non-finished account registration fits for enpdoints involving write ops.

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
dpordomingo: are we embedding the whole user or only its id?

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
dpordomingo: Agree on returning empty list for read ops, and error on write ones.

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
vmarkovtsev: Only the ids - or, more precisely, their logins, if we allow out-of-account members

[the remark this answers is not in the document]
dpordomingo: then lgtm.
(`login` can change, so it might be not the best way to identify the users in the embedded list)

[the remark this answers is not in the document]
se7entyse7en: I still need to better see how this could be done, do you need the answer now? I don't think that the way it is going to implemented will change the spec.

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
se7entyse7en: done [here](https://github.com/athenianco/athenian-api/pull/334/commits/b197c83d65ad261309a3cdb0c546443eb8a8ae0c).

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
se7entyse7en: done [here](https://github.com/athenianco/athenian-api/pull/334/commits/6ce2316202b0863ecbb1a3a98385b49771ffc446).

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,219 @@ paths:
...
+        schema:
+          type: integer
+        style: simple
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Team'
+        x-body-name: body
+      responses:
```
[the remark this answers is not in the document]
se7entyse7en: done [here](https://github.com/athenianco/athenian-api/pull/334/commits/57e0f929faa220ecdf10e91a922d70325f407f27).

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,219 @@ paths:
...
+        200:
+          content:
+            application/json:
+              schema:
+                type: object
+          description: Empty object indicates a successful operation.
```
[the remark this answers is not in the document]
se7entyse7en: done [here](https://github.com/athenianco/athenian-api/pull/334/commits/eebbedc6dcb31817057c34cddedde07b6cd6e1a8).

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
dpordomingo: I don't need an answer. I'm just providing info :D

server/athenian/api/models/web/user.py
```diff
@@ -131,7 +136,32 @@ def id(self, id: str):
...
+
+        :param login: The login of this User.
+        """
+        if login is None:
+            raise ValueError("Invalid value for `login`, must not be `None`")
+
+        self._login = login
+
+    @property
+    def native_id(self):
```
vmarkovtsev, line 164:
```suggestion
    def native_id(self) -> str:
```

se7entyse7en:
Once approved I'll squash commits to have a sane history

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1083,6 +1308,152 @@ components:
...
+            updated: 2020-05-01T17:15:51Z
+          - login: github.com/se7entyse7en
+            id: github|5599208
+            native_id: 5599208
+            name: Lou Marvin Caraig
+            email: waren@athenian.co
+            picture: https://avatars1.githubusercontent.com/u/5599208?v=4
+            updated: 2020-05-05T18:19:05Z
+      properties:
+        id:
```
vmarkovtsev, line 1412: Should we move `id` inside the `Team`? All the places that I see have them together.

I am also worried about how the team listing will work. If we accept a list of github logins then we will not always have an `id` - that's an Auth0 user ID. We cannot make it optional for the `User` either. It looks like we have to return back our `TeamMember`.

server/athenian/api/controllers/team_controller.py
```diff
@@ -0,0 +1,48 @@
...
+async def create_team(request: AthenianWebRequest, body: dict) -> web.Response:
+    """Create a team.
+
+    :param request: HTTP request.
+    :param body: List of repositories to group.
+    """
+    pass
+
+
+async def delete_team(request: AthenianWebRequest, body: dict) -> web.Response:
```
vmarkovtsev, line 15: Not all the handlers have a `body` - DELETE, for example.

vmarkovtsev, line 15:
```suggestion
async def delete_team(request: AthenianWebRequest, id: int) -> web.Response:
```

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,231 @@ paths:
...
+        name: id
+        required: true
+        schema:
+          type: integer
+        style: simple
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/TeamUpdateRequest'
```
dpordomingo, line 982: Is it a good idea to be able to change the `team.id` within the `TeamUpdateRequest`? Since the team to be updated is already identified by the called URL, the `team.id` could be omitted.

server/athenian/api/openapi/openapi.yaml
se7entyse7en: > Should we move id inside the Team? All the places that I see have them together.

Good point, indeed it was needed when the put needed to handle empty id. 👍 

> I am also worried about how the team listing will work. If we accept a list of github logins then we will not always have an id - that's an Auth0 user ID. We cannot make it optional for the User either. It looks like we have to return back our TeamMember.

Okay, I got the point. But this is related mostly to creation and update itself, isn't it? Instead of accepting a list of GH handles they should accept a list of ids.

se7entyse7en:
@vmarkovtsev in addition the schema I also need to update all the places dealing with `User` to include the `login` attribute as well.

server/athenian/api/openapi/openapi.yaml
se7entyse7en: good point, since it's already in the path itself 👍

vmarkovtsev: This is a good point :+1:

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: IDK what's the product decision here: allow any valid github login, allow any github login from the account's orgs or allow registered account users...

se7entyse7en: I'd go strict, that is users that are member of the account, that would simplify a lot. Also in that case no need to go back to `TeamMember` since all members would be regular `User`s. \cc @warenlg

se7entyse7en:
@vmarkovtsev @dpordomingo I updated the spec [here](https://github.com/athenianco/athenian-api/pull/334/commits/20fbb892bc072bc6865d9a08a6f82247686e634b):
- since as for Product requirement we need team  members that aren't users, a team now is not a list of `User`s, but a list of `TeamMember`s,
- for now, I reverted adding `login` to `User` since now it's not needed,
- added a `/contributors` endpoint to return all the contributors of an account.

@dpordomingo I didn't force-pushed, but it will be much easier to see the whole diff.

se7entyse7en:
@dpordomingo I just realized that from `/contributors` I need to return also avatars, name, etc. I'm on it.

se7entyse7en:
done [here](https://github.com/athenianco/athenian-api/pull/334/commits/b835cd1b88bef203239c907dec59f2a8488ff9ce).

se7entyse7en:
Anyway @dpordomingo I'll know better what kind of errors (non-200) will be returned once I'll implement the endpoints.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1082,6 +1082,12 @@ paths:
               schema:
                 $ref: '#/components/schemas/Contributors'
           description: List of all contributors belonging to the specified account.
+        403:
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GenericError'
+          description: The user is forbidden to list contributors belonging to the specified account.
```
dpordomingo, line 1090: I guess than when implementing this, all account members will be allowed to list account team members (since regular users are already [able to change release settings](https://athenianco.atlassian.net/browse/ENG-696?focusedCommentId=12352), I'd say that they should be also able to list team members) 
wdyt @warenlg ?

server/athenian/api/controllers/contributors_controller.py
```diff
@@ -0,0 +1,10 @@
+from aiohttp import web
+
+
+async def get_contributors(request: web.Request, id: int) -> web.Response:
+    """List all the contributors belonging to the specified account.
```
vmarkovtsev, line 5: We should really-really post this in a separate PR. This is becoming an unmanageable monster.

server/athenian/api/models/web/contributor.py
```diff
@@ -0,0 +1,130 @@
+from typing import Optional
+
+from athenian.api.models.web.base_model_ import Model
+
+
+class Contributor(Model):
+    """Contributor details."""
```
vmarkovtsev, line 7: If this is a spec description then we should be a bit more verbose here.
```suggestion
    """Details about a developer who contributed to some repositories owned by the account."""
```

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,278 @@ paths:
...
+        404:
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GenericError'
+          description: The user does not belong to the specified account.
```
dpordomingo, line 1096: (not blocker) I'd say it should be returned a `403` if the user is not allowed to list the account contributors.

server/athenian/api/models/web/team_create_request.py
```diff
@@ -0,0 +1,102 @@
...
+
+from athenian.api.models.web.base_model_ import Model
+
+
+class TeamCreateRequest(Model):
+    """Team creation request."""
+
+    def __init__(
+        self,
+            account: Optional[int] = None,
```
vmarkovtsev, line 11: Indentation fault?

server/athenian/api/models/web/team_update_request.py
```diff
@@ -0,0 +1,98 @@
+from typing import List, Optional
+
+from athenian.api.models.web.base_model_ import Model
+
+
+class TeamUpdateRequest(Model):
+    """Team update request."""
+
+    def __init__(
+            self,
```
vmarkovtsev, line 10: I think we should de-indent -4 spaces here

server/athenian/api/openapi/openapi.yaml
```diff
@@ -838,6 +838,278 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Contributors'
+          description: List of all contributors belonging to the specified account.
+        403:
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GenericError'
+          description: The user is forbidden to list contributors belonging to the specified account.
```
vmarkovtsev, line 1090: Hence this should go in a separate PR

server/athenian/api/controllers/contributors_controller.py
se7entyse7en: Do you mind if we keep just the spec and the stub? The idea is that with this version of the spec everything needed for the settings page is ready. Then after I finished the implementation I'll move to change the filtering.

server/athenian/api/models/web/team_create_request.py
se7entyse7en: yup 🤦

server/athenian/api/models/web/contributor.py
se7entyse7en: added both here and in spec

se7entyse7en:
@vmarkovtsev for better review I'd rather open separate PRs for each endpoint after this.

server/athenian/api/controllers/contributors_controller.py
se7entyse7en: This is actually also the first endpoint that I implemented that just needs to be opened 😄

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1371,7 +1371,7 @@ components:
       type: array
     Contributor:
       additionalProperties: false
-      description: Contributor details.
+      description: Details about a developer who contributed to some repositories owned by the account.
```
dpordomingo, line 1374: Can't be added new hires to their team till they make a contribution? Why not listing also organization members?

se7entyse7en: It will behave the same way as `/filter/contributors` just without the time constraint. New hires and org members won't be shown in the filtering as well as now. Is this limitation ok @warenlg ?

server/athenian/api/openapi/openapi.yaml
se7entyse7en: Since errors are minor things to be completely defined now, I'd keep them as they're now and I'll review all of them on each endpoint that I'll implement to make the errors defined in the spec reflect the implementation.

server/athenian/api/openapi/openapi.yaml
se7entyse7en: yup as described in ENG-726 all users can manage teams.

server/athenian/api/openapi/openapi.yaml
warenlg: Yes, it is ok, I don't want us to spend eng effort on new hires that have not contributed anywhere yet. The use case is too minor, and there is no need to handle it from our users.

server/athenian/api/models/web/user.py
```diff
@@ -15,6 +15,7 @@ class User(Model):
     def __init__(
         self,
         id: Optional[str] = None,
+        login: Optional[str] = None,
```
vmarkovtsev, line 18: This must be covered in the tests + handled in `auth.py`. We can revert back if it is too much side work.

se7entyse7en: Ah indeed! Sorry, as I said [here](https://github.com/athenianco/athenian-api/pull/334#issuecomment-627276698) I actually reverted it back from the spec, but then forgot to revert the model itself. I'll do when cleaning up the history since it's not required now and would require extra work.

se7entyse7en:
I cleaned up the history, I'll merge this once CI gives the green light.

server/athenian/api/models/web/user.py
vmarkovtsev: :+1:

dpordomingo:
@marvin Are these new endpoints working at staging?
```
curl -X GET "https://api-staging.athenian.co/v1/teams/1" -H "accept: application/json"
```
No response, no headers...
Imo it should return an empty array if no teams were found.

Then I tried to create a team:
```
curl 'https://api-staging.athenian.co/v1/team/create' \
    --data-binary '{
        "account":1,
        "name":"Engineers",
        "members":["github.com/dpordomingo","github.com/se7entyse7en"]
    }'
```
No response, no headers...
It should return the ID of the created group.

Then I tried to list the teams again to receive the created one
```
curl -X GET "https://api-staging.athenian.co/v1/teams/1" -H "accept: application/json"
```
No response, no headers...

vmarkovtsev:
@dpordomingo No, they are not implemented yet.

se7entyse7en:
Indeed, this PR was meant to just prepare the spec and add the stubs as minimal as possible to just make tests pass and be able to merge the spec. The first endpoint that will be ready is `/contributors/{id}`: https://github.com/athenianco/athenian-api/pull/344

PR #344, opened 2020-05-13
Implement `/contributors/{id}` endpoint

se7entyse7en:
This implements the `/contributors/{id}` endpoint and updates the spec file with the actual errors that could be returned.

### DIFF
--- a/server/athenian/api/controllers/contributors_controller.py
+++ b/server/athenian/api/controllers/contributors_controller.py
@@ -2,7 +2,7 @@ from aiohttp import web
 from sqlalchemy import and_, select
 
 from athenian.api.controllers.miners.access_classes import access_classes
-from athenian.api.controllers.miners.github.contributors import mine_all_contributors
+from athenian.api.controllers.miners.github.contributors import mine_contributors
 from athenian.api.models.state.models import UserAccount
 from athenian.api.models.web import Contributor, NotFoundError
 from athenian.api.response import FriendlyJson, ResponseError
@@ -31,7 +31,8 @@ async def get_contributors(request: web.Request, id: int) -> web.Response:
             return e.response
 
         repos = checker.installed_repos(with_prefix=False)
-        users = await mine_all_contributors(repos, request.mdb, request.cache)
+        users = await mine_contributors(repos, request.mdb, with_stats=False,
+                                        cache=request.cache)
 
         contributors = [
             Contributor(login=f"github.com/{u['login']}", name=u["name"],
--- a/server/athenian/api/controllers/contributors_controller.py
+++ b/server/athenian/api/controllers/contributors_controller.py
@@ -1,4 +1,11 @@
 from aiohttp import web
+from sqlalchemy import and_, select
+
+from athenian.api.controllers.miners.access_classes import access_classes
+from athenian.api.controllers.miners.github.contributors import mine_all_contributors
+from athenian.api.models.state.models import UserAccount
+from athenian.api.models.web import Contributor, NotFoundError
+from athenian.api.response import FriendlyJson, ResponseError
 
 
 async def get_contributors(request: web.Request, id: int) -> web.Response:
@@ -7,4 +14,29 @@ async def get_contributors(request: web.Request, id: int) -> web.Response:
     :param id: Numeric identifier of the account.
 
     """
-    return web.Response(status=200)
+    user = await request.user()
+
+    async with request.sdb.connection() as sdb_conn:
+        account_id = await sdb_conn.fetch_val(
+            select([UserAccount.account_id])
+            .where(and_(UserAccount.user_id == user.id, UserAccount.account_id == id)))
+        if account_id is None:
+            err_detail = f"Account {account_id} does not exist or user {user} is not a member."
+            return ResponseError(NotFoundError(detail=err_detail)).response
+
+        try:
+            checker = await access_classes["github"](
+                account_id, sdb_conn, request.mdb, request.cache).load()
+        except ResponseError as e:
+            return e.response
+
+        repos = checker.installed_repos(with_prefix=False)
+        users = await mine_all_contributors(repos, request.mdb, request.cache)
+
+        contributors = [
+            Contributor(login=f"github.com/{u['login']}", name=u["name"],
+                        email=u["email"], picture=u["avatar_url"]).to_dict()
+            for u in users
+        ]
+
+        return web.json_response(contributors, dumps=FriendlyJson.dumps)
--- a/server/athenian/api/controllers/contributors_controller.py
+++ b/server/athenian/api/controllers/contributors_controller.py
@@ -14,14 +14,15 @@ async def get_contributors(request: web.Request, id: int) -> web.Response:
     :param id: Numeric identifier of the account.
 
     """
-    user = await request.user()
-
     async with request.sdb.connection() as sdb_conn:
         account_id = await sdb_conn.fetch_val(
             select([UserAccount.account_id])
-            .where(and_(UserAccount.user_id == user.id, UserAccount.account_id == id)))
+            .where(and_(UserAccount.user_id == request.uid, UserAccount.account_id == id)))
         if account_id is None:
-            err_detail = f"Account {account_id} does not exist or user {user} is not a member."
+            err_detail = (
+                f"Account {account_id} does not exist or user {request.uid} "
+                "is not a member."
+            )
             return ResponseError(NotFoundError(detail=err_detail)).response
 
         try:
--- a/server/athenian/api/controllers/miners/github/contributors.py
+++ b/server/athenian/api/controllers/miners/github/contributors.py
@@ -1,8 +1,8 @@
 import asyncio
+from collections import defaultdict
 from datetime import datetime
-from itertools import chain
 import marshal
-from typing import Collection, List, Optional, Tuple
+from typing import Collection, List, Optional
 
 import aiomcache
 import databases
@@ -13,6 +13,20 @@ from athenian.api.models.metadata.github import PullRequest, PullRequestComment,
     PullRequestReview, PushCommit, Release, User
 
 
+async def mine_contributors(repos: Collection[str],
+                            db: databases.Database,
+                            time_from: Optional[datetime] = None,
+                            time_to: Optional[datetime] = None,
+                            with_stats: Optional[bool] = True,
+                            cache: Optional[aiomcache.Client] = None) -> List[dict]:
+    """Discover developers who made any important action in the given repositories and \
+    in the given time frame."""
+    time_from = time_from or datetime(1970, 1, 1)
+    time_to = time_to or datetime.now()
+
+    return await _mine_contributors(repos, time_from, time_to, db, with_stats, cache)
+
+
 @cached(
     exptime=5 * 60,
     serialize=marshal.dumps,
@@ -20,14 +34,12 @@ from athenian.api.models.metadata.github import PullRequest, PullRequestComment,
     key=lambda repos, time_from, time_to, **_: (
         ",".join(repos), time_from.timestamp(), time_to.timestamp()),
 )
-async def mine_contributors(repos: Collection[str],
-                            time_from: datetime,
-                            time_to: datetime,
-                            db: databases.Database,
-                            cache: Optional[aiomcache.Client],
-                            ) -> Tuple[dict, dict]:
-    """Discover developers who made any important action in the given repositories and \
-    in the given time frame."""
+async def _mine_contributors(repos: Collection[str],
+                             time_from: datetime,
+                             time_to: datetime,
+                             db: databases.Database,
+                             with_stats: bool,
+                             cache: Optional[aiomcache.Client]) -> List[dict]:
     assert isinstance(time_from, datetime)
     assert isinstance(time_to, datetime)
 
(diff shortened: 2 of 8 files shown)


server/athenian/api/controllers/contributors_controller.py
```diff
@@ -7,4 +14,29 @@ async def get_contributors(request: web.Request, id: int) -> web.Response:
...
     """
-    return web.Response(status=200)
+    user = await request.user()
+
+    async with request.sdb.connection() as sdb_conn:
+        account_id = await sdb_conn.fetch_val(
+            select([UserAccount.account_id])
+            .where(and_(UserAccount.user_id == user.id, UserAccount.account_id == id)))
+        if account_id is None:
+            err_detail = f"Account {account_id} does not exist or user {user} is not a member."
```
se7entyse7en, line 24: I'd need to do a separate query to distinguish a 403 (member not part of account) and 404 (account doesn't exist). I saw that we do the same in other endpoints and I think that it doesn't worth the effort now to distinguish them \cc @dpordomingo

server/tests/controllers/test_contributors_controller.py
```diff
@@ -0,0 +1,75 @@
...
+import pytest
+
+
+@pytest.mark.parametrize("cached", [False, True], ids=["no cache", "with cache"])
+async def test_get_contributors_as_admin(client, cached, headers, app, cache):
+    await _test_get_contributors(client, cached, headers, app, cache)
+
+
+@pytest.mark.parametrize("cached", [False, True], ids=["no cache", "with cache"])
+async def test_get_contributors_as_non_admin(client, cached, headers, app, cache, eiso):
```
se7entyse7en, line 12: @vmarkovtsev At the beginning it wasn't very clear to me that by just passing `eiso` would have changed the default user.

vmarkovtsev: Yeah, it is the only way to avoid calling Auth0 in the unit tests.

se7entyse7en: yup, I mean that probably it could be more explicit than just passing a variable to a function that is then never used, at least to me is counter-intuitive 😄.

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -95,3 +96,70 @@ async def fetch_releases():
...
+    exptime=5 * 60,
+    serialize=marshal.dumps,
+    deserialize=marshal.loads,
+    key=lambda repos, **_: (",".join(repos),),
+)
+async def mine_all_contributors(repos: Collection[str],
+                                db: databases.Database,
+                                cache: Optional[aiomcache.Client]) -> List[dict]:
+    """Discover all developers who made any important action in the given repositories."""
+    async def fetch_prs_authors():
```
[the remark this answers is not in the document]
se7entyse7en: I was thinking about it, then I decided to make it separate because in addition to not having to filter indeed by timestamp. But it's true that they're actually the same if we put min `datetime` and `now` since I can re-use the `groupby` instead of distinct. I'll work on it 👍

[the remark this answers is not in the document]
se7entyse7en: done [here](https://github.com/athenianco/athenian-api/pull/344/commits/3fb18a0957309d435ef95024163d4637e3798e48).

server/athenian/api/controllers/contributors_controller.py
```diff
@@ -7,4 +14,29 @@ async def get_contributors(request: web.Request, id: int) -> web.Response:
     :param id: Numeric identifier of the account.
 
     """
-    return web.Response(status=200)
+    user = await request.user()
```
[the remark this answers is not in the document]
se7entyse7en: done [here](https://github.com/athenianco/athenian-api/pull/344/commits/c74b4798400b9f9408c9656e05f056e341d5b00e).

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -12,21 +13,33 @@
...
+async def mine_contributors(repos: Collection[str],
+                            db: databases.Database,
+                            time_from: Optional[datetime] = None,
+                            time_to: Optional[datetime] = None,
+                            with_stats: Optional[bool] = True,
+                            cache: Optional[aiomcache.Client] = None) -> List[dict]:
+    """Discover developers who made any important action in the given repositories and \
+    in the given time frame."""
+    time_from = time_from or datetime(1970, 1, 1)
+    time_to = time_to or datetime.now()
```
vmarkovtsev, line 25:
```suggestion
    time_from = time_from or datetime(1970, 1, 1, tzinfo=timezone.utc)
    time_to = time_to or datetime.now(timezone.utc)
```

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -81,17 +94,25 @@ async def fetch_releases():
...
+        select([getattr(User, k) for k in keys])
+        .where(User.login.in_(stats.keys())))
+
+    indexed_keys = list(enumerate(keys))
+    contribs = []
+    for ud in user_details:
+        c = {k: ud[i] for i, k in indexed_keys}
+        if with_stats:
+            c = {**c, "stats": stats[c["login"]]}
+
```
vmarkovtsev, line 115:
```suggestion
```

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -81,17 +94,25 @@ async def fetch_releases():
...
+    stats = defaultdict(dict)
     for rows, key in zip(data, ("prs", "commenter", "commit_author", "commit_committer",
                                 "reviewer", "releaser")):
         for row in rows:
-            users.setdefault(row[0], {})[key] = row[1]
-    try:
-        del users[None]
-    except KeyError:
-        pass
```
vmarkovtsev, line 92: This is a real case in production that is unfortunately not covered in the tests. We should return it back.

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -81,17 +94,25 @@ async def fetch_releases():
...
+            stats[row[0]][key] = row[1]
+
+    keys = ["login", "email", "avatar_url", "name"]
     user_details = await db.fetch_all(
-        select([User.login, User.avatar_url, User.name])
-        .where(User.login.in_(users)))
-    user_details = {r[0]: (r[1], r[2]) for r in user_details}
-    return users, user_details
+        select([getattr(User, k) for k in keys])
```
vmarkovtsev, line 106: Can you please return back `[User.login, User.avatar_url, User.name]`, etc. It is super important for future refactoring.

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -81,17 +94,25 @@ async def fetch_releases():
...
+        .where(User.login.in_(stats.keys())))
+
+    indexed_keys = list(enumerate(keys))
+    contribs = []
+    for ud in user_details:
+        c = {k: ud[i] for i, k in indexed_keys}
```
vmarkovtsev, line 112:
```suggestion
        c = {k: d for k, d in zip(keys, ud)}
```
so you don't have to materialize `list(enumerate(keys))`

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -81,17 +94,25 @@ async def fetch_releases():
     data = await asyncio.gather(
         fetch_prs(), fetch_comments(), fetch_commit_authors(), fetch_commit_committers(),
         fetch_reviews(), fetch_releases())
-    users = {}
+
+    stats = defaultdict(dict)
```
vmarkovtsev, line 98: `marshal` does not support `defaultdict`, so I used a regular dict here. We don't expose it outside now so OK.

server/tests/controllers/test_contributors_controller.py
```diff
@@ -0,0 +1,75 @@
...
+    response = await client.request(
+        method="GET", path="/v1/contributors/1", headers=headers,
+    )
+
+    assert response.status == 200
+
+    contribs = json.loads((await response.read()).decode("utf-8"))
+
+    assert len(contribs) == 206
+    assert len(set(c["login"] for c in contribs)) == len(contribs)
```
vmarkovtsev, line 29: Can you please use `Contributor.from_dict()` here instead of dictionary lookups so that we test the validity of that model.

server/athenian/api/controllers/miners/github/contributors.py
se7entyse7en: ah yeah sorry, I should have definitely seen in the changes, I replaced it `pop` and I forgot it after further rewrite 🤦.

server/athenian/api/controllers/miners/github/contributors.py
se7entyse7en: ah good point 👍

server/athenian/api/controllers/miners/github/contributors.py
se7entyse7en: I bumped into this as it will work on sqlite, but not with pg. The problem is that iter on `ud` with sqlite will iterate on values while on pg it will iterate on `keys` (col names).

server/athenian/api/controllers/miners/github/contributors.py
se7entyse7en: thanks for pointing it out 👍

se7entyse7en:
Addressed issues [here](https://github.com/athenianco/athenian-api/pull/344/commits/bd3acb72ab33f4738fa11e60af11a8544266ce97).

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -12,21 +13,33 @@
...
+
+    return await _mine_contributors(repos, time_from, time_to, db, with_stats, cache)
+
+
 @cached(
     exptime=5 * 60,
     serialize=marshal.dumps,
     deserialize=marshal.loads,
     key=lambda repos, time_from, time_to, **_: (
         ",".join(repos), time_from.timestamp(), time_to.timestamp()),
```
vmarkovtsev, line 35:
```suggestion
    key=lambda repos, time_from, time_to, with_stats, **_: (
        ",".join(repos), time_from.timestamp(), time_to.timestamp(), with_stats),
```
sorry, missed this another important one

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -12,21 +13,33 @@
...
-                            db: databases.Database,
-                            cache: Optional[aiomcache.Client],
-                            ) -> Tuple[dict, dict]:
-    """Discover developers who made any important action in the given repositories and \
-    in the given time frame."""
+async def _mine_contributors(repos: Collection[str],
+                             time_from: datetime,
+                             time_to: datetime,
+                             db: databases.Database,
+                             with_stats: bool,
```
vmarkovtsev, line 41:
```suggestion
                             with_stats: bool,
                             db: databases.Database,
```
(since you have to touch the code above - no special meaning here, just a convention)

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -81,17 +94,24 @@ async def fetch_releases():
...
-        del users[None]
-    except KeyError:
-        pass
-    user_details = await db.fetch_all(
-        select([User.login, User.avatar_url, User.name])
-        .where(User.login.in_(users)))
-    user_details = {r[0]: (r[1], r[2]) for r in user_details}
-    return users, user_details
+            stats[row[0]][key] = row[1]
+
```
vmarkovtsev, line 103:
```suggestion
```

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -81,17 +94,24 @@ async def fetch_releases():
...
+    stats.pop(None, None)
+
+    cols = [User.login, User.email, User.avatar_url, User.name]
+    user_details = await db.fetch_all(select(cols).where(User.login.in_(stats.keys())))
+
+    indexed_keys = list(enumerate(col.key for col in cols))
+    contribs = []
+    for ud in user_details:
+        c = {k: ud[i] for i, k in indexed_keys}
```
vmarkovtsev, line 112:
```suggestion
        c = dict(ud)
```
or
```suggestion
        c = dict(ud.items())
```
should work then.

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -81,17 +94,24 @@ async def fetch_releases():
...
+        if with_stats:
+            c = {**c, "stats": stats[c["login"]]}
```
vmarkovtsev, line 114:
```suggestion
            c["stats"] = stats[c[User.login.key]]
```

server/athenian/api/controllers/miners/github/contributors.py
se7entyse7en: woo this is a very big overlook from my side 😕, I added a simple test as well

server/athenian/api/controllers/miners/github/contributors.py
se7entyse7en: ahhh I was completely fooled by the string representation of the result returned by sqlite and I thought it was a simple tuple 👍 thanks!

se7entyse7en:
All fixes should be done, once approved I'll rebase and clean up history.

server/tests/controllers/test_contributors_controller.py
eiso: I am unused variable 😂

vmarkovtsev:
> once approved I'll rebase and clean up history.

aaand then I saw this :( Sorry!

se7entyse7en:
not a prob, thanks for the review!

PR #355, opened 2020-05-14
Implement the /team/create endpoint ENG-859

se7entyse7en:
This is rebased on #352.

### DIFF
--- a/server/athenian/api/controllers/team_controller.py
+++ b/server/athenian/api/controllers/team_controller.py
@@ -1,3 +1,4 @@
+from http import HTTPStatus
 from sqlite3 import IntegrityError, OperationalError
 from typing import List, Union
 
@@ -11,10 +12,11 @@ from athenian.api.models.state.models import Team
 from athenian.api.models.web import BadRequestError, CreatedIdentifier, DatabaseConflict
 from athenian.api.models.web.team_create_request import TeamCreateRequest
 from athenian.api.models.web.team_update_request import TeamUpdateRequest
+from athenian.api.request import AthenianWebRequest
 from athenian.api.response import model_response, ResponseError
 
 
-async def create_team(request: web.Request,
+async def create_team(request: AthenianWebRequest,
                       body: Union[dict, bytes] = None) -> web.Response:
     """Create a team.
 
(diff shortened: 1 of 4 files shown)


server/athenian/api/controllers/team_controller.py
```diff
@@ -14,7 +22,25 @@ async def create_team(request: web.Request,
...
+        try:
+            members = _check_members(body.members)
+        except ResponseError as e:
+            return e.response
+        t = Team(owner=account, name=name, members=members).create_defaults()
+        try:
+            tid = await sdb_conn.execute(insert(Team).values(t.explode()))
+        except (UniqueViolationError, IntegrityError, OperationalError) as err:
+            return ResponseError(DatabaseConflict(
+                detail=f"this team already exists: {str(err)}")).response
```
se7entyse7en, line 42: Currently, a team created with the same name or same numbers will return 409 and not 400 as it could be expected. If this is ok for us I'd leave this as-is since it would require parsing the error and probably handling different cases for pg and sqlite I guess. This will affect webapp depending on the error we want to show, but we could eventually iterate in a second moment. \cc @warenlg

server/athenian/api/openapi/openapi.yaml
```diff
@@ -860,12 +860,6 @@ paths:
               schema:
                 $ref: '#/components/schemas/InvalidRequestError'
           description: The request was invalid.
-        403:
```
dpordomingo, line 863: Can any user create a team?

se7entyse7en: yup as described in [ENG-726](https://athenianco.atlassian.net/browse/ENG-726):

> Adding a new team should take place in the settings where a dedicated page allows all users (both regular and admin) to manage the teams associated with the corresponding Athenian account.

vmarkovtsev:
@se7entyse7en Can you please rebase this :pray:

se7entyse7en:
> @se7entyse7en Can you please rebase this

@vmarkovtsev done and fixed tests failing for pg

se7entyse7en:
Fixed flake8 issue and rebased.

server/athenian/api/controllers/team_controller.py
warenlg: Absolutely, thanks for letting me know.

vmarkovtsev: +1

PR #357, opened 2020-05-15
Fix typo in become script

se7entyse7en:

### DIFF
--- a/server/athenian/api/become.py
+++ b/server/athenian/api/become.py
@@ -11,7 +11,7 @@ def main():
     bearer = sys.argv[1]
     user = sys.argv[2] if len(sys.argv) > 2 else ""
     if not user.startswith("github|") and user:
-        with urlopen("https://api.github.com/users/" + user) as http:
+        with urlopen("https://api.github.com/user/" + user) as http:
             user = "github|%d" % json.loads(http.read().decode())["id"]
     url = "https://api.athenian.co/v1/become"
     if user:


server/athenian/api/become.py
```diff
@@ -11,7 +11,7 @@ def main():
     bearer = sys.argv[1]
     user = sys.argv[2] if len(sys.argv) > 2 else ""
     if not user.startswith("github|") and user:
-        with urlopen("https://api.github.com/users/" + user) as http:
+        with urlopen("https://api.github.com/user/" + user) as http:
```
vmarkovtsev, line 14: I wonder how it worked (and it did work) before...

vmarkovtsev: Actually, this does not work.

```
https://api.github.com/user/eiso
```

shows an error. This code is needed to resolve users by their logins.

vmarkovtsev: I think you probably misused the script. It has two modes:
1. `become.py <token> eiso` by gh login
2. `become.py <token> 'github|12345678'` by gh id

vmarkovtsev:
Assuming resolved.

se7entyse7en:
Now I recall the problem 😄. It's because I was passing the id instead of the username and for some reason:
- https://api.github.com/users/se7entyse7en works
- https://api.github.com/user/se7entyse7en doesn't work
- https://api.github.com/users/5599208 doesn't work
- https://api.github.com/user/5599208 works

server/athenian/api/become.py
se7entyse7en: ah indeed just read now 👍

PR #359, opened 2020-05-15
Fix port for pg override instance for local testing

se7entyse7en:


se7entyse7en:
I didn't realize I was using the database that I use for local development 😅

vmarkovtsev:
Yeah, they get wiped :D

PR #445, opened 2020-06-08
Filter contribs with param

se7entyse7en:
Partially addresses [DEV-224](https://athenianco.atlassian.net/browse/DEV-224).

This PR just adds the model and uses it. The change is backward compatible.

### DIFF
--- a/server/athenian/api/models/web/filter_contributors_request.py
+++ b/server/athenian/api/models/web/filter_contributors_request.py
@@ -1,0 +1,199 @@
+from datetime import date
+from typing import List, Optional
+
+from athenian.api.models.web.base_model_ import Model
+
+
+class FilterContributorsRequest(Model):
+    """Filters for `/filter/contributors`."""
+
+    def __init__(
+            self,
+            account: Optional[int] = None,
+            date_from: Optional[date] = None,
+            date_to: Optional[date] = None,
+            timezone: Optional[int] = None,
+            in_: Optional[List[str]] = None,
+            as_: List[str] = None,
+    ):
+        """FilterContributorsRequest - a model defined in OpenAPI
+
+        :param account: The account of this FilterContributorsRequest.
+        :param date_from: The date_from of this FilterContributorsRequest.
+        :param date_to: The date_to of this FilterContributorsRequest.
+        :param timezone: The timezone of this FilterContributorsRequest.
+        :param in_: The in_ of this FilterContributorsRequest.
+        :param as_: The as_ of this FilterContributorsRequest.
+        """
+        self.openapi_types = {
+            "account": int,
+            "date_from": date,
+            "date_to": date,
+            "timezone": int,
+            "in_": List[str],
+            "as_": List[str],
+        }
+
+        self.attribute_map = {
+            "account": "account",
+            "date_from": "date_from",
+            "date_to": "date_to",
+            "timezone": "timezone",
+            "in_": "in",
+            "as_": "as",
+        }
+
+        self._account = account
+        self._date_from = date_from
+        self._date_to = date_to
+        self._timezone = timezone
+        self._in_ = in_
+        self._as_ = as_
+
+    @property
+    def account(self) -> int:
+        """Gets the account of this FilterContributorsRequest.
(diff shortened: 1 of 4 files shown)


se7entyse7en:
Sorry @vmarkovtsev ignore this. When I started implementing I realized that what I actually need is not a `with`, but something like `as`. So the query will be something like get contributors as `mergers or committers` etc..

se7entyse7en:
okay fixed 👍

vmarkovtsev:
I shall review this upon finishing with the boss bug :+1:

server/athenian/api/openapi/openapi.yaml
```diff
@@ -2146,6 +2182,18 @@ components:
           type: integer
         in:
           $ref: '#/components/schemas/RepositorySet'
+        as:
+          items:
+            enum:
```
vmarkovtsev, line 2187: We should reference `PullRequestWith` here instead of defining from scratch.

se7entyse7en: Unfortunately, I wanted to do that, but I wasn't able to. Do you have any idea on who it could be done with json schema? I tried both looking at how to define an `object` whose `properties` are actually validated using an `enum` and also whether it is possible to, on the contrary, define an `enum` from the `properties` of an `object`, but I didn't find anything.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -2146,6 +2182,18 @@ components:
...
+            - author
+            - reviewer
+            - commit_author
+            - commit_committer
+            - commenter
+            - merger
+            - releaser
+            type: string
+          type: array
```
vmarkovtsev, line 2196: Wouldn't this work?

```suggestion
        as:
          $ref: '#/components/schemas/PullRequestWith'
```

vmarkovtsev: ok I see now, nvm

server/athenian/api/models/web/filter_contributors_request.py
```diff
@@ -0,0 +1,199 @@
...
+        allowed_values = [
+            "author",
+            "reviewer",
+            "commit_author",
+            "commit_committer",
+            "commenter",
+            "merger",
+            "releaser",
+        ]  # noqa: E501
+        if not set(as_).issubset(set(allowed_values)):
```
vmarkovtsev, line 191:
```suggestion
        allowed_values = {
            "author",
            "reviewer",
            "commit_author",
            "commit_committer",
            "commenter",
            "merger",
            "releaser",
        } # noqa: E501
        if not set(as_).issubset(allowed_values):
```

server/athenian/api/models/web/filter_contributors_request.py
```diff
@@ -0,0 +1,199 @@
...
+            raise ValueError(
+                "Invalid values for `as_` [{0}], must be a subset of [{1}]".format(
```
vmarkovtsev, line 193:
```suggestion
                "Invalid values for `as_` [%s], must be a subset of [%s]" % (
```

server/athenian/api/models/web/filter_contributors_request.py
```diff
@@ -0,0 +1,199 @@
...
+                    ", ".join(map(str, set(as_) - set(allowed_values))),
```
vmarkovtsev, line 194:
```suggestion
                    ", ".join(set(as_) - allowed_values),
```

server/athenian/api/models/web/filter_contributors_request.py
```diff
@@ -0,0 +1,199 @@
...
+                    ", ".join(map(str, allowed_values)),
```
vmarkovtsev, line 195:
```suggestion
                    ", ".join(allowed_values),
```

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: Yep you are right this is not possible, I overfitted to the other requests :)

PR #447, opened 2020-06-08
Mine contribs with param

se7entyse7en:
Closes [DEV-224](https://athenianco.atlassian.net/browse/DEV-224).

This is rebased on #445.

### DIFF
--- a/server/athenian/api/controllers/filter_controller.py
+++ b/server/athenian/api/controllers/filter_controller.py
@@ -51,10 +51,17 @@ async def filter_contributors(request: AthenianWebRequest, body: dict) -> web.Re
         return ResponseError(InvalidRequestError("?", detail=str(e))).response
     repos = await _common_filter_preprocess(filt, request)
     users = await mine_contributors(
-        repos, filt.date_from, filt.date_to, request.mdb, request.cache)
+        repos, filt.date_from, filt.date_to, request.mdb, request.cache,
+        as_roles=filt.as_)
     model = [
-        DeveloperSummary(login=f"{PREFIXES['github']}{u['login']}", avatar=u["avatar_url"],
-                         name=u["name"], updates=DeveloperUpdates(**u["stats"]))
+        DeveloperSummary(
+            login=f"{PREFIXES['github']}{u['login']}", avatar=u["avatar_url"],
+            name=u["name"], updates=DeveloperUpdates(**{
+                k: v for k, v in u["stats"].items()
+                # TODO(se7entyse7en): make `DeveloperUpdates` support all the stats we can get instead of doing this filtering. See also `mine_contributors`.  # noqa
+                if k in DeveloperUpdates.openapi_types
+            }),
+        )
         for u in sorted(users, key=operator.itemgetter("login"))
     ]
     return model_response(model)
--- a/server/athenian/api/controllers/miners/github/contributors.py
+++ b/server/athenian/api/controllers/miners/github/contributors.py
@@ -10,10 +10,8 @@ import sentry_sdk
 from sqlalchemy import and_, func, or_, select
 
 from athenian.api.cache import cached
-from athenian.api.models.metadata.github import (PullRequest,
-                                                 PullRequestComment,
-                                                 PullRequestReview, PushCommit,
-                                                 Release, User)
+from athenian.api.models.metadata.github import (PullRequest, PullRequestComment,
+                                                 PullRequestReview, PushCommit, Release, User)
 from athenian.api.tracing import sentry_span
 
 
@@ -66,6 +64,7 @@ async def _mine_contributors(repos: Collection[str],
 
     @sentry_span
     async def fetch_author():
+        # TODO(vmarkovtsev): load released PRs from the pdb
         return await db.fetch_all(
             select([PullRequest.user_login, func.count(PullRequest.user_login)])
             .where(common_prs_where)
@@ -123,10 +122,13 @@ async def _mine_contributors(repos: Collection[str],
                             "commenter", "merger", "releaser")
     locals_ = locals()
     fetchers = {k: locals_[f"fetch_{k}"]() for k in as_roles}
-    data = await asyncio.gather(*fetchers.values())
+    data = await asyncio.gather(*fetchers.values(), return_exceptions=True)
     stats = defaultdict(dict)
-    for rows, key in zip(data, fetchers.keys()):
-        for row in rows:
+    for r, key in zip(data, fetchers.keys()):
+        if isinstance(r, Exception):
+            raise r from None
+
+        for row in r:
             stats[row[0]][key] = row[1]
 
     stats.pop(None, None)
--- a/server/athenian/api/controllers/miners/github/contributors.py
+++ b/server/athenian/api/controllers/miners/github/contributors.py
@@ -30,35 +30,9 @@ async def mine_contributors(repos: Collection[str],
     time_from = time_from or datetime(1970, 1, 1, tzinfo=timezone.utc)
     now = datetime.now(timezone.utc) + timedelta(days=1)
     time_to = time_to or datetime(now.year, now.month, now.day, tzinfo=timezone.utc)
+    as_roles = as_roles or []
 
-    if as_roles:
-        fetch_stats = True
-    else:
-        fetch_stats = with_stats
-
-    contribs = await _mine_contributors(repos, time_from, time_to, fetch_stats, db, cache)
-    if not as_roles:
-        return contribs
-
-    # Doing this in post-processing to avoid decreasing cache hits. Otherwise we could
-    # consider caching the different `fetch_*` functions indidually as a inner caching
-    # layer if this becomes a bottleneck.
-    filtered_contribs = []
-    # We could get rid of these re-mapping, maybe worth looking at it along with the
-    # definition of `DeveloperUpdates`
-    role_mapping = {"author": "prs"}
-    for c in contribs:
-        has_active_role = sum(c["stats"].get(role_mapping.get(role, role), 0)
-                              for role in as_roles) > 0
-        if not has_active_role:
-            continue
-
-        if not with_stats:
-            c.pop("stats")
-
-        filtered_contribs.append(c)
-
-    return filtered_contribs
+    return await _mine_contributors(repos, time_from, time_to, with_stats, as_roles, db, cache)
 
 
 @sentry_span
@@ -66,13 +40,15 @@ async def mine_contributors(repos: Collection[str],
     exptime=5 * 60,
     serialize=marshal.dumps,
     deserialize=marshal.loads,
-    key=lambda repos, time_from, time_to, with_stats, **_: (
-        ",".join(repos), time_from.timestamp(), time_to.timestamp(), with_stats),
+    key=lambda repos, time_from, time_to, with_stats, as_roles, **_: (
+        ",".join(repos), time_from.timestamp(), time_to.timestamp(), with_stats,
+        sorted(as_roles)),
 )
 async def _mine_contributors(repos: Collection[str],
                              time_from: datetime,
                              time_to: datetime,
                              with_stats: bool,
+                             as_roles: List[str],
                              db: databases.Database,
                              cache: Optional[aiomcache.Client]) -> List[dict]:
     assert isinstance(time_from, datetime)
@@ -143,6 +119,10 @@ async def _mine_contributors(repos: Collection[str],
                         Release.published_at.between(time_from, time_to)))
             .group_by(Release.author))
 
+    def strip_stats(contribs):
+        for c in contribs:
+            c.pop("stats")
+
     data = await asyncio.gather(
         fetch_authors(), fetch_reviewers(),
         fetch_commit_authors(), fetch_commit_committers(),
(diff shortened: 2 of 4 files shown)


server/athenian/api/controllers/filter_controller.py
```diff
@@ -42,16 +43,24 @@
...
+        with_roles=filt.as_)
     model = [
-        DeveloperSummary(login=f"{PREFIXES['github']}{u['login']}", avatar=u["avatar_url"],
-                         name=u["name"], updates=DeveloperUpdates(**u["stats"]))
+        DeveloperSummary(
+            login=f"{PREFIXES['github']}{u['login']}", avatar=u["avatar_url"],
+            name=u["name"], updates=DeveloperUpdates(**{
+                k: v for k, v in u["stats"].items()
+                # It worth make `DeveloperUpdates` support all the stats we can get
+                # instead of doing this filtering. See also `mine_contributors`.
```
vmarkovtsev, line 60:
```suggestion
                # TODO(se7entyse7en): make `DeveloperUpdates` support all the stats we can get instead of doing this filtering. See also `mine_contributors`.  # noqa
```

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -21,14 +21,42 @@ async def mine_contributors(repos: Collection[str],
                             time_to: Optional[datetime],
                             db: databases.Database,
                             cache: Optional[aiomcache.Client],
-                            with_stats: Optional[bool] = True) -> List[dict]:
+                            with_stats: Optional[bool] = True,
+                            with_roles: List[str] = None) -> List[dict]:
```
vmarkovtsev, line 25: The model has `as` while we use `with` here. We should stick to one variant of naming everywhere. How about `as_roles`

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -21,14 +21,42 @@ async def mine_contributors(repos: Collection[str],
...
     """Discover developers who made any important action in the given repositories and \
     in the given time frame."""
     time_from = time_from or datetime(1970, 1, 1, tzinfo=timezone.utc)
     now = datetime.now(timezone.utc) + timedelta(days=1)
     time_to = time_to or datetime(now.year, now.month, now.day, tzinfo=timezone.utc)
 
-    return await _mine_contributors(repos, time_from, time_to, with_stats, db, cache)
+    if with_roles:
+        fetch_stats = True
```
vmarkovtsev, line 33: I think we should split the function into two at this point, the logic is getting complex.

1. The old one that is used in `/contributors`
2. The new one with blows and whistles.

Some code will be duplicated, and this is fine.

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -21,14 +21,42 @@ async def mine_contributors(repos: Collection[str],
...
+    if not with_roles:
+        return contribs
+
+    # Doing this in post-processing to avoid decreasing cache hits. Otherwise we could
+    # consider caching the different `fetch_*` functions indidually as a inner caching
+    # layer if this becomes a bottleneck.
+    filtered_contribs = []
+    # We could get rid of these re-mapping, maybe worth looking at it along with the
+    # definition of `DeveloperUpdates`
+    role_mapping = {"author": "prs"}
```
vmarkovtsev, line 47: Let's hardcode this for now, I don't expect any other deviations between the models.

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -21,14 +21,42 @@ async def mine_contributors(repos: Collection[str],
...
+        fetch_stats = with_stats
+
+    contribs = await _mine_contributors(repos, time_from, time_to, fetch_stats, db, cache)
+    if not with_roles:
+        return contribs
+
+    # Doing this in post-processing to avoid decreasing cache hits. Otherwise we could
+    # consider caching the different `fetch_*` functions indidually as a inner caching
+    # layer if this becomes a bottleneck.
+    filtered_contribs = []
```
vmarkovtsev, line 44: ok, we should run isolated queries and ignore the cache here. Some calls are more expensive than the others, we don't expect the webapp to use them now, and a simple cache key based on the user query should perfectly do the job while reducing the complexity. In other words, this is not `PullRequestMiner`.

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -48,27 +76,30 @@ async def _mine_contributors(repos: Collection[str],
     assert isinstance(time_from, datetime)
     assert isinstance(time_to, datetime)
 
+    pr_where_condition = and_(
```
vmarkovtsev, line 79:
```suggestion
    common_prs_where = and_(
```

server/athenian/api/controllers/miners/github/contributors.py
se7entyse7en: I'd ask @warenlg's input here because I seem to recall that he recently mentioned at some point we will need to also filter/sort the contributors as well when creating/updating a team.

warenlg: Yes we need one stat to sort the users listed in the breakdown of the team set up page https://athenianco.atlassian.net/browse/DEV-112

se7entyse7en: Okay so @vmarkovtsev I'll keep the functionality in common and I'll refactor it according to what you said [here](https://github.com/athenianco/athenian-api/pull/447#discussion_r437198072).

vmarkovtsev:
@se7entyse7en I don't want to steal the opportunity to do some Python coding from you :smile: I've merged #445 so you can rebase and resume the work.

server/athenian/api/controllers/miners/github/contributors.py
se7entyse7en: What do you mean exactly with hardcoding it?  You mean something like:
```
        has_active_role = sum(c["stats"].get(role if role != "author" else "prs", 0)
                              for role in as_roles) > 0
```
it seems less readable to me.

vmarkovtsev: This is both faster and more obvious:
```python
for contrib in contribs:
    role_stats = contrib["stats"]
    has_active_role = False
    for role in as_roles:
        if role == "author":
            role = "prs"
        if role_stats.get(role, 0) > 0:
            has_active_role = True
            break
```

se7entyse7en: Just pushed, I've done something similar.

se7entyse7en: Sorry just pushed a simplification [here](https://github.com/athenianco/athenian-api/pull/447/commits/3cffae66994eab4d416ff1fd515e588dd11d91c0) just in case you started reviewing in the meantime.

se7entyse7en:
@vmarkovtsev what about using `isort`?

vmarkovtsev:
`isort`? :thinking:

se7entyse7en:
https://github.com/timothycrosley/isort

vmarkovtsev:
It is possible to use whatever locally if it helps to pass the style check :+1:

se7entyse7en:
Do you know why rules exactly it follows? So that i at least [configure it locally](https://github.com/timothycrosley/isort#configuring-isort). Tomorrow I’ll try maybe the default would work already.

vmarkovtsev:
The rules are called appnexus: https://github.com/athenianco/athenian-api/blob/master/server/.flake8#L5

https://softwareengineering.stackexchange.com/a/386891

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -21,94 +23,109 @@ async def mine_contributors(repos: Collection[str],
...
+        or_(PullRequest.created_at.between(time_from, time_to),
+            and_(PullRequest.created_at < time_to,
+                 PullRequest.closed_at.is_(None)),
+            PullRequest.closed_at.between(time_from, time_to),
+            PullRequest.updated_at.between(time_from, time_to)),
+    )
+
     @sentry_span
-    async def fetch_prs():
+    async def fetch_author():
```
vmarkovtsev, line 68: Can you please insert `# TODO(vmarkovtsev): load released PRs from the pdb`

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -21,94 +23,109 @@ async def mine_contributors(repos: Collection[str],
...
 
-    data = await asyncio.gather(
-        fetch_prs(), fetch_comments(), fetch_commit_authors(), fetch_commit_committers(),
-        fetch_reviews(), fetch_releases())
-
+    as_roles = as_roles or ("author", "reviewer", "commit_author", "commit_committer",
+                            "commenter", "merger", "releaser")
+    locals_ = locals()
+    fetchers = {k: locals_[f"fetch_{k}"]() for k in as_roles}
+    data = await asyncio.gather(*fetchers.values())
```
vmarkovtsev, line 126: We can do this with less hacks :smile: 
```suggestion
    @sentry_span
    async def fetch_author():
        return "prs", await db.fetch_all(...)

    tasks = [fetch_author(), ..., fetch_releaser()]
    data = await asyncio.gather(*tasks, return_exceptions=True)
    for r in data:
        if isinstance(r, Exception):
            raise r from None
    for key, rows in data:
        ...
```

server/athenian/api/controllers/miners/github/contributors.py
```diff
@@ -21,94 +23,109 @@ async def mine_contributors(repos: Collection[str],
...
     assert isinstance(time_to, datetime)
 
+    common_prs_where = and_(
+        PullRequest.repository_full_name.in_(repos),
+        PullRequest.hidden.is_(False),
+        or_(PullRequest.created_at.between(time_from, time_to),
+            and_(PullRequest.created_at < time_to,
+                 PullRequest.closed_at.is_(None)),
+            PullRequest.closed_at.between(time_from, time_to),
+            PullRequest.updated_at.between(time_from, time_to)),
```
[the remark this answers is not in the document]
se7entyse7en: Is this a new product requirement? It totally makes sense to me, indeed when I was debugging this with the webapp there were some contributors being returned I didn't understand why all their PRs where closed before the `time_from`, then I realized it was because some PRs have been updated in the time interval (a label added).

\cc @warenlg

server/athenian/api/controllers/miners/github/contributors.py
se7entyse7en: The idea here is to avoid calling specific `fetch_*` functions if they're not necessary, that is a `fetch_X` function is called if `X` is in `as_roles` or if `as_roles` is empty. Anyway, I can achieve the same without using `locals()` if this is what bothers you 👍, but I'd still need the mapping between the role name and the function and to filter the functions to call depending on the `as_roles` parameter.

I'll add the `return_exceptions` 👍

se7entyse7en:
> The rules are called appnexus: https://github.com/athenianco/athenian-api/blob/master/server/.flake8#L5

> https://softwareengineering.stackexchange.com/a/386891

Thanks! I think I found the config to use. Here it is in case you wanna give it a try as well:

```
// content of `.isort.cfg`
[settings]
line_length=99
force_sort_within_sections=true
```

server/athenian/api/controllers/miners/github/contributors.py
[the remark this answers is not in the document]
vmarkovtsev: This is how the rest of the miners work atm. I used to have `updated_at` everywhere in the past, and it was a source of inconsistencies. Just forgot to remove it here.

server/athenian/api/controllers/miners/github/contributors.py
vmarkovtsev: I am against any `locals()` here. The code becomes hostile to refactorings, IDE intellisense, etc. Explicit is better than implicit. Besides, this is closer to the Go way.

vmarkovtsev: As you see, you don't need the mapping if you return it from each function.

vmarkovtsev:
Thanks! PyCharm does everything without having to configure the rules (among many other things).

server/athenian/api/controllers/miners/github/contributors.py
se7entyse7en: > I am against any locals() here. The code becomes hostile to refactorings, IDE intellisense, etc. Explicit is better than implicit. Besides, this is closer to the Go way.

👍 

> As you see, you don't need the mapping if you return it from each function.

I cannot make it return from each function because I need to filter them before calling them in order to avoid fetching unnecessary things.

server/athenian/api/controllers/filter_controller.py
```diff
@@ -49,10 +49,17 @@ async def filter_contributors(request: AthenianWebRequest, body: dict) -> web.Re
...
+        as_roles=filt.as_)
     model = [
-        DeveloperSummary(login=f"{PREFIXES['github']}{u['login']}", avatar=u["avatar_url"],
-                         name=u["name"], updates=DeveloperUpdates(**u["stats"]))
+        DeveloperSummary(
+            login=f"{PREFIXES['github']}{u['login']}", avatar=u["avatar_url"],
+            name=u["name"], updates=DeveloperUpdates(**{
+                k: v for k, v in u["stats"].items()
+                # TODO(se7entyse7en): make `DeveloperUpdates` support all the stats we can get instead of doing this filtering. See also `mine_contributors`.  # noqa
+                if k in DeveloperUpdates().openapi_types
```
vmarkovtsev, line 60:
```suggestion
                if k in DeveloperUpdates.openapi_types
```
(I have just changed all the models by hand, one hour of manual work :man_facepalming:)

se7entyse7en: thanks for that! Indeed it was weird

se7entyse7en:
done and rebased 👍

PR #495, opened 2020-06-24
Define default JIRA_INSTALLATION_URL tpl in docker-compose

dpordomingo:
Define default `ATHENIAN_JIRA_INSTALLATION_URL_TEMPLATE` in docker-compose

It will be used also in pre-staging envs.
If the user already uses a `.env` file, it will be used instead of this default

## TL;DR

I locally tested it this way:
```
$ docker-compose up
$ docker-compose exec -T api /bin/bash
echo $ATHENIAN_JIRA_INSTALLATION_URL_TEMPLATE
```
The value from local `.env` is kept.
The value from `docker-compose.yml::environment` is only used if the variable is not in the local `.env`

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -22,6 +22,8 @@ services:
             --precomputed-db=postgresql://${POSTGRES_USER:-api}:${POSTGRES_PASSWORD:-api}@postgres:5432/precomputed \
             --ui"
     env_file: .env
+    environment:
+      - ATHENIAN_JIRA_INSTALLATION_URL_TEMPLATE=${ATHENIAN_JIRA_INSTALLATION_URL_TEMPLATE:-https://installation.athenian.co/jira/%s/atlassian-connect.json}
     ports:
       - ${API_HOST_PORT:-8080}:8080
     depends_on:


dpordomingo:
I'm not sure that the default values should be provided by an optional file containing user values. If it's done this way, everytime a new env value is required, it's needed to re-generate that `.env` file, which can be a pain when handling many different environments like pre-staging, staging, prod...

The current behavior caused pre-staging to unexpectedly break, and its image had to be manually regenerated. 

Anyhow, @lwsanty is in charge of managing that envs, so if he prefers to be in charge of maintaining all its env values whenever they're required, I'll be ok.

vmarkovtsev:
We just need to update the pre-staging chart with new envvars, like we already do for both staging and prod. There is no magic.

lwsanty:
I need to set them anyways in `.env` or `/etc/environments` so currently I have fixed everything on my side in `.env`

dpordomingo:
With this PR, the default values of new env vars will be gracefully handled by the API, instead of failing and forcing the user to define them whenever it's required.
If you are both ok on breaking the API if the user didn't confirm an eventually new default value, feel free to close this and #496 :wink:

vmarkovtsev:
So the resolution is: do not merge PRs to staging until Olek adds the new environment variable.

PR #497, opened 2020-06-25
[DEV-347] Add @linc and @linear-app to the list of bots

vmarkovtsev:


server/athenian/api/controllers/miners/github/hardcoded.py
```diff
@@ -5,4 +5,6 @@
     "codecov-commenter",
     "dependabot",
     "dependabot-preview",
+    "linc",
+    "linear-app",
```
vmarkovtsev, line 9: @warenlg Bots are easy to add, so you can do it yourself when I go on vacation.

PR #502, opened 2020-06-26
Print a repository count in the slack messages

dennwc:


vmarkovtsev:
There is a flaky test inside, I will fix it later.

PR #504, opened 2020-06-26
[DEV-368] Feature flags

vmarkovtsev:


server/athenian/api/models/state/models.py
```diff
@@ -167,3 +167,28 @@ class ReleaseSetting(create_time_mixin(updated_at=True), Base):
     branches = Column(String(1024))
     tags = Column(String(1024))
     match = Column(SmallInteger())
+
+
+class Feature(create_time_mixin(updated_at=True), Base):
```
se7entyse7en, line 172: Do you think it would need to have also a `params` column? Or you think that it would be better to make a feature flat so that different params would be equivalent to different features? E.g.: a feature flag for limiting the number of teams. It could either a feature named `limit_teams_to_5` or `limt_teams` with a param column `{value: 5}`. If we want to just change the value we just need to update the db, in the other case, I'd expect the code somewhere to have something like:
```
if feature_enabled("limit_teams_to_5") and len(teams) >= 5:
    raise
```
so it would require changing the code.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -52,11 +52,11 @@ paths:
...
       x-openapi-router-controller: athenian.api.controllers.user_controller
       x-codegen-request-body-name: body
-  /account/{id}:
+  /account/{id}/members:
     get:
-      operationId: get_account
+      operationId: get_account_members
       parameters:
-      - description: Numeric identifier of the account to list. The user must belong
+      - description: Numeric identifier of the account. The user must belong
```
se7entyse7en, line 59: Can we first keep the other endpoint as well so that we do not force webapp to update? This is a breaking change isn't it? This would require bumping api and webapp together from staging to production and the non-atomic rolling update would cause some errors in case of traffic during the deploy.

server/athenian/api/models/state/models.py
```diff
@@ -167,3 +167,28 @@ class ReleaseSetting(create_time_mixin(updated_at=True), Base):
...
+    """Product features."""
+
+    __tablename__ = "features"
+    __table_args__ = (UniqueConstraint("name", "component", name="uc_feature_name_component"),
+                      {"sqlite_autoincrement": True})
+
+    id = Column(Integer(), primary_key=True)
+    name = Column(String(128), nullable=False)
+    component = Column(String(64), nullable=False)
```
se7entyse7en, line 181: Would it worth using an enum type? That would also make the column more self-documented.

server/athenian/api/models/state/models.py
vmarkovtsev: Let's a new column `parameters` of type JSON. It should be per-account rather than global though.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: ok, I'll add a redirection :+1:

server/athenian/api/models/state/models.py
vmarkovtsev: ok

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: It appears hard to add a redirection for `{id}`-like paths. We will have to deploy together :(

vmarkovtsev:
@se7entyse7en How about now?

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: Added via duplicating the endpoint in the spec + calling the new handler from the old one in Python.

vmarkovtsev:
Now I remember @se7entyse7en: the webapp is not using `/account/{id}` this is why I coded it backward-incompatible. I will PR the removal now.

vmarkovtsev:
https://github.com/athenianco/athenian-api/pull/508

PR #506, opened 2020-06-27
[DEV-378] API Keys authentication

vmarkovtsev:
This PR adds the second authentication method to the API, "apiKey" in OpenAPI terms.

The server does not store the tokens. Instead, it encrypts the token ID using [Google Key Management Service](https://cloud.google.com/kms) and sends it to the user, similar to how GitHub Personal Access Tokens work.
When the user supplies the `X-API-Key` header, the token is decrypted with KMS and checked for validity.
There are no expiration restrictions by design.

Not all the endpoints allow "apiKey" auth: I marked only read-only ones such as metrics, filters, etc.


vmarkovtsev:
@lwsanty We need to define 4 new environment variables in CI and pre-staging and 3 in staging and prod.

Go to Security -> Cryptographic Keys:
image
Enable that API.

Create a new keyring ("CREATE KEY RING") named "PAT". Leave Location set to the default value ("global"). Create three new keys named:

1. `CI`
2. `staging`
3. `production`

Set "Rotation period" to "Never (manual rotation)", the rest are defaults.

Create a new service account in that project ("IAM & Admin" -> "Service Accounts"). There should be no permissions, we will set them using the command line. Name it `CI-KMS`. Then run this command ([docs](https://cloud.google.com/kms/docs/iam#kms-add-iam-member-cli)):

```
gcloud kms keys add-iam-policy-binding CI --keyring PAT --location global --member serviceAccount:CI-KMS-355@athenian-1.iam.gserviceaccount.com --role roles/cloudkms.cryptoKeyEncrypterDecrypter
```

Replace `355` with the actual service account ID. Then create a new JSON key for that service account and copy it.

Go Travis project settings and set:

1. `GOOGLE_KMS_PROJECT` to `athenian-1`
2. `GOOGLE_KMS_KEYRING` to `PAT`
3. `GOOGLE_KMS_KEYNAME` to `CI`
4. `GOOGLE_KMS_SERVICE_ACCOUNT_JSON_INLINE` to the copied JSON.

server/athenian/api/auth.py
```diff
@@ -189,13 +191,14 @@ def __enter__(self):
         def verify_security(auth_funcs, required_scopes, function):
             @functools.wraps(function)
             async def wrapper(request: ConnexionRequest):
-                if "Authorization" not in request.headers or self.force_user:
+                has_auth = "Authorization" in request.headers or "X-API-Key" in request.headers
+                if not has_auth or self.force_user:
                     # Otherwise we will never reach self.extract_token and self._set_user
                     request.headers = CIMultiDict(request.headers)
                     request.headers["Authorization"] = "Bearer null"
                 token_info = get_authorization_info(auth_funcs, request, required_scopes)
```
se7entyse7en, line 199: I think it would be cleaner to do as follows so that you don't need to check for specific headers since the security schemes are actually defined in the yaml and handled by `connexion`:

```
    def __enter__(self):
        """Monkey-patch connexion.operations.secure.verify_security()."""
        self._verify_security = secure.verify_security

        def get_auth_info(auth_funcs, request, required_scopes):
            fallback_token = {"token": null, "method": "bearer"}  # this will be still handled correctly by `self._extract_bearer_token`
            if self.force_user:  # no need to get real authorization info
                return fallback_token

            try:
                return get_authorization_info(auth_funcs, request, required_scopes)
            except OAuthProblem:  # nor "Authorization" nor "X-API-Key" headers provided, but no need to check them manullay
                return fallback_token

        def verify_security(auth_funcs, required_scopes, function):
            @functools.wraps(function)
            async def wrapper(request: ConnexionRequest):
                token_info = get_auth_info(auth_funcs, request, required_scopes)
                # token_info = {"token": <token>, "method": <method>} at this point, now do the real work
                await self._set_user(request.context, **token_info)
                return await function(request)
            return wrapper

        secure.verify_security = verify_security
        return self
```

Maybe even

server/athenian/api/auth.py
```diff
@@ -347,32 +350,44 @@ async def _get_user_info_cached(self, token: str) -> User:
...
                 return await self.get_user(request.uid)
             return await self._get_user_info(token)
 
         request.user = get_user_info
 
-    async def _extract_token(self, token: str) -> Dict[str, Any]:
+    async def _extract_bearer_token(self, token: str) -> Dict[str, Any]:
         if token == "null":
             return {"sub": self.force_user or self._default_user_id}
+        # People who understand what's going on here:
```
se7entyse7en, line 388: 😂 

Anyway reviewing the code helps a lot! But I'd definitely feel uncomfortable being listed here 😂

server/athenian/api/controllers/security_controller.py
```diff
@@ -1,14 +1,135 @@
...
+        await get_user_account_status(request.uid, model.account, conn, request.cache)
+        try:
+            token = await conn.execute(insert(UserToken).values(UserToken(
+                name=model.name, account_id=model.account, user_id=request.uid,
+            ).create_defaults().explode()))
+        except (UniqueViolationError, IntegrityError, OperationalError) as e:
+            raise ResponseError(DatabaseConflict(
+                detail="Token '%s' already exists: %s: %s" % (model.name, type(e).__name__, e)),
+            ) from None
+    plaintext = struct.pack("<q", token)
```
se7entyse7en, line 66: `"<Q"` should be better since `ids` are positive.

se7entyse7en:
This was very interesting to review 👍

server/requirements.txt
```diff
@@ -18,4 +18,9 @@ aiomcache==0.6.0
 uvloop==0.14.0
 bootstrapped==0.0.2
 tqdm==4.46.1
-slackclient[optional]==2.7.1
\ No newline at end of file
+slackclient[optional]==2.7.1
+gcloud-aio-kms==3.2.0
+# gcloud-aio-kms does not pin the exact versions of the following packages
```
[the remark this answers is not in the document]
vmarkovtsev: `requirements.txt` misusage 101... https://github.com/talkiq/gcloud-aio/blob/master/auth/requirements.txt
I am also fighting to fix the installation https://github.com/talkiq/gcloud-aio/pull/160

Dockerfile
```diff
@@ -74,6 +74,7 @@ ADD databases.patch /databases.patch
 RUN apt-get update && \
     apt-get install -y --no-install-suggests --no-install-recommends python3-dev gcc g++ patch && \
     pip3 install --no-cache-dir -r /server/requirements.txt && \
+    pip3 uninstall -y typing && \
```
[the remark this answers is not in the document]
vmarkovtsev: https://github.com/talkiq/gcloud-aio/pull/160

server/athenian/api/auth.py
vmarkovtsev: haha

server/athenian/api/controllers/security_controller.py
vmarkovtsev: Gotcha! `BigInteger` and `Integer` are actually signed and you can have negative IDs - for example, if you create DB records by hand. Anyways, when we reach `1<<63` API keys we'll have other problems :D

server/athenian/api/auth.py
vmarkovtsev: We cannot do this:
```
 try:
                return get_authorization_info(auth_funcs, request, required_scopes)
            except OAuthProblem:
                return fallback_token
```
because it means: if the user provided a wrong JWT (e.g., expired), do not return 401 Unauthorized but make them @gkwillie. This is certainly not what we want :D

The current logic is: if we don't supply `Authorization` header, we are @gkwillie. Otherwise, check the token. I'll add comment.

vmarkovtsev: I added more comments.

se7entyse7en: Ahh indeed, yes sorry I didn't notice that `OAuthProblem` is not raised only [here](https://github.com/zalando/connexion/blob/f55cb1c923a2b51f402513fc6bdd3c9ebe31353a/connexion/decorators/security.py#L132), otherwise it should have worked fine. Otherwise, with this code we'd need to check for the description which is not very good IMO. It would have been better to me if `connexion` raised a different child error of `OAuthProblem` in case of no auth token provided.

server/athenian/api/controllers/security_controller.py
se7entyse7en: > for example, if you create DB records by hand.

yeah I just excluded this case as a thing to avoid 😛 , but good point and also since the schema as well is actually integer that's fine 👍 .

> Anyways, when we reach 1<<63 API keys we'll have other problems :D

😂 💯

server/athenian/api/auth.py
vmarkovtsev: This code is certainly a hack with monkey-patching, I don't like it too. I didn't see another way to do what we need back in February, unfortunately.

server/requirements.txt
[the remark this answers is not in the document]
TheKevJames: FYI: as a general rule, libraries should pin compatible version _ranges_ and apps should pin _specific_ versions: you can read a lot more about reasoning why [eg. here](https://stackoverflow.com/questions/28509481/should-i-pin-my-python-dependencies-versions) but the main advantage is that it helps app developers solve conflicts -- for example, if `slackclient[optional]` had depended on a specific older version of `backoff`, everything still would have worked out since we've already verified that all of the 1.x releases of `backoff` are compatible with `gcloud-aio`. :)

server/athenian/api/kms.py
```diff
@@ -0,0 +1,49 @@
...
+        self._kms = KMS(**evars, service_file=service_file)
+        self.log.info("Using Google KMS %(keyproject)s/%(keyring)s/%(keyname)s" % evars)
+
+    async def encrypt(self, plaintext: Union[bytes, str]) -> str:
+        """Encrypt text using Google KMS."""
+        return await self._kms.encrypt(encode(plaintext))
+
+    async def decrypt(self, ciphertext: str) -> bytes:
+        """Decrypt text using Google KMS."""
+        # we cannot use gcloud.aio.kms.decode because it converts bytes to string with str.decode()
```
TheKevJames, line 42: Hmm, that's a good point, and we should probably not be enforcing that type requirement on users, especially for what is basically just a convenience method. Will look into it :)

server/requirements.txt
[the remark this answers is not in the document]
vmarkovtsev: All is correct except that you seem to confuse `setup.py`'s `install_requires` and `requirements.txt` @TheKevJames

https://packaging.python.org/discussions/install-requires-vs-requirements/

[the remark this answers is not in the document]
vmarkovtsev: The difference is the same as between `go.mod` and `go.lock` (to lesser extent), `package.json` and `yarn.lock`, `Cargo.toml` and `Cargo.lock`, `Gemfile` and `Gemfile.lock`, etc.

server/athenian/api/kms.py
vmarkovtsev: Thank you! This seemed to little feature to raise an issue for :smile:

lwsanty:
`GOOGLE_KMS_SERVICE_ACCOUNT_B64` is set we need to decode it before the tests

server/requirements.txt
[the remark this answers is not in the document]
TheKevJames: Huh, I'm confused -- when you install `gcloud-aio-kms` here you're using the `install_requires`, which according to your instructions is meant to be version ranges, as we have it. So... from the perspective of `athenian-api` or any downstream app, you should _never_ get pinned versions from your libraries.

If I'm reading that link correctly, the only folks who should ever use the `requirements.txt` file are folks developing `gcloud-aio` itself. In which case... doesn't it still do what it should, since it let's us verify that the latest compatible versions are guaranteed compatible?

We always pin all dependencies for applications, but pinning them for libraries doesn't make sense to me -- if we had them pinned for our local developement, wouldn't that just hide potential version conflicts with newer releases and make it harder for us to detect issues before users do?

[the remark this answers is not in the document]
vmarkovtsev: Look: this is what you do in `setup.py` - this is a well known Python packaging gotcha: https://github.com/talkiq/gcloud-aio/blob/master/auth/setup.py#L24 What you should do instead is to copy your current `requirements.txt` to `setup.py` and write a new `requirements.txt` with all the versions pinned. When you `pip install`, it reads `setup.py`'s `install_requires` by default. BTW, I was the same confused fellow 7 years ago, I did it [here](https://github.com/Samsung/veles/blob/master/requirements.txt), for example. Learn from my mistakes :D

The reason why I don't pin all the nested deps in my project's `requirements.txt` is just because I am too lazy and not always `pip freeze`. BTW, `pip freeze` is exactly for this: to pin everything to your `requirements.txt`. The reason for pinning is to have reproducible builds. There are numerous cases when people are not attentive enough to semantic versioning and release breaking changes in minor or patch version bumps. The more deps you have, the more likely the chance is you'll get a broken build. Heck, even scipy has recently broken backward compatibility in 1.2, those guys took a big shit on semantic versioning. Besides, you'll not be able to test with all possible version combinations if your project is big enough. You'll get a combinatorial explosion of all possible version sets. 

> if we had them pinned for our local developement, wouldn't that just hide potential version conflicts with newer releases

This is why we have @dependabot and @snyk who automatically submit version bumping and security fixing PRs to your `requirements.txt` and `setup.py`.

PR #507, opened 2020-06-29
Skip the big boss in the heater

vmarkovtsev:


vmarkovtsev:
We will hack it with an empty reposet.

PR #533, opened 2020-07-17
Address the 32k arguments bindings query limit

se7entyse7en:
Closes [DEV-484](https://athenianco.atlassian.net/browse/DEV-484).

### DIFF
--- a/server/athenian/api/hacks/heat_cache.py
+++ b/server/athenian/api/hacks/heat_cache.py
@@ -99,6 +99,7 @@ def main():
             if progress.finished_date is None:
                 log.warning("Skipped account %d / reposet %d because the progress is not 100%",
                             reposet.owner_id, reposet.id)
+                continue
             repos = {r.split("/", 1)[1] for r in reposet.items}
             log.info("Heating reposet %d of account %d", reposet.id, reposet.owner_id)
             try:
--- a/server/athenian/api/models/__init__.py
+++ b/server/athenian/api/models/__init__.py
@@ -92,7 +92,7 @@ class AutoInAnyColumn(Column):
     # This switch also affects performance. `IN` is better for smaller values and `ANY` for larger
     # ones:
     #   - https://blog.jooq.org/2017/03/30/sql-in-predicate-with-in-list-or-with-array-which-is-faster/ # noqa: E501
-    IN_AUTO_ANY_THRESHOLD = 100  # This specific value doesn't have any specific meaning
+    IN_AUTO_ANY_THRESHOLD = 100  # This value doesn't have any specific meaning
 
     assert IN_AUTO_ANY_THRESHOLD < MAX_ALLOWED_QUERY_ARGUMENTS
 
@@ -101,7 +101,7 @@ class AutoInAnyColumn(Column):
     def in_(self, other):
         """Change the `in` operator into an `any` if too many values"""
         if len(other) > self.IN_AUTO_ANY_THRESHOLD:
-            self.log.info("automatically switching `IN` clause to `ANY`")
+            self.log.debug("automatically switching `IN` clause to `ANY`")
             # `in_` works if you pass a `dict` for example, but `any_` does not
             return self == any_(list(other))
 
@@ -110,7 +110,7 @@ class AutoInAnyColumn(Column):
     def notin_(self, other):
         """Change the `notin` operator into a `!= all` if too many values"""
         if len(other) > self.IN_AUTO_ANY_THRESHOLD:
-            self.log.info("automatically switching `NOT IN` clause to `!= ALL`")
+            self.log.debug("automatically switching `NOT IN` clause to `!= ALL`")
             # `notin_` works if you pass a `dict` for example, but `all_` does not
             return self != all_(list(other))
 
--- a/server/athenian/api/models/__init__.py
+++ b/server/athenian/api/models/__init__.py
@@ -80,7 +80,7 @@ def create_base() -> BaseType:
     return declarative_base(cls=(Refreshable, Explodable))
 
 
-class SmartColumn(Column):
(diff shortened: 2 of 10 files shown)


se7entyse7en:
@vmarkovtsev since you'll be available again on Monday I'd rather wait for you for this PR.

Anyway, I was able to locally both run the heater and render the dashboard with ES data. 👍
I first made it by manually checking the length on each query [here](https://github.com/athenianco/athenian-api/pull/533/commits/df22b5ba69f4f8ff03c8a5c87a275be8610827f1), then I made it smarter by applying it to everything and reverted that previous commit [here](https://github.com/athenianco/athenian-api/pull/533/commits/cc8d25c754bdbd33be833bf03c80aa7a296f3757).

server/athenian/api/models/__init__.py
```diff
@@ -78,6 +80,48 @@ def create_base() -> BaseType:
...
+        return super(SmartColumn, self).in_(other)
+
+    def notin_(self, other):
+        """Change the `notin` operator into a `!= all` if too many values"""
+        log = logging.getLogger("smart_column.notin_")
+        log.info("len(other) = %d", len(other))
+        if len(other) > self.IN_AUTO_ANY_THRESHOLD:
+            log.warning("switching to `!= all`")
+            # `notin_` works if you pass a `dict` for example, but `all_` does not
+            return self != all_(other.__iter__())
```
se7entyse7en, line 118: I initially used `not_(self == any_(other))`, but it seems that `SQLAlchemy` is too smart... https://stackoverflow.com/questions/62952771/sqlalchemy-how-to-negate-an-any-expression

server/athenian/api/models/__init__.py
```diff
@@ -78,6 +80,42 @@ def create_base() -> BaseType:
...
+
+    # 32767 is the maximum value for postgres:
+    #   - https://sentry.io/organizations/athenianco/issues/1785993425/
+    MAX_ALLOWED_QUERY_ARGUMENTS = 32767
+    # Anyway we cannot use exactly the upper limit because the length of the values used for
+    # `IN` and `NOT IN` could be near the the maximum allowed, but it often comes with other
+    # clauses that could make the total number of arguments bindings overflow the limit.
+    # This switch also affects performances:
+    #   - https://blog.jooq.org/2017/03/30/sql-in-predicate-with-in-list-or-with-array-which-is-faster/ # noqa: E501
+    IN_AUTO_ANY_THRESHOLD = 16383  # This specific value doesn't have any specific meaning
```
se7entyse7en, line 94: I'd like to also run all the metrics test with this value overridden to `0`, wdyt?

server/athenian/api/hacks/heat_cache.py
```diff
@@ -99,6 +99,7 @@ async def async_run():
             if progress.finished_date is None:
                 log.warning("Skipped account %d / reposet %d because the progress is not 100%",
                             reposet.owner_id, reposet.id)
+                continue
```
vmarkovtsev, line 102: 🤦‍♂️ 🤦‍♂️ 🤦‍♂️

server/athenian/api/models/__init__.py
```diff
@@ -78,6 +80,42 @@ def create_base() -> BaseType:
     return declarative_base(cls=(Refreshable, Explodable))
 
 
+class SmartColumn(Column):
+    """Column that automatically converts `IN` and `NOT IN` to `ANY` and `!= ALL` repsecitvely"""
+
+    # 32767 is the maximum value for postgres:
+    #   - https://sentry.io/organizations/athenianco/issues/1785993425/
+    MAX_ALLOWED_QUERY_ARGUMENTS = 32767
```
vmarkovtsev, line 88: Let's set it to 100 straight away. A big number of IN args results really bad performance that I saw in many profiles.

server/athenian/api/models/__init__.py
vmarkovtsev: IMHO we don't need two thresholds as soon as we set the main one to 100.

server/athenian/api/models/__init__.py
```diff
@@ -78,6 +80,42 @@ def create_base() -> BaseType:
     return declarative_base(cls=(Refreshable, Explodable))
 
 
+class SmartColumn(Column):
```
vmarkovtsev, line 83: Let's name it `AutoInAnyColumn`

se7entyse7en:
done

se7entyse7en:
Damn sqlite. 🤦 

@vmarkovtsev could you remind me why it's still relevant to test with sqlite? How we could circumvent this? Maybe the easiest is to change the threshold for sqlite tests?

vmarkovtsev:
We continue to support SQLite because it is the best for running the unit tests locally. Sometimes a test crash, the DB cleanup does not run or is poorly written, and you end up with a screwed Postgres which you have to clean by hand or even reload the dump.

vmarkovtsev:
I think we can do this: instead of coding the switch between IN and ANY inside `def in_`, it is possible to define a custom SQLAlchemy op with that switch inside and return it instead. That op can derive the threshold, but, unlike `Column`, when it renders it knows the dialect so it can override based on sqlite

vmarkovtsev:
@se7entyse7en I am taking over this PR. I'll fix SQLite one way or another and merge.

se7entyse7en:
> @se7entyse7en I am taking over this PR. I'll fix SQLite one way or another and merge.

okay thanks!

se7entyse7en:
> when it renders it knows the dialect so it can override based on sqlite

Indeed that's what I missed when I was thinking on how to solve this 👍  Cool!

vmarkovtsev:
Tests pass, but the new unit test does not.

PR #538, opened 2020-07-20
[DEV-478] Add /memory endpoint

vmarkovtsev:


server/athenian/api/controllers/status_controller.py
```diff
@@ -112,6 +115,23 @@ async def __call__(self, request: web.Request) -> web.Response:
         return resp
 
 
+async def summarize_memory(request: web.Request) -> web.Response:
```
se7entyse7en, line 118: Do we have some x-header to know which pod was it?

PR #551, opened 2020-07-22
[DEV-518] Add /histogram/prs spec

vmarkovtsev:


server/athenian/api/openapi/openapi.yaml
```diff
@@ -4039,3 +4086,78 @@ components:
...
+          type: array
+        y:
+          description: Series of histogram bar heights.
+          items:
+            type: integer
+          type: array
+      required:
+      - xmin
+      - xmax
+      - bars
```
dpordomingo, line 4162: This does not match the defined `CalculatedPullRequestHistogram` properties.
I'd also rename the properties. `x` and `y` are not that _histogram terms_, wdyt?
```suggestion
      - ticks # or also `thresholds`
      - values
```

dpordomingo: Also, if the `scale` is not part of the `PullRequestHistogramRequest` (because it will be optional), I'd send it in the `CalculatedPullRequestHistogram`.

vmarkovtsev:
> we need more metrics → @warenlg, can you confirm it?,
API should be able to return many metrics at once,
API should be able to return as many bins as requested → @warenlg, can you confirm that this number will change?,
I'd prefer to make the scale mandatory in the request.

All makes sense, I'll add these today.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -4039,3 +4086,78 @@ components:
...
+      properties:
+        x:
+          description: Series of horizontal bar borders. Their count is `len(y) + 1`
+            because there are `N` intervals between `(N + 1)` ticks.
+          items:
+# {% if False %}
+            allOf:
+# {% else %}{{ "\n" }}            oneOf:{% endif %}
+            - type: number
+            - format: date-time
```
dpordomingo, line 4151: By the way, webapp is not using datetimes yet, and I don't see its necessity in the short term.

On the other hand, time intervals would be currently needed instead.

vmarkovtsev: I meant our regular time intervals that we express as `"Xs"` in seconds,exactly. Morning brain freeze.

vmarkovtsev:
@dpordomingo How about now?

warenlg:
> we need more metrics → @warenlg, can you confirm it?

Yes, we will probably plot multiple distributions on the same chart in the near future. 

> API should be able to return as many bins as requested → @warenlg, can you confirm that this number will change?

That would be useful for UI adjustments yes.

> I'd prefer to make the scale mandatory in the request.

So far we only need it with log scale. Same foe the distribution of PR sizes. But we might need the linear scale in the future, so up to you.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -2245,7 +2297,6 @@ components:
             Sets of developers and repositories for which to calculate the metrics.
             The aggregation is `AND` between repositories and developers.
             The aggregation is `OR` inside both repositories and developers.
-            Any contribution role is matched: authors, reviewers, commenters, committers, releasers.
```
dpordomingo, line 2248: mmm... changing other things in this PR... ok...
Let me confirm:
- is this something new? or are you just updating the docs for a feature released some time ago?

server/athenian/api/openapi/openapi.yaml
```diff
@@ -4039,3 +4102,118 @@ components:
...
       type: object
+    TimeZone:
+      description: Local time zone offset in minutes, used to adjust `date_from`
+        and `date_to`.
+      minimum: -720
+      maximum: 720
+      type: integer
+    TimeDuration:
+      description: Time duration value.
+      pattern: ^\d+s$
```
dpordomingo, line 4113: Good idea!

server/athenian/api/openapi/openapi.yaml
```diff
@@ -4039,3 +4102,118 @@ components:
...
+          type: array
+        metrics:
+          description: List of desired histogram topics.
+          items:
+            $ref: '#/components/schemas/PullRequestHistogramTopic'
+          type: array
+        scale:
+          $ref: '#/components/schemas/HistogramScale'
+        bins:
+          description: Number of bars in the histogram. 0 means automatic.
```
dpordomingo, line 4135: what should we expect about the automatic mode? a default number for all the histograms? different value for each metric? does it depend on the data being processed?

wouldn't it be a good idea to document the automatic mode?

(not complaining, just trying to know what to expect if using this)

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: This doc is outdated. We switched to the fine-grained role filtering a month ago, or probably even older.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: There are several algorithms for determining the best number of histogram bins, e.g. Freedman-Diaconis. I have not started to write the code yet so I cannot detail here.

dpordomingo: Great! So just please document what could we expect when that part is ready.

dpordomingo:
I'm not sure if the disclaimer from this https://github.com/athenianco/athenian-webapp/pull/543 should change something here, but just in case, you @vmarkovtsev might be interested in taking a look into it.

vmarkovtsev:
Thanks David, indeed that is relevant.

PR #608, opened 2020-08-05
[DEV-634] Take the logarithm of zero

vmarkovtsev:


server/athenian/api/controllers/features/github/pull_request.py
```diff
@@ -147,7 +150,11 @@ class PullRequestHistogramCalculator(PullRequestMetricCalculator):
 
     def histogram(self, scale: Scale, bins: int) -> Histogram[T]:
         """Calculate the histogram over the current distribution."""
-        return calculate_histogram(self.samples, scale, bins)
+        if scale == Scale.LOG and self.shift_log:
+            samples = [(s if s != 0 else s + 1) for s in self.samples]
```
dpordomingo, line 154: Depending on the circumstance, we could have issues with logarithms of numbers between `0` and `1`.
I this range, I'd return 1.

(but idk if this applies in this exact case)

vmarkovtsev: Once we have such metrics, I'll revisit this. This is intentional.

vmarkovtsev: OK the CI failed so I can do it now :+1:

PR #636, opened 2020-08-14
[DEV-668] Add quantiles to /metrics/prs and /histograms/prs

vmarkovtsev:


vmarkovtsev:
@se7entyse7en this also adds parallel processing of ForSet-s in metrics and histograms.

warenlg:
Awesome, very impatient to use this one. Thank you Vadim!

warenlg:
@vmarkovtsev question: Do you return the percentile value in the endpoint?

We need it to draw the vertical line below in the distribution chart

image (55) (1)

warenlg:
Actually, @navassouza you can take the max value from [0, 0.95] or the min from [0.95, 1]

vmarkovtsev:
Yes, it is better to do that on the front. I return the frequencies, that's all which is needed.

warenlg:
@vmarkovtsev I just realized that we don't have the max value of [0, 0.95], could we add it in /metric/prs?

PR #656, opened 2020-08-18
[DEV-685] Add the spec of /metrics/releases

vmarkovtsev:


vmarkovtsev:
@warenlg In order to proceed, I need to have at least a weak approval here to minimize the chances to rewrite stuff the next day.

warenlg:
Yes absolutely, it's in my ToDo of this morning

server/athenian/api/models/web/release_metric_id.py
```diff
@@ -0,0 +1,7 @@
+from athenian.api.models.web.base_model_ import Enum, Model
+
+
+class ReleaseMetricID(Model, metaclass=Enum):
+    """Linear release metric identifier."""
+
+    RELEASE_COUNT = "release-count"
```
warenlg, line 7: Is it possible to differentiate the count for the 2 release strategies: one count for the GH tags and another one for the releases into branch?

server/athenian/api/models/web/release_metric_id.py
```diff
@@ -0,0 +1,7 @@
+from athenian.api.models.web.base_model_ import Enum, Model
+
+
+class ReleaseMetricID(Model, metaclass=Enum):
```
warenlg, line 4: FYI, the next release metrics are going to be:

release-size (in terms of PRs, commits and loc)
release-frequency (avg release age)
The designs and Jira tickets associated are still in progress but in case you want to tackle all of them at once, I can adapt the priorities.

warenlg (commented):
At the end, I only have one question related to `release-count` and the release strategies. Otherwise, my comments are about the next metrics coming.

server/athenian/api/models/web/release_metrics_request.py
```diff
@@ -0,0 +1,228 @@
...
+        granularities: Optional[List[str]] = None,
+        timezone: Optional[int] = None,
+        account: Optional[int] = None,
+    ):
+        """ReleaseMetricsRequest - a model defined in OpenAPI
+
+        :param for_: The for of this ReleaseMetricsRequest.
+        :param metrics: The metrics of this ReleaseMetricsRequest.
+        :param date_from: The date_from of this ReleaseMetricsRequest.
+        :param date_to: The date_to of this ReleaseMetricsRequest.
```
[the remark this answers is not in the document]
vmarkovtsev: Note: `[date_from, date_to]` after applying the granularity, that is, if we split by 1 day intervals, only releases that happened on that day will be counted.

server/athenian/api/models/web/release_metric_id.py
vmarkovtsev: For the same repository?

vmarkovtsev: Or do you mean that the spec's response should include the applied release strategy?

server/athenian/api/models/web/release_metric_id.py
vmarkovtsev: OK, I'll add all of them to the spec but will not code yet.

server/athenian/api/models/web/release_metric_id.py
warenlg: The release strategy yes. I'm thinking that some users will want to know the number of release GH tags, without counting the other releases from the release into a branch strategy.

vmarkovtsev: ok :+1:

warenlg:
Two other points further down the roadmap:

1. Regarding the Jira filters, they are going to influence those release metrics as well. A query from the Jira filter is going to select a subset of PRs. Among them, the released PRs are inferring a subset or releases which are the releases used to compute the metrics.

2. At some point, we may also include in the product a filter by release. Ex: give me the analytics for all the PRs released in slack-ios v3.2.0

vmarkovtsev:
(2) is straightforward, but (1) will require some clever tricks to join pdb with mdb (= not easy, requires careful planning).

warenlg:
cc @eiso

warenlg (approved):
Good to go!

PR #704, opened 2020-09-02
Move Github installation views to the correct namespace

dennwc:
This is the first small step of the AppID migrations: move Github installation views to the namespace they belong to.
Until now it was wrapped by a compat view redirecting to the correct namespace. Once this is merged, I can clean those temporary views.


vmarkovtsev (changes requested):
We need to update the test data:
```
cd server/tests
xz -d test_data.sql.xz
sed -i 's/github_installation_owners/github.installation_owners/g' test_data.sql
sed -i 's/github_installation_repos_compat/github.installation_repos_compat/g' test_data.sql
xz -9 test_data.sql
```

vmarkovtsev (approved):

PR #742, opened 2020-09-14
Experiment/metrics aggregation

se7entyse7en:
This PR introduces a PoC for experimenting a new metrics aggregation mode by pre-computing them from events.

This contains two scripts:
- `athenian/api/experiments/aggregates/worker.py`, for ingestion
- `athenian/api/experiments/aggregates/benchmark.py`, for benchmarking

The ingestion script currently avoids handling some edge-cases and the metrics are probably not accurate yet. The benchmark tests a limited scenario to assess at high-level the performance of this approach.

The results from running the benchmark locally on ES data is summarized by the following table:
```
                        from                        to      label granularity      mean       std  number_of_prs
18 2020-06-01 00:00:00+00:00 2020-07-01 00:00:00+00:00    1 month         all  0.068376  0.015003           2446
16 2020-05-01 00:00:00+00:00 2020-07-01 00:00:00+00:00   2 months         all  0.114525  0.040843           4147
14 2020-04-01 00:00:00+00:00 2020-07-01 00:00:00+00:00   3 months         all  0.119342  0.003510           5468
12 2020-01-01 00:00:00+00:00 2020-07-01 00:00:00+00:00   6 months         all  0.216771  0.006140           8886
10 2019-10-01 00:00:00+00:00 2020-07-01 00:00:00+00:00   9 months         all  0.319136  0.008724          11801
8  2019-07-01 00:00:00+00:00 2020-07-01 00:00:00+00:00  12 months         all  0.390919  0.006383          14133
6  2019-01-01 00:00:00+00:00 2020-07-01 00:00:00+00:00  18 months         all  0.485881  0.014224          17377
4  2018-07-01 00:00:00+00:00 2020-07-01 00:00:00+00:00  24 months         all  0.603167  0.056475          20387
2  2017-07-01 00:00:00+00:00 2020-07-01 00:00:00+00:00  36 months         all  0.866530  0.062935          28503
0  2015-07-01 00:00:00+00:00 2020-07-01 00:00:00+00:00  60 months         all  1.309548  0.074257          47761
19 2020-06-01 00:00:00+00:00 2020-07-01 00:00:00+00:00    1 month         day  0.058992  0.005143           2446
17 2020-05-01 00:00:00+00:00 2020-07-01 00:00:00+00:00   2 months         day  0.096043  0.006009           4147
15 2020-04-01 00:00:00+00:00 2020-07-01 00:00:00+00:00   3 months         day  0.126809  0.005623           5468
13 2020-01-01 00:00:00+00:00 2020-07-01 00:00:00+00:00   6 months         day  0.218129  0.019170           8886
11 2019-10-01 00:00:00+00:00 2020-07-01 00:00:00+00:00   9 months         day  0.289554  0.011145          11801
9  2019-07-01 00:00:00+00:00 2020-07-01 00:00:00+00:00  12 months         day  0.351380  0.013456          14133
7  2019-01-01 00:00:00+00:00 2020-07-01 00:00:00+00:00  18 months         day  0.439027  0.028950          17377
5  2018-07-01 00:00:00+00:00 2020-07-01 00:00:00+00:00  24 months         day  0.885002  0.199153          20387
3  2017-07-01 00:00:00+00:00 2020-07-01 00:00:00+00:00  36 months         day  0.757354  0.053243          28503
1  2015-07-01 00:00:00+00:00 2020-07-01 00:00:00+00:00  60 months         day  1.208145  0.060296          47761
```


server/athenian/api/controllers/features/github/pull_request_filter.py
```diff
@@ -554,17 +553,25 @@ async def fetch_pull_requests(prs: Dict[str, Set[int]],
...
         pdb_misses = 0
         for pr in prs:
             node_id = pr.pr[PullRequest.node_id.key]
             if node_id not in facts:
-                facts[node_id] = facts_miner(pr)
-                pdb_misses += 1
+                try:
+                    facts[node_id] = facts_miner(pr)
+                except ImpossiblePullRequest:
+                    print("ImpossiblePullRequest")
```
vmarkovtsev, line 568: Right, this endpoint is not used in the current production, so if `ImpossiblePullRequest` is properly handled in e.g. `/filter/pull_requests`, it was ignored here.

```suggestion
                    continue
```

server/athenian/api/controllers/settings.py
```diff
@@ -91,6 +91,48 @@ def _cached_release_settings_sql(account: int, repos: Collection[str]) -> Select
         return select([ReleaseSetting]).where(and_(ReleaseSetting.account_id == account,
                                                    ReleaseSetting.repository.in_(repos)))
 
+    @staticmethod
+    @lru_cache(1024)
+    def _cached_release_settings_for_repos_sql(repos: Collection[str]) -> Select:
+        return select([ReleaseSetting]).where(ReleaseSetting.repository.in_(repos))
```
vmarkovtsev, line 97:
```suggestion
```

We have to persist the account check, the whole design is based on the security model in which each repo is always bound to some account. If you load the repository sets, for example, you should have it as the `owner_id`. This cutting the corners is OK for an experiment but ofc is not OK to include to a production codebase :wink:

server/athenian/api/controllers/settings.py
```diff
@@ -91,6 +91,48 @@ def _cached_release_settings_sql(account: int, repos: Collection[str]) -> Select
...
+
+    @classmethod
+    async def _list_release_matches_for_repos(
```
vmarkovtsev, line 100: Therefore I believe that it should be possible to refactor the experiment to leave this class intact.

server/athenian/api/experiments/aggregates/benchmark.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+async def benchmark(adb_conn_uri: str, repositories: List[str], n: int):
+    adb_conn = ParallelDatabase(adb_conn_uri)
+    await adb_conn.connect()
+
+    to = datetime(2020, 7, 1).replace(tzinfo=pytz.utc)
+
+    froms = [
+        (datetime(2020, 6, 1).replace(tzinfo=pytz.utc), "1 month"),
```
vmarkovtsev, line 28: Can you please use `timezone.utc` everywhere (`from datetime import timezone`).

server/athenian/api/experiments/aggregates/worker.py
```diff
@@ -0,0 +1,569 @@
...
+        ),
+    )
+
+    if not unfresh_prs:
+        log.info("No non-fresh PRs found")
+        return {}, []
+
+    log.info("%d non-fresh PRs aggregations found!", len(unfresh_prs))
+
+    unfresh_prs_df = pd.DataFrame(unfresh_prs)
```
vmarkovtsev, line 246: There is a handy `read_sql_query` function in `async_read_sql_query.py` to wrap this :+1:

server/athenian/api/experiments/aggregates/worker.py
```diff
@@ -0,0 +1,569 @@
...
+) -> PullRequestsCollection:
+    log.info("Retrieving PRs with outdated aggregations from %s (freshness threshold = %ds)...",
+             table_fqdn(adb, PullRequestStatus), freshness_threshold_seconds)
+    # 1. fetch non-fresh prs from aggregates
+    # SELECT repository_node_id, repository_full_name, node_id, number, last_aggregation_timestamp
+    # FROM pull_requests_status
+    # WHERE EXTRACT(EPOCH FROM (NOW() - last_aggregation_timestamp)) > :threshold
+    unfresh_prs = await adb.fetch_all(
+        select([
+            PullRequestStatus.repository_node_id,
```
vmarkovtsev, line 228: We can add `account` to `PullRequestStatus` to avoid hacking the release settings, it looks like.

server/athenian/api/models/metadata/github.py
```diff
@@ -39,6 +39,7 @@ class UserMixin:
 
 class RepositoryMixin:
     repository_full_name = Column(Text, nullable=False)
+    repository_node_id = Column(Text, nullable=False)
```
vmarkovtsev, line 42: This can potentially break all the unit tests :cry: New columns require corresponding data filled in `tests/test_data.sql.xz`.

vmarkovtsev (changes requested):
I see, so we don't directly work with the events, but rather with some pre-aggregated model.

I feel that it should be possible to gradually move from the current 3-table hydra to the single table without nasty pickle-s inside, and this PR should serve as a lighthouse of what we want to achieve.

server/athenian/api/controllers/settings.py
se7entyse7en: > This cutting the corners is OK for an experiment but ofc is not OK to include to a production codebase 😉

💯

server/athenian/api/experiments/aggregates/worker.py
se7entyse7en: Indeed, I didn't because as you said I just wanted to come up with something that roughly provides the idea and that was possible to test. But I totally agree that this has to be done. Moreover, this should also remove the need for security-check when serving metrics since repos that are not part of an account will just match no rows in the `aggregates.pull_requests_events` table, thus squeezing even more the load.

server/athenian/api/models/metadata/github.py
se7entyse7en: Ok, I'll have a look into it, thanks! 👍

se7entyse7en:
> I see, so we don't directly work with the events, but rather with some pre-aggregated model.

Yup, it's much easier to look at the code and the table schema rather than explaining it 😂 .

The pre-aggregated model would give us a LOT of advantages and also room for further performance squeezing.

The `/metrics/pull_requests` endpoint are called by the webapp (almost?) always with `author = <dev>`. We could have a separate table like `aggregates.pull_requests_events_authors` that aggregates together on an hourly basis all the events of the same author. For example, let's say we have author `xxx` that within an hour it merges a PR and releases it. We would have at least 3 rows:
- {"merging-time": 10, "merged": 1, "event-type": "merged"},
- {"closed": 1, "event-type": "closed"},
- {"release-time": 20, "released": 1, "event-type": "released"}

But by pre-aggregating by author we could have as single row:
- {"merging-time": 10, "merged": 1, "closed": 1, "release-time": 20, "released": 1}

Now, let's say that the same author does the same for multiple PRs within an hour, it's easy to see that the number of rows is reduced thus the final SQL aggregation needs to go over fewer rows.

This can be further extended to a day granularity, so each row could actually correspond to tens of events. Now here's the caveat, since we support timezones, we cannot go less granular than an hour (actually the timezone supported is by minute, but whatever). Here it comes the crazy idea! Instead of having a single table, we could have `24` of them, one for each timezone. This would reduce a lot the cardinality of the rows to be aggregated when serving the metrics.

vmarkovtsev:
> Instead of having a single table, we could have 24 of them, one for each timezone.

There are cases when the timezone is a multiple of 30min, there are also daylight saving rules for each country, and there is a boss with Arabic countries where the time offset is bound to the sunset.

But that's fine: we don't have to make complex pre-aggregations with our current scale. That's a problem for the future self :smile:

se7entyse7en:
Okay, everything shuold have been addressed. I'm gonna look into [this](https://github.com/athenianco/athenian-api/pull/742#discussion_r488069526) now.

se7entyse7en:
> There are cases when the timezone is a multiple of 30min, there are also daylight saving rules for each country,

Indeed, I remember there were also things such as +/- 45mins... 🤦 

TIL:

> and there is a boss with Arabic countries where the time offset is bound to the sunset.

WOW. 😲 

> But that's fine: we don't have to make complex pre-aggregations with our current scale. That's a problem for the future self 😄

HAHAHA yes I agree, I just wanted to share it 😂

vmarkovtsev (approved):
It is left to break through the CI.

se7entyse7en:
> It is left to break through the CI.

@vmarkovtsev there's one test that will continue to fail there's the schema check for precomputed releases. Should I add `repository_node_id` to the `PrecomputedRelease` as well? Or should I remove that column from the test? I guess the former, but I'd rather ask for your confirmation also because I'm not aware of how this would impact in terms of operations that we need to eventually perform.

se7entyse7en:
> This tests/controllers/test_invitation_controller.py on the other hand is not failing locally. 🤔

nvm it's failing locally as well 👍

vmarkovtsev:
`RepositoryMixin` is used in 8 different models. If we change it, then not only we have to change the fixture DB dump, but also to PR it to the metadata schema. Some models must be identical to pdb, that's true. If we change pdb then we have to reset pdb in both staging and prod.

It looks like too much for merging an experiment. There are other options:

1. Add the node it only to the model that is required for the experiment.
2. Query node ids from `NodeRepository`.

se7entyse7en:
> If we change it, then not only we have to change the fixture DB dump,

done already, the failing tests are fixed except for the release tests ones.

> but also to PR it to the metadata schema.

nope, that field is there already, I didn't add it.

se7entyse7en:
@vmarkovtsev since the `repository_node_id` column is already available in MD, what about if I just add it as well to `GitHubRelease`? Resetting `pdb` seems like something we don't perform that rarely, or I'm missing something? In the end, since that column is actually available in MD, having visibility of that also at API level seems sensible.

vmarkovtsev:
OK, if metadata already has it in all the views then yes, let's write a new migration that adds that column to `GitHubRelease` and `TRUNCATE`-s it. It will additionally simplify `/contributors` where I cannot query by repository ID in many places.

Now I remember, I asked Denys to add the IDs a month ago when we were optimizing SQL.

se7entyse7en:
> let's write a new migration that adds that column to GitHubRelease

Ok, on it 👍.

> It will additionally simplify /contributors where I cannot query by repository ID in many places.

Great 🚀 

> Now I remember, I asked Denys to add the IDs a month ago when we were optimizing SQL.

Ok it makes sense now why it wasn't there already 👍

se7entyse7en:
@vmarkovtsev could you please help me understand why the `test_microintegration` is failing? I don't get it. All the other tests even using the new `repository_node_id` are passing, but the migration in that test seems to fail. But if I run the migration manually everything works with no errors.

vmarkovtsev:
`test_microintegration` tries to launch the server on fake data. What's in the logs?

se7entyse7en:
That's the problem the logs do not say anything 😄 : https://travis-ci.com/github/athenianco/athenian-api/jobs/389678460#L633

They just print the migrations and then it just says that the script exited with non-zero code. So I wondered whether you bumped into something similar already for previous migrations.

vmarkovtsev:
@se7entyse7en The problem is here:
```
INFO  [alembic.runtime.migration] Running upgrade d42614e7a60a -> 5a5515e2c599, Add repository_node_id to github_releases table
Traceback (most recent call last):
  File "tests/gen_sqlite_db.py", line 53, in <module>
    exit(main())
  File "tests/gen_sqlite_db.py", line 39, in main
    migrate(name, conn_str, exec=False)
  File "/home/travis/build/[secure]/athenian-[secure]/server/athenian/[secure]/models/__init__.py", line 212, in migrate
    subprocess.run(" ".join(args[1:]), check=True, shell=True)
  File "/opt/python/3.8.1/lib/python3.8/subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '/home/travis/virtualenv/python3.8.1/bin/python3 -m athenian.[secure].sentry_wrapper alembic.config upgrade head' returned non-zero exit status 1.
```

So it fails to migrate `d42614e7a60a -> 5a5515e2c599`. I will TAL.

vmarkovtsev:
Yep, it fails locally for me with the same error message.

vmarkovtsev:
You see @dennwc, Marvin suffers the same way as I suffer when I prepare migrations to metadata hehe.

vmarkovtsev:
OK, so this happens because the migration fails, and no error message is printed, e.g.
```
vadim@vadim-XPS-13-9380:~/Projects/athenian-api/server$ python3 -m athenian.api.sentry_wrapper alembic.config upgrade head
Skipped Sentry initialization: SENTRY_KEY envvar is missing
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
vadim@vadim-XPS-13-9380:~/Projects/athenian-api/server$ echo $?
1
```

There is probably a cmdline flag to print the error in Alembic. This problem is related to the way we call Alembic - through `python3 -m` and not through `/usr/local/bin/alembic`.

vmarkovtsev:
The reason is
```
    try:
        runpy.run_module(sys.argv[0], run_name="__main__", alter_sys=True)
    except BaseException as e:
        sentry_sdk.capture_exception(e)
        return 1
```
Fixing this.

vmarkovtsev:
Fixed.
```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) Cannot add a NOT NULL column with default value NULL
[SQL: ALTER TABLE github_releases ADD COLUMN repository_node_id TEXT NOT NULL]
```

vmarkovtsev:
@se7entyse7en It prints the error message now. Do you prefer to fix the migration yourself or shall I do it?

se7entyse7en:
ahhh thanks! if you could to it that would be great (and faster 😛 )! I'll see how you'll fix it 👍

vmarkovtsev (approved):

vmarkovtsev:
Thanks a lot for the code :mechanical_arm:

se7entyse7en:
Thanks to you for the help 🙇 👍

PR #747, opened 2020-09-16
[DEV-823] Add pr-reviewed and replace pr-released with pr-done

vmarkovtsev:


server/athenian/api/controllers/features/github/pull_request_metrics.py
```diff
@@ -467,6 +467,24 @@ def _analyze(self, facts: pd.DataFrame, min_time: datetime, max_time: datetime,
...
+                 **kwargs) -> np.ndarray:
+        """Calculate the actual state update."""
+        dtype = facts["created"].dtype
+        min_time = np.array(min_time, dtype=dtype)
+        max_time = np.array(max_time, dtype=dtype)
+        review_timestamps = np.concatenate(facts["reviews"])
+        reviews_in_range = (min_time <= review_timestamps) & (review_timestamps < max_time)
+        # we cannot sum `reviews_in_range` because there can be several reviews for the same PR
+        review_offsets = np.cumsum(facts["reviews"].apply(len).values)
+        # some numpy vector magic to find the distinct number of PRs
```
vmarkovtsev, line 484: @se7entyse7en I've stopped here.

server/athenian/api/models/precomputed/models.py
```diff
@@ -40,7 +40,7 @@ class GitHubDonePullRequestFacts(Base, UpdatedMixin):
 
     pr_node_id = Column(CHAR(32), primary_key=True)
     release_match = Column(Text(), primary_key=True)
-    format_version = Column(Integer(), primary_key=True, default=6, server_default="6")
```
se7entyse7en, line 43: @vmarkovtsev you bumped these because of the new `reviews` field in the `PullRequestFacts` dataclass, right?

vmarkovtsev: Every time we change the facts, pdb has to be reset. The fundamental pitfall of precomputation :smile:

server/athenian/api/controllers/features/github/pull_request_metrics.py
```diff
@@ -464,6 +463,24 @@ def _analyze(self, facts: pd.DataFrame, min_time: datetime, max_time: datetime,
...
+
+    dtype = int
+
+    def _analyze(self, facts: np.ndarray, min_time: datetime, max_time: datetime,
+                 **kwargs) -> np.ndarray:
+        dtype = facts["created"].dtype
+        min_time = np.array(min_time, dtype=dtype)
+        max_time = np.array(max_time, dtype=dtype)
+        reviews = np.array(facts["reviews"])
+        result = np.vectorize(
```
se7entyse7en, line 478: I wasn't able to find any smart ways for using vectorization here 😞, maybe you @vmarkovtsev can. I thought about different options such as using `df.apply`, but that's basically the same as `np.vectorize` and I didn't know how to proceed with your approach of using `np.cumsum`. I thought about also to build a boolean matrix where the number of cols is given by the max number of reviews in a PR, but that would have required iterating on each row anyways since `np.vstack` do not accept arrays with different sizes by automatically expanding them and filling with empty value.

vmarkovtsev: I did a similar trick in the past, let me sketch it.

vmarkovtsev: Here it is:

```
@register_metric(PullRequestMetricID.PR_REVIEWED)
class ReviewedCalculator(SumMetricCalculator[int]):
    """Number of reviewed PRs."""

    def _analyze(self, facts: np.ndarray, min_time: datetime, max_time: datetime,
                 **kwargs) -> np.ndarray:
        """Calculate the actual state update."""
        dtype = facts["created"].dtype
        min_time = np.array(min_time, dtype=dtype)
        max_time = np.array(max_time, dtype=dtype)
        review_timestamps = np.concatenate(facts["reviews"])
        reviews_in_range = (min_time <= review_timestamps) & (review_timestamps < max_time)
        # we cannot sum `reviews_in_range` because there can be several reviews for the same PR
        review_offsets = np.zeros(len(facts) + 1, dtype=int)
        review_offsets[1:] = np.cumsum(facts["reviews"].apply(len).values)
        # np.searchsorted aliases several reviews of the same PR to the right border of a `review_offsets` interval
        # np.unique collapses duplicate indexes
        reviewed_indexes = np.unique(np.searchsorted(
            review_offsets, np.where(reviews_in_range)[0], side="right") - 1)
        result = np.full(len(facts), False)
        result[reviewed_indexes] = True
        return result
```

se7entyse7en: Awesome! This is twice faster!

se7entyse7en:
@vmarkovtsev I just realized that the webapp needs to not call `pr-released` first. In the future, we could better think about how to improve this process, but currently, the webapp queries ALL the metrics available in the openapi spec. In this case, if we deploy this the webapp will continue querying `pr-released` raising a 400. There's an easier alternative:
1. we merge this,
2. staging will be broken,
3. webapp updates the spec with the latest bumped version so that it stops querying `pr-released`,
4. we promote webapp,
5. we promote api.

We cannot promote together. \cc @dpordomingo @navassouza fyi

se7entyse7en:
Ah no, we cannot even promote webapp, because it will start asking for `pr-reviewed` 😞 , I guess we need to to specify the metrics by hand for now.

se7entyse7en:
The alternative would be to never add and remove metrics in the same PR at the same time. @vmarkovtsev what about then splitting this PR? One for adding the `pr-reviewed` and `pr-not-reviewed` and the other for renaming `pr-relased` to `pr-done`. The former is already separated into a single commit.

vmarkovtsev:
I cannot approve this PR because I am the author :man_facepalming: 

I see that there is no exact test for the new metric values. Let's hope for the best then.

PR #770, opened 2020-09-24
[DEV-923] Add grouping to /metrics/prs and /histograms/prs

vmarkovtsev:


server/athenian/api/openapi/openapi.yaml
```diff
@@ -2331,6 +2331,17 @@ components:
       properties:
         repositories:
           $ref: '#/components/schemas/RepositorySet'
+        repogroups:
```
dpordomingo, line 2334: Is this a sort of `groupBy` as proposed in the causing issue, but using array indexes instead of repeating the repo names?
```
for:[{
  repositories:[repo1,repo2,repo3,...,repoN],
  groupBy: [[repo1,repo2], [repo1], [repo2], [repo3,...,repoN]]
}]
```
If so, I like the idea.

But reviewing the specs, I just realize... did you consider this other scheme?
```
for:[{
  repositories:[repo1,repo2,repo3,...,repoN],
  groupBy: {
    repos int[][],
    developers int[][],
  }
}]
```

Let me explain my pov: Maybe `developers` grouping is not being implemented right now, but it's quite possible that it will be needed considering current plans; so this grouping scheme seems neater than polluting `ForSet` with extra properties with a different meaning, doesn't it?

vmarkovtsev: Yes, it is the grouping using array indexes so that we don't send too much data.

Unfortunately, repeating the trick with developers is much harder because they are an inherent part of the many underlying filters. That is, a week of work instead of one day. So when we have a spare week and a strong necessity, we'll add it in backward-compatible way.

PR #771, opened 2020-09-25
Add an in-memory indexer for contributors filtering

se7entyse7en:
This is what I was talking about [here](https://athenian.slack.com/archives/C011J3UTLEM/p1600984868039400).

This is how I ran it:
```
contribs_indexer = await indexer.ContributorsIndexer.create(mdb_conn)

date_from = datetime(2015, 1, 1).replace(tzinfo=timezone.utc)
date_to = datetime(2020, 3, 31).replace(tzinfo=timezone.utc)

contribs_indexer.search(date_from, date_to, op="OR")
```

with `mdb_conn` pointing to staging.

So if we wanna be completely precise since this draft does not include any repo or account filtering, actually the data are not ES only but ES + Athenian.


vmarkovtsev:
@se7entyse7en I am eager to merge this as soon as it passes the CI :wink:

se7entyse7en:
Linted and rebased.

PR #777, opened 2020-09-25
[DEV-836] Add stage-pending count metrics

se7entyse7en:


server/athenian/api/controllers/features/github/pull_request_metrics.py
```diff
@@ -654,3 +654,53 @@ def _analyze(self, facts: pd.DataFrame, min_time: datetime, max_time: datetime,
...
+
+        stage_indexes = {}
+        other = ~facts["done"].values
+        stage_indexes["release"] = np.where(merged_mask & other)[0]
+        other &= ~merged_mask
+        stage_indexes["merge"] = np.where(approved_mask & other)[0]
+        other &= ~approved_mask
+        stage_indexes["review"] = np.where(frr_mask & other)[0]
+        other &= ~frr_mask
+        stage_indexes["wip"] = np.where(other)[0]
```
vmarkovtsev, line 678: This is suboptimal. Each child class will run the same code. There are two options that I see:

1. Composition instead of inheritance. Declare this analysis as a dependency for each of the four Counter-s using `deps` (e.g., like in `WorkInProgressCounter`) and run that dependency manually in `_calc_stage_timings`.
2. Do only the required calculations in each of the four Counter-s.

server/athenian/api/controllers/features/github/pull_request_filter.py
```diff
@@ -226,32 +240,22 @@ def _calc_stage_timings(self) -> Dict[str, Sequence[int]]:
         node_id_key = PullRequest.node_id.key
```
vmarkovtsev, line 240: Let's rename `_calc_stage_timings` to `_calc_stage_details` or smth similar.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1808,6 +1808,10 @@ components:
         * `pr-closed`: number of closed PRs.
         * `pr-done`: number of released, closed unmerged, and force push dropped PRs.
         * `pr-size`: average size of PRs (added + deleted lines) which are active on the time interval after applying the quantiles.
+        * `pr-wip-pending-count`: number of PRs currently in Work In Progress stage
```
vmarkovtsev, line 1811:
```suggestion
        * `pr-wip-pending-count`: number of PRs currently in Work In Progress stage.
```
(trailing dots for all the four)

server/tests/controllers/test_metrics_controller.py
```diff
@@ -267,21 +267,26 @@ async def test_calc_metrics_prs_reposet(client, headers):
...
+    (PullRequestMetricID.PR_REVIEWED, 375),
+    (PullRequestMetricID.PR_NOT_REVIEWED, 273),
+    (PullRequestMetricID.PR_CLOSED, 588),
+    (PullRequestMetricID.PR_MERGED, 537),
+    (PullRequestMetricID.PR_REJECTED, 51),
+    (PullRequestMetricID.PR_DONE, 467),
+    (PullRequestMetricID.PR_WIP_PENDING_COUNT, 0),
+    (PullRequestMetricID.PR_REVIEW_PENDING_COUNT, 312),
+    (PullRequestMetricID.PR_MERGING_PENDING_COUNT, 52),
+    (PullRequestMetricID.PR_RELEASE_PENDING_COUNT, 6292),
```
vmarkovtsev, line 287: Much better now, thanks :+1:

vmarkovtsev (changes requested):

server/athenian/api/controllers/features/github/pull_request_filter.py
```diff
@@ -226,32 +240,22 @@ def _calc_stage_timings(self) -> Dict[str, Sequence[int]]:
...
         stage_timings = {}
-        for k, calc in self._calcs.items():
+        for k, calcs in self._calcs.items():
+            time_calc, pending_count_calc = calcs["time"], calcs["pending_count"]
+            pending_count_calc(df_facts, no_time_from, now)
+
             kwargs = {
                 "override_event_time": now - timedelta(seconds=1),  # < time_max
-                "override_event_indexes": stage_indexes[k],
+                "override_event_indexes": np.where(pending_count_calc.peek != None)[0],
```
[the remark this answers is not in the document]
se7entyse7en: Thanks!

server/athenian/api/controllers/features/github/pull_request_metrics.py
se7entyse7en: Ok, now this computation is done only once, PTAL if it fits you as well. Applying option 2 wouldn't have helped much since those calculations are done in a chain, so each counter would have had to re-do some computation anyways. One alternative would have been to make them a dependency chain but seems more complex than it should.

server/athenian/api/controllers/features/github/pull_request_metrics.py
```diff
@@ -654,3 +654,70 @@ def _analyze(self, facts: pd.DataFrame, min_time: datetime, max_time: datetime,
...
+        self._stage_indexes = stage_indexes
+
+        return np.array([], dtype=object)
```
vmarkovtsev, line 681: I am not sure if it will work, but we can try:

```python
return np.array([stage_indexes])
```

and then

```python
result[self._calcs[0].peek[0][self.stage_name]] = 1
```

server/athenian/api/controllers/features/github/pull_request_filter.py
```diff
@@ -63,11 +65,28 @@ def __init__(self,
         self._facts = facts
         self._events = events
         self._stages = stages
+        self._pending_counter_calc_dep = StagePendingDependencyCalculator(quantiles=(0, 1))
```
vmarkovtsev, line 68:
```suggestion
        self._pending_counter_dep = StagePendingDependencyCalculator(quantiles=(0, 1))
```

vmarkovtsev (changes requested):

server/athenian/api/controllers/features/github/pull_request_metrics.py
vmarkovtsev: ...we'll also have to set `dtype = object` in `StagePendingDependencyCalculator`

se7entyse7en: this is much cleaner 👍

vmarkovtsev (approved):

PR #813, opened 2020-10-08
[DEV-982] Add the pagination spec

vmarkovtsev:


se7entyse7en (approved):
LGTM! Thanks!

But please wait for @navassouza's review as well 👍

se7entyse7en:
To further gain performance, it should be better to even use a smaller batch size, but to parallelize the calls to `/filter/pull_requests`, right? Something like asking in batches of 3 chunks of 20 PRs in parallel, instead of asking 1 chunk of 60 serially.

vmarkovtsev:
Yes, except that the batch size should be around ~500 because there is a big constant in the timing.

navassouza:
So double checking my understanding of the spec:
- **batch** controls the amount of pr in the chunk
- **updated_from**, **updated_to** helps with the range, working as a cursor replacing some kind of offset in the query

se7entyse7en:
@navassouza yes, and

> updated_from, updated_to helps with the range, working as a cursor replacing some kind of offset in the query

you have those values from the first request to `/paginate/pull_requests` that returns all the timestamps to use. And as we were discussing with @vmarkovtsev the batch has to be big enough to avoid the constant overhead and we can even request multiple chunks in parallel.

Idea: since we want both to reduce the overall time to fetch everything in order to enable filtering and sorting and at the same time we also want to reduce the time to first render, we could ask for two different batch size, and for the first page we use a smaller one. The problem is @vmarkovtsev how heavy is calling twice `/paginate/pull_requests`?

vmarkovtsev:
It should not exceed 1s even for 1-year time spans.

vmarkovtsev:
ok, merging

PR #839, opened 2020-10-16
[DEV-1045] Add the spec for /metrics/jira

vmarkovtsev:


server/athenian/api/openapi/openapi.yaml
```diff
@@ -5079,3 +5155,105 @@ components:
...
+        metrics:
+          description: List of measured metrics.
+          items:
+            $ref: '#/components/schemas/JIRAMetricID'
+          type: array
+        granularities:
+          description: Splits of the specified time range `[date_from, date_to)`.
+          items:
+            $ref: '#/components/schemas/Granularity'
+          type: array
```
vmarkovtsev, line 5233: Note: there are no `quantiles` here. If you think we are going to need them, we'd rather add them sooner than later. @warenlg

server/athenian/api/openapi/openapi.yaml
```diff
@@ -4698,8 +4739,9 @@ components:
       type: object
     FoundJIRAStuff:
       additionalProperties: false
-      description: 'Response from `/filter/jira`: found JIRA epics, labels, and issue
-        types.'
+      description: |-
+        Response from `/filter/jira`: found JIRA epics, labels, issue types, priorities,
+        and mentioned users.
```
warenlg, line 4744: what are mentioned users here?
When (later) we'll add a table of Jira epics for instance, the filtering should work similarly to how we do it for releases: an epic is an atomic element, we don't break it.

If we select dev X in the filter, epic Y is returned if:
* dev X is assigned to epic Y
* or if dev X is assigned to any tickets (or sub tickets) of epic Y
* or if dev X is the author of any PR linked to that epic (or its children)

server/athenian/api/openapi/openapi.yaml
```diff
@@ -5079,3 +5155,105 @@ components:
...
+          description: Selected issue reporter users.
+          items:
+            $ref: '#/components/schemas/DeveloperID'
+          type: array
+        commenters:
+          description: Selected issue commenter users.
+          items:
+            $ref: '#/components/schemas/DeveloperID'
+          type: array
+        stakeholders:
```
warenlg, line 5219: commenters and stakeholders are not in the specs for the MVP, but if they don't harm and you prefer to add them now, I'm fine as well.

server/athenian/api/openapi/openapi.yaml
warenlg: We don't need quantiles for bug counts of course, but we do need them for MTTRs which are the next Jira metric to add. So I'd say yes, let's add quantiles now.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -5079,3 +5155,105 @@ components:
           - $ref: '#/components/schemas/DeveloperSet'
           - description: Release publishers.
       type: object
+    JIRAMetricID:
+      description: |-
+        Currently supported JIRA activity metrics.
+        * `jira-bug-open` - number of bug-like issues active by the end of the time interval.
```
warenlg, line 5161: active meaning non-resolved yes.

Also, we should be able to get the timeline of the open bugs during the time period selected. Not sure if that would come with the daily granularity or not?

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: I think "mentioned" should mean any possible relation of a user to the returned issues, both regular and epics.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: One more, one less does not complicate the work here. It mostly depends on the metadata schema, I'd say.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: Yes, granularity exists especially for that.

vmarkovtsev:
OK, @warenlg I need your approval to play nicely with our own product's metrics.

warenlg (approved):

PR #868, opened 2020-10-27
[DEV-1047] [DEV-1077] Return the mentioned JIRA priorities

vmarkovtsev:


server/athenian/api/openapi/openapi.yaml
```diff
@@ -5244,6 +5244,7 @@ components:
         - jira-bug-raised
         granularities:
         - all
+        exclude_inactive: true
```
warenlg, line 5247: I think this is also missing in other endpoints examples like filter/pull_requests if I recall correctly (I can not check as swagger is not available for staging)

vmarkovtsev: ok, I'll check :+1:

vmarkovtsev: You can check in prod btw.

PR #946, opened 2020-11-23
[DEV-1206] Add the Team's parent to the spec

vmarkovtsev:
This specification change allows:
- Create a new team with an explicit parent.
- Update a team to set a new parent.
- List the teams with their hierarchy via the new `parent` field.


vmarkovtsev:
cc @warenlg

PR #1010, opened 2020-12-17
Handle current transaction being `None`

se7entyse7en:

### DIFF
--- a/server/athenian/api/defer.py
+++ b/server/athenian/api/defer.py
@@ -9,7 +9,6 @@ from sentry_sdk.tracing import Transaction
 from athenian.api import metadata
 from athenian.api.typing_utils import wraps
 
-
 _defer_launch_event = None
 _defer_sync = None
 _defer_counter = None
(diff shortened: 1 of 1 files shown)


se7entyse7en:
@vmarkovtsev I don't know if this is the most proper way to handle this, but I needed to do this locally.

server/athenian/api/defer.py
```diff
@@ -81,17 +80,23 @@ def launch():
             transaction = Transaction(name="defer", sampled=False, hub=Hub.current)
         else:
             parent = Hub.current.scope.transaction
-            transaction = Transaction(name="defer " + parent.name,
-                                      sampled=parent.sampled,
-                                      op=parent.op,
-                                      description=parent.description,
-                                      hub=Hub.current)
+            if parent is None:
```
vmarkovtsev, line 83: Can you please change `if Hub.current.scope.span is None:` to `if (parent := Hub.current.scope.transaction) is None:` because it seems that the current check is wrong (IDK how because a span cannot live without a parent transaction).

vmarkovtsev (changes requested):

server/athenian/api/defer.py
se7entyse7en: done

vmarkovtsev (approved):

PR #1012, opened 2020-12-17
[DEV-1371] Make using the API with real data as a service easier

se7entyse7en:
Partially addresses [DEV-1371](https://athenianco.atlassian.net/browse/DEV-1371).

This makes using the API as a service with real data easier. Since staging database for `metadata` is GBs of data and dumping it will take a lot of time, for now, I just made the API running in docker-compose access staging directly. For `state` and `precomputed` dumps are used. Moreover, `state` and `precomputed` are also written by the API making them non-safe to be used directly from staging compared to `metadata` which is RO by the API.

Just to be sure, this morning I also checked the Network Egress spend in the last month (since I'm locally using staging data the same way, just without docker-compose) and it was around ~5$, so it's not a big deal for now.

This is not for sure the optimal solution but is currently the fastest way to make collaborators (@znegrin) productive asap with real data. We can then discuss later on a better solution.

As stated in the README, the main requirements are `docker-compose` and being logged in locally with Google Cloud SDK with an Athenian email account with the proper permissions to access Cloud SQL.

To further ease the development, on the [webapp](https://github.com/athenianco/athenian-webapp) I'll also add a Make target that downloads both the `docker-compose.yml` and the `load_pg_data.py` file, and runs the scripts as described in the README. This will also remove the necessity to locally clone the whole API repo.


load_pg_data.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
vmarkovtsev, line 1: Would you kindly move this file somewhere inside, for example, `server/tests` or similar nested.

vmarkovtsev (changes requested):

load_pg_data.sh
se7entyse7en: Done 👍

vmarkovtsev (approved):

PR #1171, opened 2021-02-19
[DEV-1768] Support release notifications in the API

vmarkovtsev:


server/athenian/api/models/persistentdata/versions/860cdc895ef3_add_release_notifications.py
```diff
@@ -0,0 +1,52 @@
...
+    if op.get_bind().dialect.name == "postgresql":
+        op.execute("CREATE SCHEMA athenian;")
+        schema_arg = {"schema": "athenian"}
+    else:
+        name = "athenian." + name
+        schema_arg = {}
+    op.create_table(
+        name,
+        sa.Column("account_id", sa.Integer(), primary_key=True),
+        sa.Column("repository_id", sa.Text(), primary_key=True),
```
dennwc, line 30: Can you clarify what this is? Is it NodeID? Full name? Better to specify explicitly in the DB.
I know there's description in the spec, but we also need the DB to be self-documented, ideally.

server/athenian/api/models/persistentdata/versions/860cdc895ef3_add_release_notifications.py
```diff
@@ -0,0 +1,52 @@
...
+        sa.Column("commit_hash_prefix", sa.Text(), primary_key=True),
+        sa.Column("resolved_commit_hash", sa.Text()),
+        sa.Column("resolved_commit_node_id", sa.Text()),
+        sa.Column("name", sa.Text()),
+        sa.Column("author", sa.Text()),
+        sa.Column("url", sa.Text()),
```
dennwc, line 36: Is it for UI?

server/athenian/api/openapi/openapi.yaml
```diff
@@ -6830,3 +6882,92 @@ components:
...
+      example:
+      - author: Vadim Markovtsev
+        commit: c3f4d3a
+        name: 0.8.66
+        repository: github.com/athenianco/athenian-api
+        published_at: 2021-02-23T15:17:07.000+01:00
+        url: https://github.com/athenianco/athenian-api/releases/tag/v0.8.66
+      properties:
+        account:
+          description: Account ID.
```
dennwc, line 6897: It must be resolved by other means, e.g. Auth. We shouldn't allow users to put arbitrary stuff here.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -6830,3 +6882,92 @@ components:
...
+    NotifiedReleaseStatuses:
+      description: Status of the accepted release notifications.
+      items:
+        $ref: '#/components/schemas/NotifiedReleaseStatus'
+      type: array
+    NotifiedReleaseStatus:
+      additionalProperties: false
+      description: Status of the accepted release notification.
+      properties:
+        repository:
```
dennwc, line 6954: Why do we need to echo it back?

server/athenian/api/openapi/openapi.yaml
```diff
@@ -6830,3 +6882,92 @@ components:
...
+          $ref: '#/components/schemas/RepositoryID'
+        commit:
+          $ref: '#/components/schemas/CommitHash'
+        status:
+          description: |-
+            Status of the release notification.
+            * `applied` the release commit was found and the release was successfully registered.
```
dennwc, line 6961: I think, we shouldn't explicitly tell if we mapped something or not. We should think of this endpoint as async (as if it always returns `pending`).

dennwc (changes requested):

server/athenian/api/openapi/openapi.yaml
dennwc: It might actually be a good idea to return it in the response, btw.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -1303,6 +1303,56 @@ paths:
 # {% endif %}
       x-openapi-router-controller: athenian.api.controllers.metrics_controller
       x-codegen-request-body-name: body
+  /notify/release:
```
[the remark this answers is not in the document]
vmarkovtsev: I see what you mean. How about this: `/events/releases`.

server/athenian/api/models/persistentdata/versions/860cdc895ef3_add_release_notifications.py
vmarkovtsev: So `repository_node_id`?

server/athenian/api/models/persistentdata/versions/860cdc895ef3_add_release_notifications.py
vmarkovtsev: Yes. It is used to link to the releases in the tables in the frontend.

server/athenian/api/openapi/openapi.yaml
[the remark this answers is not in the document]
dennwc: Yep, sounds good!

server/athenian/api/models/persistentdata/versions/860cdc895ef3_add_release_notifications.py
dennwc: I don't think we can resolve all of repo full names to NodeIDs when the request arrives.
Lets do the same as you did with commits: have a `repository_full_name` and `resolved_repo_id` or something like this.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: I have to explain how this works.

We allow the same user to be a member of several Athenian accounts at the same time, that's a product-enforced feature. So almost all of our queries require the client to specify their effective account number. The pair `<Account ID> - <User ID>` is checked by searching in `user_accounts` table in the state DB.

Depending on the account number, we put the registered release event to the corresponding account.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: We identify a release by a pair `(repository, commit)`, because there can be the same commit hashes in different repositories in the same account, e.g. hard forks.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: Fair enough, this is our implementation detail.

server/athenian/api/openapi/openapi.yaml
dennwc: This is exactly why I think we should limit this to access tokens and avoid use auth for now. It should be bound to a specific account already, so we don't have to deal with authenticated users that belong to more than one org (read: having all state bound already).

This is how GH does it as well: you could have a "deploy keys" for a specific repo.

Imagine that we send a Github install link, start fetching and to speed-up the onboarding we could instantly generate a token that is bound to the account even though it has no `state` and no users. I think it's pretty powerful.

server/athenian/api/openapi/openapi.yaml
dennwc: OK. But why send them back again, if user already specified those in the request?

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: Well, let's return nothing then. An empty response signals a successful operation, like in some other places.

dennwc: As mentioned in the other comment, we could return account ID. But I don't think it's useful, so yeah, we can return nothing, just 200.

server/athenian/api/models/persistentdata/versions/860cdc895ef3_add_release_notifications.py
vmarkovtsev: You are right.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: OK, so you suggest including the account ID inside the APIKey. Fair enough.

dennwc: Yeah, maybe encode it there or just have a table with `(account, token)` (or whatever works from security PoV).
Also, since releases are bound to the source control, I think it might be useful to bind to `github.accounts` directly. Will require less JOINs to find commits, etc.

vmarkovtsev: We already have this table, yes. Regarding the accounts, I'd rather not hardcode them directly because APIKeys are used in other endpoints that would benefit from regular accounts.

dennwc: Agree, for now it doesn't really matter, I think. Later if we decide to optimize it we could query both tables, for example.

server/athenian/api/models/persistentdata/versions/860cdc895ef3_add_release_notifications.py
vmarkovtsev: I've just realized that we have to validate the access to this repository according to the release settings. This automatically means that the repository already exists everywhere and we know the node ID. Let's leave as-is then.

vmarkovtsev: There is also a use-case when the user sends an invalid repository name (malformed, with a typo, etc.), and if we don't resolve it straight away, we accept garbage.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: I see that I thought through APIKeys well last year: the APIKey is an encrypted DB token ID that leads to all the info: user, account, enabled/disabled, permissions, etc.

server/athenian/api/models/persistentdata/versions/860cdc895ef3_add_release_notifications.py
dennwc: > I've just realized that we have to validate the access to this repository according to the release settings.

Can you elaborate please?

> There is also a use-case when the user sends an invalid repository name (malformed, with a typo, etc.)

We can do basic validation (e.g. should look like a URL part, shouldn't be long, should start with `github.com`).

Strict validation is good, but think about different kinds of race condition that may happen on repo rename. MD fetchers may be blocked for _hours_ because of the rate limit kicking in. Thus, it's better to assume no validation or optional validation and then see how we can handle it for all the edge cases.

vmarkovtsev: If the release settings of the repo are not set to "event", we must reject the notification. This is a product feature to enforce the release settings for the account.

Renaming repositories is indeed a good example when we risk rejecting a valid release. We can document this as a product limitation and see if the clients frequently have problems with them, then re-iterate. I honestly think that this will be a rare enough problem to not sacrifice the release settings validation.

dennwc: But it might be still useful to have notifications flowing, while user plays with release settings, right?

For example, they already work with Athenian and strategy is set to `tag`. They are thinking to migrate to our Releases API, but it's not instant. They first setup everything, events start flowing, but they still want to keep old strategy or may switch back to validate something.

eiso: Some thoughts from product:

If someone is sending correct events to mark a PR as released, can't we just ignore the settings? Why? Imagine a mono repo where someone is doing a manual release but usually PRs are released via merge to production branch. This is pretty common. 

I also prefer this strategy for large accounts who have 100s of repos, and will ask the repo owners to update their CD to include an API call but don't want to follow up with checking which repo has and which one hasn't. The user updating the deployment script is often not the Athenian user. 

Regarding strict validating, let's validate on repository and SHA, this way we can give people useful feedback if they mess either up.

vmarkovtsev: To summarize:
- Do not check if some repository has a mismatching release setting. We can return a warning in the reponse, but let's leave it for the future.
- We must resolve repository node IDs and check the security access rules like we do in the other endpoints. The product is fine with some edge cases like repository renames.

eiso: Agree fully on your second point.

Regarding your first point, it is not a mismatching release setting, you have: Automatic, Tag, Branch, API, **but you can still use the API when you are on Automatic, Tag or Branch** because of manual individual releases or complex release processes.

vmarkovtsev: Merging the PR then :+1:

se7entyse7en: Letting notifications flowing disregarding the release strategy is also more consistent with the other strategies. To me the equivalent sounds like considering an error when we receive a `tag creation` event on MD while the repo is configured with "branch" release strategy.

This also makes sense according to how we have split the data across the different databases: we have `persistentdata` and `metadata` dealing with "raw" data and the concept of release is "materialized" by the user settings defined in the `state` db.

PR #1228, opened 2021-03-08
Setup restart policy in docker compose for memcached and cloud sql proxy

se7entyse7en:
For some unknown reasons, in pre-staging environment, those containers are exiting with code 0 without anyone stopping them. This should at least make them restart in case it happens.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -33,6 +33,7 @@ services:
 
   memcached:
     image: launcher.gcr.io/google/memcached1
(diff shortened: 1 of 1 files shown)


vmarkovtsev (approved):

vmarkovtsev:
Are there any logs?

se7entyse7en:
Unfortunately not. No logs for both containers, the only thing that I noticed is both exiting with 0 at the same time. 🤷‍♂️

PR #1318, opened 2021-04-13
Endpoints controller feature flags

se7entyse7en:
Closes [DEV-2124](https://athenianco.atlassian.net/browse/DEV-2124).

This PR adds a way to add different implementations for metrics endpoints controllers using feature flags. Each implementation variation is by default loaded from the `athenian.api.experiments` module. In order to enable a variation, we need the followings:
- enable a feature flag named `github_features_entries_<name>`, where `<name>` corresponds to the variation package in `atenian.api.experiments.<name>`,
- enable that feature for the accounts we're interested in.

The feature flag doesn't require any parameter. An optional parameter is `god_only` that is a boolean: when `True`, the feature flag is enabled only in `god` mode (and ofc when enabled).

The variations functions have the same names as the default ones as defined in `METRIC_ENTRIES`. In case the module corresponding to the feature flag is not found or in the case that there's no implementation for a specific calculator, no error is raised in any case, but the default implementation is used as a fallback. In these cases, a log with `WARNING` level is printed.


server/athenian/api/controllers/features/entries.py
```diff
@@ -545,6 +550,122 @@ async def calc_developer_metrics_github(devs: Sequence[Collection[str]],
     return result, topics_seq
 
 
+async def get_calculator_for_user(
+    service: str,
```
vmarkovtsev, line 554: Marvin, the ideology of the code inside "features" and "miners" is that it is completely separate from the state. That is, if you look everywhere around, I don't accept `sdb`. Is it possible to re-architect this to choose which calculator to use _outside_ of `entries.py` - for example, in a controller, in a new module, or in `Settings` class (`settings.py`) :pray:  It will help to retain the (kind of) clean architecture separation of responsibilities.

vmarkovtsev (changes requested):

server/athenian/api/controllers/features/entries.py
se7entyse7en: Sure! It makes sense, thanks for pointing it out 👍

se7entyse7en: Done.

One thing that I was considering is that having all these funcs in `entries.py` as `classmethod`s of a class would make the overriding of the logic much easier. WDYT?

Otherwise how for example would you override the logic of the `PullRequestMiner`? Among the things that I considered:
- make these functions accept a `PullRequestMiner` parameter so that the experiment will just call the original function and will pass a `PullRequestMinerWhatever`, but it doesn't seem elegant to change the signature just for this,
- make everything `classmethod`s and the `get_selector` function instead of returning the calculator function will return the class variation that will have the responsibility also of calling the default implementation (this will be granted if the variation class inherits from the original class). This sounds much cleaner to me.
- hack it with monkey-patching or even with `unittest.mock`, but I really don't wanna go into this direction. 😄

vmarkovtsev: Yes, having a class is so much better. In fact, let's have those functions as regular methods instead of `classmethod`, and accept the common arguments like `pdb`, `mdb`, `rdb`, etc. in `__init__`. Caveat: you'll have some fun time updating the unit tests.

se7entyse7en: > classmethod, and accept the common arguments like pdb, mdb, rdb, etc. in __init__. 

I like it, deal! 👍 

> Caveat: you'll have some fun time updating the unit tests.

😂

vmarkovtsev: OK, please ping when ready :+1:

se7entyse7en:
This is pretty much done, I just need to investigate some non-trival errors happening in the `@cached` decorator. I'll investigate tomorrow morning, but should likely caused by the usage of `@cached` on methods rather than on function, so the signature and arguments seem to mess up when binding because of the `self`.

se7entyse7en:
@vmarkovtsev ok I understood the issue. The problem is that when the `@cached` decorator is applied to a method, the signature check through `signature.bind` fails when `reset_cache` is called. The reason is the following:
- the signature has `self`,
- when the decorated method is called, the `signature.bind` is fine because `*args` contains `self` since it's passed automatically given that it's a method,
- when `reset_cache` function attached to the method is called, the `signature.bind` raises an error because `*args` now doesn't contain `self`.

Do you have any idea on how would be the best to solve this? Among things that I thought:
- implement the `@cached` as class-based which could solve I guess but will become for sure a bit more complex (https://stackoverflow.com/questions/1288498/using-the-same-decorator-with-arguments-with-functions-and-methods),
- avoid attaching the `reset_cache` to the func/method,
- find a way to propagate the object owner to the `reset_cache` without making the decorator class-based.

Here's the minimal code for reproducing what I described:
```
In [47]: def play(param):
    ...:
    ...:     def wrapper(func):
    ...:         signature = inspect.signature(func)
    ...:
    ...:         def wrapped(*args, **kwargs):
    ...:             args_dict = signature.bind(*args, **kwargs).arguments
    ...:             return func(**args_dict)
    ...:
    ...:         def sidecar(*args, **kwargs):
    ...:             args_dict = signature.bind(*args, **kwargs).arguments
    ...:             return 0
    ...:
    ...:         wrapped.sidecar = sidecar
    ...:         return wrapped
    ...:
    ...:     return wrapper
    ...:
    ...:
    ...: @play("param")
    ...: def test(a, b, c):
    ...:     return a, b, c
    ...:
    ...:
    ...: class A:
    ...:     @play("param")
    ...:     def test(self, a, b, c):
    ...:         return a, b, c
    ...:

In [48]: test(1,2,3)
Out[48]: (1, 2, 3)

In [49]: a = A()

In [50]: a.test(1,2,3)
Out[50]: (1, 2, 3)

In [51]: a.test.sidecar(1,2,3)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-51-cb86eb13e2c6> in <module>
----> 1 a.test.sidecar(1,2,3)

<ipython-input-47-63332b0c0733> in sidecar(*args, **kwargs)
      9
     10         def sidecar(*args, **kwargs):
---> 11             args_dict = signature.bind(*args, **kwargs).arguments
     12             return 0
     13

~/miniconda3/envs/athenian-api/lib/python3.8/inspect.py in bind(self, *args, **kwargs)
   3023         if the passed arguments can not be bound.
   3024         """
-> 3025         return self._bind(args, kwargs)
   3026
   3027     def bind_partial(self, /, *args, **kwargs):

~/miniconda3/envs/athenian-api/lib/python3.8/inspect.py in _bind(self, args, kwargs, partial)
   2938                             msg = 'missing a required argument: {arg!r}'
   2939                             msg = msg.format(arg=param.name)
-> 2940                             raise TypeError(msg) from None
   2941             else:
   2942                 # We have a positional argument to process

TypeError: missing a required argument: 'c'
```

se7entyse7en:
I tried even making the decorator class-based, but I didn't find any way from the decorated function when it is a method to retrieve the _owner_ instance of the method. The reason is that when passed to the decorator, it is actually an instance of `function` and not an instance of `bound_method`.

se7entyse7en:
The easiest, but super ugly would be to call:
```
a.test.sidecar(a.test,1,2,3)
```

vmarkovtsev:
I see the problem, let me try to code a solution here. I am finally over with the bug fixes :tada:

se7entyse7en:
> I see the problem, let me try to code a solution here. I am finally over with the bug fixes 🎉

Awesome! Thanks for the help! 🙇

vmarkovtsev:
https://github.com/athenianco/athenian-api/pull/1328

se7entyse7en:
Thanks a lot! I'll rebase then after that one gets merged 👍

server/athenian/api/cache.py
```diff
@@ -215,28 +215,38 @@ def cache_key(*args, **kwargs) -> bytes:
...
+            # This is needed in case the `@cached` decorator is not the outermost one
+            func = getattr(func, "__wrapped__", None)
+            if not func:
+                break
 
-                    @functools.wraps(func.cache_key)
-                    def cache_key(*args, **kwargs):
-                        return func.cache_key(self, *args, **kwargs)
+        if not func:
+            continue
```
se7entyse7en, line 227: @vmarkovtsev PTAL at this. Basically, it's because of the usage of `@cached` as the non-outermost decorator as tested [here](https://github.com/athenianco/athenian-api/pull/1318/files#diff-13caa706ed0a55f3fb81caf5590fce9fc87f7c9e3f66a41ebd72b717872c8f7eR116).

server/athenian/api/cache.py
```diff
@@ -215,28 +215,38 @@ def cache_key(*args, **kwargs) -> bytes:
...
-                        return await func(self, *args, **kwargs)
+        outer_func = func
 
-                    async def reset_cache(*args, **kwargs):
-                        return await func.reset_cache(self, *args, **kwargs)
+        while not getattr(func, "__cached__", False) or getattr(func, "__cached_method__", False):
+            # This is needed in case the `@cached` decorator is not the outermost one
+            func = getattr(func, "__wrapped__", None)
+            if not func:
+                break
```
vmarkovtsev, line 224: `:=` is awesome
```suggestion
            if not (func := getattr(func, "__wrapped__", None)):
                break
```

server/athenian/api/controllers/events_controller.py
```diff
@@ -207,8 +207,8 @@ async def login_loader() -> str:
             JIRAFilter.empty(), settings, prefixer, model.account, meta_ids, mdb, pdb, rdb, None,
             force_fresh=True)
         await wait_deferred()
-        await calc_pull_request_facts_github(
+        await MetricEntriesCalculator(mdb, pdb, rdb, None).calc_pull_request_facts_github(
```
vmarkovtsev, line 210: Can you please also add `account` and `meta_ids` inside the constructor :pray:

server/athenian/api/controllers/histograms_controller.py
```diff
@@ -41,12 +41,15 @@ async def calculate_for_set_histograms(service, repos, withgroups, labels, jira,
                 bins=h.bins,
                 ticks=tuple(h.ticks) if h.ticks is not None else None,
             )].append(h.metric)
+        calculator = await get_calculator_for_user(
+            service, filt.account, (await request.user()).id,
+            request.sdb, request.mdb, request.pdb, request.rdb, request.cache,
+        )
```
vmarkovtsev, line 47: The same effect:
```suggestion
        calculator = await get_calculator_for_user(
            service, filt.account, request.uid,
            request.sdb, request.mdb, request.pdb, request.rdb, request.cache,
        )
```

server/athenian/api/controllers/selector.py
```diff
@@ -0,0 +1,80 @@
...
+                Feature.enabled,
+            ),
+        ),
+    )
+
+    if not all_metrics_variations_features:
+        return _get_calculator()
+
+    all_metrics_variations_features = {
+        row[0]: {"name": row[1][len(feature_name_prefix):], "params": row[2] or {}}
```
vmarkovtsev, line 47: I found that taking values by index quickly goes out of sync with the selected columns list.
```suggestion
        row[Feature.id.key]: {"name": row[Feature.name.key][len(feature_name_prefix):], "params": row[Feature.default_parameters.key] or {}}
```

server/athenian/api/controllers/selector.py
```diff
@@ -0,0 +1,80 @@
...
+        select([AccountFeature.feature_id, AccountFeature.parameters]).where(
+            and_(
+                AccountFeature.account_id == account_id,
+                AccountFeature.feature_id.in_(all_metrics_variations_features),
+                AccountFeature.enabled,
+            ),
+        ),
+    )
+
+    if not metrics_variation_feature:
```
vmarkovtsev, line 61: This can break with crazy DB backends. Better `if metrics_variation_feature is None`

server/athenian/api/controllers/selector.py
```diff
@@ -0,0 +1,80 @@
...
+
+    selected_metrics_variation = all_metrics_variations_features[
+        metrics_variation_feature[0]
+    ]
+    metrics_variation_params = {
+        **selected_metrics_variation["params"],
+        **(metrics_variation_feature[1] or {}),
+    }
+
+    is_god = await sdb.fetch_one(select([God.user_id]).where(God.user_id == user_id))
```
vmarkovtsev, line 72: We have `request.god_id` defined in god mode.

If that does not fit, we should batch this and one of the previous queries together in `gather()`. Otherwise, we've got 3 queries to sdb standing on the critical path, that's too much.

Is it possible to merge the previous two queries together somehow? A join maybe?

server/athenian/api/controllers/selector.py
```diff
@@ -0,0 +1,80 @@
+from typing import Callable, Optional
```
vmarkovtsev, line 1: Let's name it `feature_selector` or `calculator_selector`. My first association was UNIX's `select()` :smile:

server/athenian/api/controllers/features/entries.py
```diff
@@ -288,273 +551,46 @@ def _compose_cache_key_participants(participants: List[PRParticipants]) -> str:
...
-        "prs_linear": calc_pull_request_metrics_line_github,
-        "prs_histogram": calc_pull_request_histograms_github,
-        "code": calc_code_metrics_github,
-        "developers": calc_developer_metrics_github,
-        "releases_linear": calc_release_metrics_line_github,
-    },
-}
+        mod = importlib.import_module(f"{base_module}.{variation}")
+    except ModuleNotFoundError:
+        if raise_err:
```
vmarkovtsev, line 572: We should do only one thing: either raise or log an error. I don't want special execution modes, it is very easy to confuse them and get drown in combinatorial complexity.

server/athenian/api/controllers/features/entries.py
```diff
@@ -288,273 +551,46 @@ def _compose_cache_key_participants(participants: List[PRParticipants]) -> str:
...
+            raise
+
+        log.warning(
```
vmarkovtsev, line 575: This is clearly an error.
```suggestion
        log.error(
```

server/athenian/api/controllers/features/entries.py
```diff
@@ -288,273 +551,46 @@ def _compose_cache_key_participants(participants: List[PRParticipants]) -> str:
...
+            "Invalid variation '%s' calculator, using default implementation",
+            variation,
+        )
+        return MetricEntriesCalculator(mdb, pdb, rdb, cache)
+    else:
+        try:
+            cls = mod.MetricEntriesCalculator
+        except AttributeError:
+            if raise_err:
```
vmarkovtsev, line 584: Same here, either raise or log an error.

server/athenian/api/controllers/histograms_controller.py
```diff
@@ -41,12 +41,15 @@ async def calculate_for_set_histograms(service, repos, withgroups, labels, jira,
                 bins=h.bins,
                 ticks=tuple(h.ticks) if h.ticks is not None else None,
             )].append(h.metric)
+        calculator = await get_calculator_for_user(
```
vmarkovtsev, line 44: Let's consider `gather()`-ing this with the previous coroutines, e.g. fetching the release settings is independent of finding the calculator and can run in parallel.

vmarkovtsev (changes requested):

server/athenian/api/cache.py
se7entyse7en: Still need to get used to this idiom, I guess this means also that it's been a long time since the last time I seriously programmed in go 😂

server/athenian/api/controllers/selector.py
se7entyse7en: haha 👍

server/athenian/api/controllers/features/entries.py
se7entyse7en: The reason why I added this double-execution way it's to protect from the human error of adding feature flags, but it's true that we'd check the API immediately after setting it.

server/athenian/api/controllers/selector.py
se7entyse7en: also much more readable 👍

server/athenian/api/controllers/selector.py
se7entyse7en: Nice catch! Thanks!

server/athenian/api/controllers/selector.py
se7entyse7en: `request.god_id` will work, thanks for pointing it out, and I agree with the perf bottleneck here otherwise 👍

server/athenian/api/controllers/features/entries.py
se7entyse7en: Solved by always fall backing to original implementation and notifying to Sentry

se7entyse7en:
@vmarkovtsev everything should have been addressed and tests are finally passing

server/athenian/api/controllers/features/entries.py
```diff
@@ -66,203 +68,452 @@ def _postprocess_cached_facts(result: Tuple[Dict[str, List[PullRequestFacts]], b
...
-            precomputed_facts, ambiguous, time_from, time_to, repositories,
-            participants, labels, jira, exclude_inactive, branches,
-            default_branches, release_settings, account, meta_ids, mdb, pdb, rdb, cache)
+@cached_methods
+class MetricEntriesCalculator:
+    """Calculator for different metrics."""
+
+    def __init__(self, account: int, meta_ids: Tuple[int, ...],
+                 mdb: Database, pdb: Database, rdb: Database,
+                 cache: Optional[aiomcache.Client]) -> "MetricEntriesCalculator":
```
vmarkovtsev, line 77: `__init__` does not return anything :wink:

server/athenian/api/controllers/features/entries.py
```diff
@@ -288,273 +539,45 @@ def _compose_cache_key_participants(participants: List[PRParticipants]) -> str:
...
-        "releases_linear": calc_release_metrics_line_github,
-    },
-}
+        mod = importlib.import_module(f"{base_module}.{variation}")
+    except ModuleNotFoundError as e:
+        sentry_sdk.capture_exception(e)
+        log.error(
+            "Invalid variation '%s' calculator, using default implementation",
+            variation,
+        )
```
vmarkovtsev, line 569: `log.error` automatically logs to sentry. We should write
```suggestion
        log.exception(
            "Invalid variation '%s' calculator, using default implementation",
            variation,
        )
```
and it will go to Sentry. BUT in this case we don't really care about the exception really, so `log.error` should be enough.

server/athenian/api/controllers/histograms_controller.py
```diff
@@ -28,8 +28,15 @@ async def calc_histogram_prs(request: AthenianWebRequest, body: dict) -> web.Res
     meta_ids = await get_metadata_account_ids(filt.account, request.sdb, request.cache)
     filters, repos = await compile_repos_and_devs_prs(filt.for_, request, filt.account, meta_ids)
     time_from, time_to = filt.resolve_time_from_and_to()
-    release_settings = \
-        await Settings.from_request(request, filt.account).list_release_matches(repos)
+    services = set(s for s, _ in filters)
+    res = await gather(
+        Settings.from_request(request, filt.account).list_release_matches(repos),
```
vmarkovtsev, line 33: We can chain `compile_repos_and_devs_prs(filt.for_, request, filt.account, meta_ids)` and `Settings.list_release_matches`.
```suggestion
    async def load_repos_and_release_settings():
        filters, repos = await compile_repos_and_devs_prs(filt.for_, request, filt.account, meta_ids)
        release_settings = await Settings.from_request(request, filt.account).list_release_matches(repos)
        return filters, repos, release_settings
    (filters, repos, release_settings), calculators = await gather(load_repos_and_release_settings(), *[get_calculator_for_user()])
```

server/athenian/api/controllers/metrics_controller.py
```diff
@@ -432,10 +442,13 @@ async def calculate_for_set_metrics(service, repos, for_sets):
                               ("pr_author", ReleaseParticipationKind.PR_AUTHOR),
                               ("commit_author", ReleaseParticipationKind.COMMIT_AUTHOR))
         } for with_ in (filt.with_ or [])]
-        release_metric_values, release_matches = await METRIC_ENTRIES[service]["releases_linear"](
+        calculator = await get_calculator_for_user(
```
vmarkovtsev, line 445: We can move this inside `tasks`.

server/athenian/api/controllers/metrics_controller.py
```diff
@@ -85,11 +85,14 @@ async def calc_metrics_pr_linear(request: AthenianWebRequest, body: dict) -> web
 
     @sentry_span
     async def calculate_for_set_metrics(service, repos, withgroups, labels, jira, for_set):
-        metric_values = await METRIC_ENTRIES[service]["prs_linear"](
+        calculator = await get_calculator_for_user(
```
vmarkovtsev, line 88: We can join this with `release_settings`.

vmarkovtsev (changes requested):
A few final changes and we should be good to go!

server/athenian/api/controllers/histograms_controller.py
se7entyse7en: The problem is that we need `filters` for `get_calculator_for_user` since we need to build a calculator for each service.

vmarkovtsev: I see, NVM then.

se7entyse7en:
Rebased and latest issues addressed in last two commits: https://github.com/athenianco/athenian-api/pull/1318/commits/04b6f134657fdc2591bb465923a2753a071e0270, https://github.com/athenianco/athenian-api/pull/1318/commits/a695f2aeeace913f7a177be26942cfd18f17dbaf

vmarkovtsev (approved):
I will torture you no more :D

vmarkovtsev:
Merging this after #1341 - it will not deploy otherwise.

vmarkovtsev:
Not merging until Auth0 fixes their shit.

PR #1320, opened 2021-04-14
Preloading setup

se7entyse7en:
Closes [DEV-2146](https://athenianco.atlassian.net/browse/DEV-2146).

This adds the following:
- an optional flag `--preload-dataframes` for preloading the dataframes,
- an async startup task for preloading the dataframes,
- the scaffolding for working with the cached dataframes.

Everything is backward compatible and even passing `--preload-dataframes` won't actually cause anything since the `options` returned for configuring the cache is empty.

I'll add the tests once I'll also add at least an option.

For the specific option mentioned in the comment [here](https://github.com/athenianco/athenian-api/compare/preloading-setup?expand=1#diff-62f3892e47aa88b2f700ecf5e6be31fd069986bde4307d91a55058b2ba4902e9R234), the size between the pre-`_squeeze` and post-`_squeeze` went down from ~240MB to ~40MB with production data (~17% of the original size).

This is the size percentage between the squeezed and raw version for each column (the lower the better):
```
Index                   100.000000
acc_id                   12.528637
closed                  100.000000
closed_at                50.000000
created_at               50.000000
hidden                  100.000000
merged                  100.000000
merged_at                50.000000
merged_by_login           4.092790
node_id                  35.955485
repository_full_name      3.371206
updated_at               50.000000
user_login                4.695507
dtype: float64
```


server/athenian/api/__init__.py
```diff
@@ -382,8 +392,23 @@ def main() -> Optional[AthenianApp]:
...
         max_load=float(os.getenv("ATHENIAN_MAX_LOAD", 12)))
     app.run(host=args.host, port=args.port, print=lambda s: log.info("\n" + s))
     return app
+
+
+def _get_preloader(mdb_conn, pdb_conn, log):
+
+    async def _prepare_df_cache(app):
+        log.info("Preparing MemoryCache")
+        mdb, pdb = databases.Database(mdb_conn), databases.Database(pdb_conn)
```
vmarkovtsev, line 407: This code will stress our pgbouncer and launch sequence: we'll boot two competing connection pools for mdb and pdb. I suggest adding the ability to register an event callback in `AthenianApp` that fires when we connect to all 4 our DBs and execute code in it.

server/athenian/api/__init__.py
```diff
@@ -382,8 +392,23 @@ def main() -> Optional[AthenianApp]:
...
+        await asyncio.gather(mdb.connect(), pdb.connect())
```
vmarkovtsev, line 408: Hence we'll not need `asyncio.gather` here. But if we needed it, I'd suggest `athenian.api.utils.gather` instead, exception handling is a bitch.

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+        """Return the id of the `CachedDataFrame`."""
+        return self._id
+
+    async def load(self) -> "CachedDataFrame":
+        """Load the DataFrame from the database."""
+        query = select(self._cols)
+        if self._filtering_clause:
+            query = query.where(self._filtering_clause)
+
+        res = await self._db.fetch_all(query)
```
vmarkovtsev, line 65: We should use `read_sql_async` here because datetime columns must be properly handled. I had a few nasty bugs a year ago.

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+        df = pd.DataFrame(res)
+        if self._debug_memory:
+            size = df.memory_usage(index=True, deep=True)
+            self._mem["raw"] = {"series": size, "total": sum(size)}
+
+        self._df = self._squeeze(df)
+        if self._debug_memory:
+            size = self._df.memory_usage(index=True, deep=True)
+            self._mem["processed"] = {"series": size, "total": sum(size)}
```
vmarkovtsev, line 74: There should be a way to write this to our `/status` using Prometheus. See `status_controller.py`

vmarkovtsev (changes requested):

server/athenian/api/__init__.py
se7entyse7en: It makes a lot of sense 👍

server/athenian/api/experiments/preloading/cache.py
se7entyse7en: Make sense, but I'd rather do this separately. I'll create a separate ticket for this.

se7entyse7en: https://athenianco.atlassian.net/browse/DEV-2167

server/athenian/api/experiments/preloading/cache.py
se7entyse7en: done

server/athenian/api/__init__.py
se7entyse7en: done

server/athenian/api/__init__.py
se7entyse7en: done

server/athenian/api/__init__.py
```diff
@@ -32,6 +33,8 @@
 from athenian.api import metadata
 from athenian.api.auth import Auth0
 from athenian.api.connexion import AthenianApp
+from athenian.api.experiments.preloading.cache import build_memory_cache_options, \
```
vmarkovtsev, line 36: Let's move this code in the main tree, `experiments` is for other kinds of experiments :)

server/athenian/api/__init__.py
```diff
@@ -126,6 +129,8 @@ def level_from_msg(msg: str) -> Optional[str]:
     parser.add_argument("--memcached", required=False,
                         help="memcached (users profiles, preprocessed metadata cache) address, "
                              "for example, 0.0.0.0:11211")
+    parser.add_argument("--preload-dataframes", required=False, action="store_true",
+                        help="Whether to preload data frames in memory")
```
vmarkovtsev, line 133:
```suggestion
                        help="Whether to preload DB tables in memory and refresh periodically.")
```

server/athenian/api/__init__.py
```diff
@@ -385,5 +391,16 @@ def main() -> Optional[AthenianApp]:
...
     return app
+
+
+def _preload_dataframes(log, dbs, fut):
+    log.info("Preparing MemoryCache")
+    mdb, pdb = dbs["mdb"], dbs["pdb"]
+    options = build_memory_cache_options(mdb, pdb)
+    mc = MemoryCache.get_instance(mdb, pdb, options)
+    task = asyncio.create_task(mc.load())
+    task.add_done_callback(lambda *_: log.info("MemoryCache READY"))
```
vmarkovtsev, line 406: We should think about graceful destruction here: if the app crashes before the task finishes, we'll end with "task pending but destroyed"-like errors. The easiest seems to be registering a callback on application shutdown: https://docs.aiohttp.org/en/stable/web_advanced.html#graceful-shutdown

How about making a `DataFramePreloader` class and putting code as his methods? I feel that the code will nicer.

server/athenian/api/connexion.py
```diff
@@ -282,6 +286,15 @@ async def connect_to_db(name: str, shortcut: str, db_conn: str, db_options: dict
         self._slack = self.app["slack"] = slack
         self._boot_time = psutil.boot_time()
 
+    def on_dbs_conected(self, callback):
+        """Register a callback on dbs connected."""
+        task = asyncio.ensure_future(gather(*self._db_futures.values()))
+        cb = partial(callback, {
+            db: getattr(self, db)
+            for db in ("sdb", "mdb", "pdb", "rdb")
```
vmarkovtsev, line 294:
```suggestion
            for db in self._db_futures
```

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+import sqlalchemy as sa
+from sqlalchemy import select
+
+from athenian.api.async_utils import read_sql_query
+from athenian.api.typing_utils import DatabaseLike
+
+
+def _humanize_size(v):
+    unit = ""
+    for u in ["B", "KB", "MB", "GB"]:
```
vmarkovtsev, line 14:
```suggestion
    for u in ["B", "KiB", "MiB", "GiB"]:
```
`/pedant mode off`

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+    This can be read as a normal `pandas.DataFrame`.
+
+    """
+
+    def __init__(
+        self,
+        id_: str,
+        db: DatabaseLike,
+        cols: Optional[List[sa.Column]] = None,
+        timestamp_cols: Optional[List[sa.Column]] = None,
```
vmarkovtsev, line 35: If we can infer `timestamp_cols` from `cols` then we don't need this at all.

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+        categorical_cols: Optional[List[sa.Column]] = None,
+        identifier_cols: Optional[List[sa.Column]] = None,
```
vmarkovtsev, line 37: What are identifier columns? A docstring please :pray:

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+        filtering_clause: Optional[sa.sql.ClauseElement] = None,
+        debug_memory: Optional[bool] = False,
```
vmarkovtsev, line 39: Instead of making it a ctor arg, we'd rather reference the log level of `athenian.api.CachedDataFrame`, given that it's as easy to change it as to define `flog_athenian.api.CachedDataFrame=DEBUG` environment variable.

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+        return self._id
+
+    async def load(self) -> "CachedDataFrame":
+        """Load the DataFrame from the database."""
+        query = select(self._cols)
+        if self._filtering_clause:
+            query = query.where(self._filtering_clause)
+
+        res = await read_sql_query(query, self._db, self._cols)
+        df = pd.DataFrame(res)
```
vmarkovtsev, line 67: `read_sql_query` already returns `pd.DataFrame` :)

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+        def _cast_identifiers(df, cols):
+            for c in cols:
+                max_length = df[c.name].apply(len).max()
+                df[c.name] = df[c.name].astype(f"|S{max_length}")
+            return df
+
+        return (
+            df.pipe(_cast_category, self._categorical_cols)
+            .pipe(_cast_timestamp, self._timestamp_cols)
+            .pipe(_cast_identifiers, self._identifier_cols)
```
vmarkovtsev, line 126:
```suggestion
            df.pipe(_cast_category, self._categorical_cols)
              .pipe(_cast_timestamp, self._timestamp_cols)
              .pipe(_cast_identifiers, self._identifier_cols)
```
(if flake8 is happy, otherwise whatever)

se7entyse7en:
Rebased on master.

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+        )
+
+    def __repr__(self):
```
vmarkovtsev, line 129:
```suggestion
    def __str__(self):
```
I bet you meant `__str__` here.

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+            raise ValueError("DataFrame not loaded")
+
+        return self._df.__getattr__(attr)
+
+    def __getitem__(self, key):
+        """Make items access propagate to the wrapped DataFrame."""
+        if self._df is None:
+            raise ValueError("DataFrame not loaded")
+
+        return self._df.__getitem__(key)
```
vmarkovtsev, line 158:
```suggestion
@property
def df(self) -> pd.DataFrame:
    assert self._df is not None, "Need to await refresh() at least once."
    return self._df
```
This way we don't need to proxy all dataframe's methods here.

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+                df[c.name] = df[c.name].astype("category")
+            return df
+
+        def _cast_timestamp(df, cols):
+            for c in cols:
+                df[c.name] = (df[c.name].astype("int64") / 10 ** 9).astype("int32")
+                df.loc[df[c.name] < 0, c.name] = 0
+            return df
+
+        def _cast_identifiers(df, cols):
```
vmarkovtsev, line 117:
```suggestion
        def _cast_identifiers(df: pd.DataFrame, cols: List[sa.Column]) -> pd.DataFrame:
```

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+    @classmethod
+    def get_instance(
+        cls,
+        mdb: Optional[DatabaseLike] = None,
+        pdb: Optional[DatabaseLike] = None,
+        options: Optional[Dict[str, Dict]] = None,
+        debug_memory: Optional[bool] = False,
+    ) -> "MemoryCache":
+        """Return the singleton instance."""
+        if cls._instance is None:
```
vmarkovtsev, line 189: What if we avoid the singleton pattern like this:

- Create `MemoryCache` in our `_preload_dataframes` callback.
- 
```
for db in (mdb, pdb):
    db.cache = memory_cache
```
- Change `ParallelDatabase` to redirect queries to that cache. We already intercept all the queries there, so it shouldn't be too hard.

vmarkovtsev (changes requested):

server/athenian/api/__init__.py
se7entyse7en: Agreed, I'll add a `DataFramePreloader` 👍

server/athenian/api/experiments/preloading/cache.py
se7entyse7en: It makes a lot of sense, I wasn't aware that it was that easy to change the log level of each logger, that's cool 👍 

BTW TIL what `ctor` means 😂

server/athenian/api/experiments/preloading/cache.py
se7entyse7en: hahaha damn, this is when I replaced the `fetch_all` 🤦

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def _squeeze(self, df: pd.DataFrame) -> pd.DataFrame:
+        def _cast_category(df, cols):
+            for c in cols:
+                df[c.name] = df[c.name].astype("category")
+            return df
+
+        def _cast_timestamp(df, cols):
+            for c in cols:
+                df[c.name] = (df[c.name].astype("int64") / 10 ** 9).astype("int32")
```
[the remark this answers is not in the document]
se7entyse7en: The only reason was to squeeze more space, but it has never been the majority of the space, so we could even remove this and that would save us from converting the timestamps in the controllers when filtering.

server/athenian/api/experiments/preloading/cache.py
```diff
@@ -0,0 +1,259 @@
...
+                df.loc[df[c.name] < 0, c.name] = 0
+            return df
+
+        def _cast_identifiers(df, cols):
+            for c in cols:
+                max_length = df[c.name].apply(len).max()
+                df[c.name] = df[c.name].astype(f"|S{max_length}")
```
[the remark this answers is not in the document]
se7entyse7en: ah cool, I wasn't aware of that, thanks for the hint 👍

server/athenian/api/experiments/preloading/cache.py
se7entyse7en: You're right, I was just too lazy to even type `.df` when playing in ipython 😅

server/athenian/api/experiments/preloading/cache.py
se7entyse7en: The first two points are trivial and I agree with them, it would be cleaner. 👍 

> We already intercept all the queries there, so it shouldn't be too hard

This is an interesting approach 🤔 .

If we already intercept the queries (sorry I didn't see that part of the code yet), it should be pretty easy to then use [pandasql](https://github.com/yhat/pandasql/) that I mentioned to you in one of our calls already: so we should be able to just compile the SQLA query and pass it to pandasql. In case of an error of missing table/dataframe then it will fall back to the "normal" DB query. I never used pandasql and I'm not much familiar with query compilation, but worth exploring this option. This should also save "translating" the current queries into pandas operations as you can see [here in this draft PR](https://github.com/athenianco/athenian-api/pull/1337/files#diff-0ad06f21667ae6b7fa99290cd35ebd84fdcab18d79d7711e6c505449bb328e9fR70).

I'm just wondering whether there are some weird cases that we might bump into by using pandasql such as unsupported things. Now that I'm checking probably it won't be that straightforward since it works with SQLite syntax, but in production, the query would be prepared for PG, so we would need to prepare the query for both dialects I guess?

For sure it's something that I'd like to have long-term: a unified interface with SQL query that is aware of whether to query the in-mem cache or the underlying DB. But I'm wondering whether this can at least wait for once we see the gain on at least the first fully-speeded-up endpoint before investing the time on this.

wdyt?

server/athenian/api/experiments/preloading/cache.py
se7entyse7en: Actually I meant `__repr__`. I know that `__repr__` should be a "unique" representation of the object, but I took inspiration from [how `pandas` does it](https://github.com/pandas-dev/pandas/blob/9be0b8224622bc340333f05dd62be2d02996945a/pandas/core/frame.py#L922). Basically it was simply because it was handier when playing in ipython, but it's fine as well for me to have `__str__` instead.

se7entyse7en:
Partially addressed all the issues, as waiting for your input for a couple of remaining ones. 👍

server/athenian/api/experiments/preloading/cache.py
se7entyse7en: Anyway I'm exploring this idea since it turned out that it's super simple compile the query in any dialect:
```
In [24]: from sqlalchemy.dialects import postgresql, sqlite

In [25]: from athenian.api.models.metadata.github import PullRequest

In [26]: query = sa.select([PullRequest.acc_id, PullRequest.node_id]).where(PullRequest.acc_id.in_([1,2,3]))

In [27]: str(query.compile(dialect=postgresql.dialect(), compile_kwargs={"literal_binds": True}))
Out[27]: 'SELECT github.api_pull_requests.acc_id, github.api_pull_requests.node_id \nFROM github.api_pull_requests \nWHERE github.api_pull_requests.acc_id IN (1, 2, 3)'

In [28]: str(query.compile(dialect=sqlite.dialect(), compile_kwargs={"literal_binds": True}))
Out[28]: 'SELECT github.api_pull_requests.acc_id, github.api_pull_requests.node_id \nFROM github.api_pull_requests \nWHERE github.api_pull_requests.acc_id IN (1, 2, 3)'
```

I'm gonna try this approach + `pandasql` in the draft PR to see whether a real query works as expected.

se7entyse7en: damn I just realized that the latest release for `pandasql` was in 2016... and py2.7 only 🤦

se7entyse7en: ok nvm the code is super tiny, there shouldn't be any issue, I'll test it. When installing it just builds the wheel without issues 👍

server/athenian/api/experiments/preloading/cache.py
vmarkovtsev: Sure, we can keep the arg if you like it :smile: 

```python
from athenian.api import metadata

class Foo:
    _log = logging.getLogger(f"{metadata.__package__}.Foo")

    def __init__(self):
        debug_mode = self._log.isEnabledFor(logging.DEBUG)
```

server/athenian/api/experiments/preloading/cache.py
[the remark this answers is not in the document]
vmarkovtsev: The timestamps are very notorious, some contain timezones, some don't. I always struggle with them across the codebase. The low-level code works with np.datetime64 and that dtype does not support timezones... it makes so much fun. Anyway, we'd rather ignore this optimization for now, that would be a separate challenge.

server/athenian/api/experiments/preloading/cache.py
vmarkovtsev: Yeah, pandas is not an example of great idiomatic Python haha

server/athenian/api/experiments/preloading/cache.py
vmarkovtsev: `literal_binds` does not always work, unfortunately. We use some Postgres and SQLite specific extensions that do not compile in that mode, for example, HSTORE, JSON, ARRAY. AFAIR even datetimes do not inline well. This is a piece of serious work. I studied the machinery in fine detail while maintaining the Postgres backend of `databases`. The low-level asyncpg code deals with SQL string and a series of arguments. The serialization of the arguments can be stolen from asyncpg, I think. Basically, we need our own asyncpg.Connection imposter. And yes, some engine to convert either SQLAlchemy or raw SQL to pandas commands.

So you originally planned to swap DB queries with pandas without hacking the DB interface by rewriting each query manually, I see.

server/athenian/api/experiments/preloading/cache.py
se7entyse7en: haha I'll keep it as `__str__` 👍

server/athenian/api/experiments/preloading/cache.py
se7entyse7en: Ah damn, I forgot about those extensions. As I mentioned in DM despite I'd love to have a common interface in the future, the effort that might be required for doing it properly _now_ scares me. I'd rather go easy for now and re-evaluate later on once we are benefiting already from the performance speedup.

In any case, I'll have a look at attaching the `MemoryCache` to the databases as you proposed 👍

server/athenian/api/__init__.py
```diff
@@ -357,6 +359,8 @@ def compose_db_options(mdb: str, sdb: str, pdb: str, rdb: str) -> Dict[str, Dict
 
 def main() -> Optional[AthenianApp]:
     """Server's entry point."""
+    from athenian.api.experiments.preloading.cache import MemoryCachePreloader
```
vmarkovtsev: A conflict in imports? I'll move `main()` away from `__init__`.

server/athenian/api/__init__.py
```diff
@@ -372,6 +376,8 @@ def main() -> Optional[AthenianApp]:
     auth0_cls = create_auth0_factory(args.force_user)
     kms_cls = None if args.no_google_kms else AthenianKMS
     slack = create_slack(log)
+
```
vmarkovtsev, line 378:
```suggestion
```

server/athenian/api/connexion.py
```diff
@@ -161,6 +162,8 @@ def __init__(self,
                  sdb_options: Optional[Dict[str, Any]] = None,
                  pdb_options: Optional[Dict[str, Any]] = None,
                  rdb_options: Optional[Dict[str, Any]] = None,
+                 on_startup_callbacks: Optional[List[Callable[..., None]]] = None,
+                 on_shutdown_callbacks: Optional[List[Callable[..., None]]] = None,
```
vmarkovtsev, line 166:
```suggestion
                 on_startup_callbacks: Optional[Collection[Callable[..., None]]] = None,
                 on_shutdown_callbacks: Optional[Collection[Callable[..., None]]] = None,
```

server/athenian/api/connexion.py
```diff
@@ -264,7 +267,12 @@ async def connect_to_db(name: str, shortcut: str, db_conn: str, db_options: dict
                 self.log.exception("Failed to connect to the %s DB at %s", name, db_conn)
                 raise GracefulExit() from None
 
-        self.app.on_shutdown.append(self.shutdown)
+        if on_startup_callbacks:
+            self.app.on_startup.extend(on_startup_callbacks)
+
+        on_shutdown_callbacks = on_shutdown_callbacks or []
```
vmarkovtsev, line 273:
```suggestion
        on_shutdown_callbacks = list(on_shutdown_callbacks or [])
```

server/athenian/api/__init__.py
```diff
@@ -357,6 +358,8 @@ def compose_db_options(mdb: str, sdb: str, pdb: str, rdb: str) -> Dict[str, Dict
 
 def main() -> Optional[AthenianApp]:
     """Server's entry point."""
+    from athenian.api.preloading.cache import MemoryCachePreloader
```
vmarkovtsev, line 361: This must be a conflict of the imports? I'll move `main()` to a separate module.

se7entyse7en:
Moved `athenian.api.experiments.preloading.cache` into `athenian.api.preloading.cache` and removed singleton in favor of a `MemoryCache` for each `ParallelDatabase` which is much cleaner.

server/athenian/api/preloading/cache.py
```diff
@@ -0,0 +1,274 @@
...
+    ):
+        """Initialize a `CachedDataFrame`.
+
+        The wrapped `pandas.DataFrame` can be accessed through the `df` property and
+        its dtypes are processed in order to reduce the memory usage.
+
+        :param cols: Columns to include.
+        :param categorical_cols: Columns that can be cast to `category` dtype
+        :param identifier_cols: Columns that can be cast to bytes, `S` dtype
+
```
vmarkovtsev, line 50:
```suggestion
```

vmarkovtsev (approved):
I couldn't find anything important. As soon as I refactor `__init__` should be good to merge.

vmarkovtsev:
https://github.com/athenianco/athenian-api/pull/1348

se7entyse7en:
Thanks! Indeed it was because of a circular import, I'll wait for that PR to be merged 👍

vmarkovtsev:
Merged #1348

se7entyse7en:
Rebased and conflicts solved.

PR #1337, opened 2021-04-19
Preloading fetch prs

se7entyse7en:
Closes [DEV-2147](https://athenianco.atlassian.net/browse/DEV-2147).

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -74,7 +74,7 @@ jobs:
     strategy:
       matrix:
         db: [sqlite, postgres]
-        type: [heater, slim, fat]
+        type: [heater, slim, fat, fat-preloading]
       fail-fast: false
     services:
       memcached:
@@ -193,6 +193,14 @@ jobs:
           export SLACK_API_TOKEN=
         fi
         godotenv -s pytest -n 4 $LIMIT --log-level=info --cov-report=xml --cov=athenian.api tests/controllers/test_filter_controller.py tests/controllers/test_metrics_controller.py tests/controllers/test_histograms_controller.py tests/controllers/test_pagination_controller.py tests/controllers/test_jira_controller.py
+    - name: test fat with preloading
+      if: matrix.type == 'fat-preloading'
+      working-directory: server
+      run: |
+        set -x
+        export WITH_PRELOADING=1
+        export SLACK_API_TOKEN=
+        godotenv -s pytest -n 4 $LIMIT --log-level=info --cov-report=xml --cov=athenian.api tests/controllers/test_metrics_controller.py tests/controllers/test_histograms_controller.py
     - name: test heater
       if: matrix.type == 'heater'
       working-directory: server
--- a/server/athenian/api/connexion.py
+++ b/server/athenian/api/connexion.py
@@ -293,6 +293,7 @@ class AthenianApp(connexion.AioHttpApp):
         on_shutdown_callbacks = list(on_shutdown_callbacks or [])
         on_shutdown_callbacks.append(self.shutdown)
         self.app.on_shutdown.extend(on_shutdown_callbacks)
+        self._on_dbs_conected_events = []
         # schedule the DB connections when the server starts
         self._db_futures = {
             args[1]: asyncio.ensure_future(connect_to_db(*args))
@@ -313,15 +314,19 @@ class AthenianApp(connexion.AioHttpApp):
     def on_dbs_conected(self, callback):
         """Register a callback on dbs connected."""
 
-        def callback_wrapper(*_):
+        async def task_wrapper(*_):
+            await gather(*self._db_futures.values())
             dbs = {
                 db: getattr(self, db)
                 for db in self._db_futures
             }
             callback(**dbs)
 
-        task = asyncio.ensure_future(gather(*self._db_futures.values()))
-        task.add_done_callback(callback_wrapper)
+        self._on_dbs_conected_events.append(asyncio.ensure_future(task_wrapper()))
+
+    async def wait_for_dbs_connected_callbacks(self):
+        """Wait until all callbacks on dbs connected are done."""
+        await gather(*self._on_dbs_conected_events)
 
     def __del__(self):
         """Check that shutdown() was called."""
--- a/server/athenian/api/connexion.py
+++ b/server/athenian/api/connexion.py
@@ -293,7 +293,7 @@ class AthenianApp(connexion.AioHttpApp):
         on_shutdown_callbacks = list(on_shutdown_callbacks or [])
         on_shutdown_callbacks.append(self.shutdown)
         self.app.on_shutdown.extend(on_shutdown_callbacks)
-        self._on_dbs_conected_events = []
+        self._on_dbs_connected_callbacks = []
         # schedule the DB connections when the server starts
         self._db_futures = {
             args[1]: asyncio.ensure_future(connect_to_db(*args))
@@ -311,7 +311,7 @@ class AthenianApp(connexion.AioHttpApp):
         self._slack = self.app["slack"] = slack
         self._boot_time = psutil.boot_time()
 
-    def on_dbs_conected(self, callback):
+    def on_dbs_connected(self, callback):
         """Register a callback on dbs connected."""
 
         async def task_wrapper(*_):
@@ -322,11 +322,11 @@ class AthenianApp(connexion.AioHttpApp):
             }
             callback(**dbs)
 
-        self._on_dbs_conected_events.append(asyncio.ensure_future(task_wrapper()))
+        self._on_dbs_connected_callbacks.append(asyncio.ensure_future(task_wrapper()))
 
     async def wait_for_dbs_connected_callbacks(self):
         """Wait until all callbacks on dbs connected are done."""
-        await gather(*self._on_dbs_conected_events)
+        await gather(*self._on_dbs_connected_callbacks)
 
     def __del__(self):
         """Check that shutdown() was called."""
--- a/server/athenian/api/controllers/features/entries.py
+++ b/server/athenian/api/controllers/features/entries.py
@@ -76,6 +76,8 @@ def _postprocess_cached_facts(result: Tuple[Dict[str, List[PullRequestFacts]], b
 class MetricEntriesCalculator:
     """Calculator for different metrics."""
 
+    miner = PullRequestMiner
+
     def __init__(self, account: int, meta_ids: Tuple[int, ...],
                  mdb: Database, pdb: Database, rdb: Database,
                  cache: Optional[aiomcache.Client]):
@@ -521,7 +523,7 @@ class MetricEntriesCalculator:
                 len(participants.get(prpk.AUTHOR, [])) > unfresh_participants_threshold) and \
                 not fresh and not (participants.keys() - {prpk.AUTHOR, prpk.MERGER}):
             facts = await fetch_pull_request_facts_unfresh(
-                precomputed_facts, ambiguous, time_from, time_to, repositories,
+                self.miner, precomputed_facts, ambiguous, time_from, time_to, repositories,
                 participants, labels, jira, exclude_inactive, branches,
                 default_branches, release_settings, self._account, self._meta_ids,
                 self._mdb, self._pdb, self._rdb, self._cache)
@@ -538,14 +540,14 @@ class MetricEntriesCalculator:
         # the adjacent out-of-range pieces [date_from, time_from] and [time_to, date_to]
         # are effectively discarded later in BinnedMetricCalculator
         tasks = [
-            PullRequestMiner.mine(
+            self.miner.mine(
                 date_from, date_to, time_from, time_to, repositories, participants,
                 labels, jira, branches, default_branches, exclude_inactive, release_settings,
                 self._account, self._meta_ids, self._mdb, self._pdb, self._rdb, self._cache,
                 pr_blacklist=blacklist),
         ]
         if jira and precomputed_facts:
-            tasks.append(PullRequestMiner.filter_jira(
+            tasks.append(self.miner.filter_jira(
                 precomputed_facts, jira, self._meta_ids, self._mdb, self._cache,
                 columns=[PullRequest.node_id]))
             (miner, unreleased_facts, matched_bys, unreleased_prs_event), filtered = \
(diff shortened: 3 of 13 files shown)


server/athenian/api/preloading/cache.py
```diff
@@ -239,35 +259,41 @@ def _done(self, *_):
...
+                ],
+                "identifier_cols": [
+                    PullRequest.node_id,
+                    PullRequest.merge_commit_id,
+                    PullRequest.merge_commit_sha,
+                    PullRequest.htmlurl,
+                ],
+            },
+        },
+    }
```
se7entyse7en, line 299: Some more columns were required, we'll be able to check memory once we expose it through prometheus. I didn't manually check it since my prod dump has an outdated schema and will take me a lot to check again.

se7entyse7en:
Ah damn sorry, forgot to add tests on CI

se7entyse7en:
added

se7entyse7en:
@vmarkovtsev this is ready for review, all tests are passing 👍

se7entyse7en:
Rebased and conflicts solved.

server/athenian/api/preloading/cache.py
se7entyse7en: Well actually I can easily do it by preloading directly from production without dumping, I'll add it in a separate PR a script to do it that would be useful for development.

server/athenian/api/controllers/features/github/unfresh_pull_request_metrics.py
```diff
@@ -50,21 +51,21 @@ async def fetch_pull_request_facts_unfresh(done_facts: Dict[str, PullRequestFact
     :return: Map from PR node IDs to their facts.
     """
     add_pdb_hits(pdb, "fresh", 1)
-    blacklist = PullRequest.node_id.notin_any_values(done_facts)
+    blacklist = done_facts
```
vmarkovtsev, line 54: Wow-wow this affects prod much and not properly covered with tests, we cannot edit it.

server/athenian/api/controllers/miners/github/pull_request.py
```diff
@@ -208,11 +208,13 @@ async def _mine(cls,
         assert (updated_min is None) == (updated_max is None)
         time_from, time_to = (pd.Timestamp(t, tzinfo=timezone.utc) for t in (date_from, date_to))
         if pr_blacklist is not None:
-            pr_blacklist, ambiguous = pr_blacklist
-            if len(pr_blacklist) > 0:
-                pr_blacklist = PullRequest.node_id.notin_any_values(pr_blacklist)
+            pr_blacklist_map, ambiguous = pr_blacklist
+            if len(pr_blacklist_map) > 0:
+                pr_blacklist_keys = list(pr_blacklist_map)
```
vmarkovtsev, line 213: Again, we cannot change this in prod. These lines are literally written with my blood and sweat.

server/athenian/api/controllers/miners/github/pull_request.py
```diff
@@ -251,18 +253,17 @@ async def dummy_unreleased():
             # These PRs are released by branch and not by tag, and we require by tag.
             # Now fetch only them, respecting the filters.
             # TODO(vmarkovtsev): do not load the releases from scratch in map_releases_to_prs()
-            inverse_pr_blacklist = PullRequest.node_id.in_(
-                list(chain.from_iterable(missed_prs.values())))
+            pr_whitelist = list(chain.from_iterable(missed_prs.values()))
```
vmarkovtsev, line 256: Indeed, an inverse blacklist is called a whitelist hehe

server/athenian/api/controllers/miners/github/pull_request.py
```diff
@@ -841,8 +844,10 @@ async def _fetch_prs_by_filters(cls,
             filters.append(PullRequest.updated_at >= time_from)
         if updated_min is not None:
             filters.append(PullRequest.updated_at.between(updated_min, updated_max))
-        if pr_blacklist is not None:
-            filters.append(pr_blacklist)
+        if pr_blacklist is not None and len(pr_blacklist) > 0:
```
vmarkovtsev, line 847: So this should stay a `BinaryExpression`. The reason is performance, the lists can have big lengths and there are cases when I have to call `fetch_prs` twice with the same blacklist. The lengths may be so big that simply constructing the `notin` expression consumes considerable time inside SQLAlchemy. I recommend making the whitelist an optional BinaryExpression, too.

server/athenian/api/experiments/preloading/entries.py
```diff
@@ -1,8 +1,137 @@
...
+
+        assert (updated_min is None) == (updated_max is None)
+
+        df = mdb.cache.dfs["prs"].df
+
+        mask = (
+            (~df["closed"] | (df["closed_at"] >= time_from))
+            & (df["created_at"] < time_to)
+            & (df["acc_id"].isin(meta_ids))
+            & (~df["hidden"])
```
vmarkovtsev, line 72: We can exclude hidden PRs straight away from the preloaded table, they are not used anywhere.

server/athenian/api/experiments/preloading/entries.py
```diff
@@ -1,8 +1,137 @@
...
+            & (df["repository_full_name"].isin(repositories))
+        )
+
+        if pr_blacklist is not None:
```
vmarkovtsev, line 76: I see why you had to switch from BinaryExpression to plain lists. I can help you with traversing the sqla AST here, I did it a few times already. It's not so difficult actually.

server/athenian/api/experiments/preloading/entries.py
```diff
@@ -1,8 +1,137 @@
...
+                updated_min=updated_min,
+                updated_max=updated_max,
+            )
+
+        assert (updated_min is None) == (updated_max is None)
+
+        df = mdb.cache.dfs["prs"].df
+
+        mask = (
+            (~df["closed"] | (df["closed_at"] >= time_from))
```
vmarkovtsev, line 69: As a rule of thumb, try to avoid Pandas operations because they are much slower than plain numpy. Proved by numerous hours spent in the profiler. For example, here:
```python
(~df["closed"].values | (df["closed_at"].values >= np.array([time_from], dtype=df["closed_at"].dtype))
```
will work faster. There are some caveats with datetime64 though.

server/athenian/api/experiments/preloading/entries.py
```diff
@@ -1,8 +1,137 @@
...
+            & (df["created_at"] < time_to)
+            & (df["acc_id"].isin(meta_ids))
+            & (~df["hidden"])
+            & (df["repository_full_name"].isin(repositories))
+        )
+
+        if pr_blacklist is not None:
+            pr_blacklist = [v.encode() for v in pr_blacklist]
+            mask &= ~df["node_id"].isin(pr_blacklist)
```
vmarkovtsev, line 78: Here, this will be way faster:
```suggestion
            mask &= np.in1d(df["node_id"].values.astype("S"), pr_blacklist, inverse=True)
```
provided `pr_blacklist` is already an "S" array.

server/athenian/api/preloading/cache.py
```diff
@@ -95,6 +97,35 @@ async def refresh(self) -> "CachedDataFrame":
...
+    ) -> pd.DataFrame:
+        """Filter the wrapped `pandas.Datarame` with the provided mask.
+
+        :param mask: boolean mask for filtering the wrapped DataFrame.
+        :param columns: the columns to select.
+        :param index: the column to set as index.
+        :param uncast: whether to uncast the columns dtypes, see the `_squeeze` method.
+
+        """
+        df = self._df[mask].copy()
```
vmarkovtsev, line 115:
```suggestion
        df = self._df.take(np.nonzero(mask)[0])
```
this works much, much faster.

vmarkovtsev (changes requested):
Some preliminary suggestions.

server/athenian/api/experiments/preloading/entries.py
se7entyse7en: What about this?
```
In [29]: whitelist = PullRequest.node_id.in_(['abc','def'])

In [30]: blacklist = PullRequest.node_id.notin_(['ghi','jkl'])

In [31]: [bp.value for bp in whitelist.get_children()[1].get_children()[0].get_children()]
Out[31]: ['abc', 'def']

In [32]: [bp.value for bp in blacklist.get_children()[1].get_children()[0].get_children()]
Out[32]: ['ghi', 'jkl']
```

.github/workflows/push.yml
```diff
@@ -201,6 +201,19 @@ jobs:
...
+      run: |
+        set -x
+        export WITH_PRELOADING=1
+        # if [ ${{ matrix.db }} = "sqlite" ]; then
+        #   export LIMIT="--limit 0.9"
+        # fi
+        if [ ${{ matrix.db }} = "postgres" ]; then
+          export SLACK_API_TOKEN=
+        fi
+        godotenv -s pytest -n 4 $LIMIT --log-level=info --cov-report=xml --cov=athenian.api tests/controllers/test_filter_controller.py tests/controllers/test_metrics_controller.py tests/controllers/test_histograms_controller.py tests/controllers/test_pagination_controller.py tests/controllers/test_jira_controller.py
```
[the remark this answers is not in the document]
se7entyse7en: For sure they're not in this PR, so I'll remove them for now.

server/athenian/api/controllers/features/entries.py
```diff
@@ -72,6 +72,8 @@ def _postprocess_cached_facts(result: Tuple[Dict[str, List[PullRequestFacts]], b
 class MetricEntriesCalculator:
     """Calculator for different metrics."""
 
+    PRMiner = PullRequestMiner
```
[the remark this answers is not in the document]
se7entyse7en: Renamed as `miner` since indeed it's dep inj.

server/athenian/api/connexion.py
```diff
@@ -298,17 +299,24 @@ async def connect_to_db(name: str, shortcut: str, db_conn: str, db_options: dict
...
+        ev = asyncio.Event()
+        self._on_dbs_conected_events.append(ev)
 
         def callback_wrapper(*_):
             dbs = {
                 db: getattr(self, db)
                 for db in self._db_futures
             }
             callback(**dbs)
+            ev.set()
```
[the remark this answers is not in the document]
se7entyse7en: The problem is that here I actually want not only to wait for the dbs connected, but also on the callback being finished. AFAIK there's no other way to `await` for the callback since it's not an `async` task.

server/athenian/api/experiments/preloading/entries.py
se7entyse7en: Probably make sense to apply your changes for all operations except for those related to `datetime64`?

server/athenian/api/experiments/preloading/entries.py
se7entyse7en: I'm wondering whether it makes sense to have these utils somewhere as we'll have a lot of `np.ind1d ...` calls.  Is this going to be faster even if `pr_blacklist` has to be converted to `"S"`? How much faster is it? Actually I could just measure it.

server/athenian/api/experiments/preloading/entries.py
vmarkovtsev: Yes, smth like this, exactly. They can be `in_values` and `not_in_values`, too, we should check that they work the same. Those are defined in `athenian/api/models/__init__.py`

server/athenian/api/connexion.py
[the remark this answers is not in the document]
vmarkovtsev: I mean this:

```python
async def task_wrapper():
    await gather(*self._db_futures.values())
    dbs = {
        db: getattr(self, db)
        for db in self._db_futures
    }
    callback(**dbs)
asyncio.create_task(task_wrapper(), name=f"on_dbs_conected({callback})")
```

vmarkovtsev (changes requested):

server/athenian/api/experiments/preloading/entries.py
vmarkovtsev: Smth like this, yes. I'd say if it works with `np.datetime64` then awesome, if it doesn't - we have to fall back to Pandas.

server/athenian/api/experiments/preloading/entries.py
vmarkovtsev: "U" -> "S" conversion comes almost for free. What matters here is that both arrays become strongly typed vs. `object` and `np.in1d` takes the fastest path.

server/athenian/api/experiments/preloading/entries.py
se7entyse7en:
```
In [4]: extract = lambda be: [bp.value for bp in be.get_children()[1].get_children()[0].get_children()]

In [5]: extract(PullRequest.node_id.in_(['abc','def']))
Out[5]: ['abc', 'def']

In [6]: extract(PullRequest.node_id.notin_(['abc','def']))
Out[6]: ['abc', 'def']

In [7]: extract(PullRequest.node_id.in_any_values(['abc','def']))
Out[7]: ['abc', 'def']

In [8]: extract(PullRequest.node_id.notin_any_values(['abc','def']))
Out[8]: ['abc', 'def']
```

confirmed 👍

server/athenian/api/connexion.py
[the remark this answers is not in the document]
se7entyse7en: done, I needed to use `ensure_future` since there was no loop running yet though.

se7entyse7en:
@vmarkovtsev do you think that we can postpone the other optimizations? I'd rather move forward with the other parts rather than fine-tune this to reach faster an improved prototype for the `/metrics` endpoint. wdyt?

vmarkovtsev:
It requires some practice to replace pandas with numpy, true. We can always do it later, just leave some `FIXME` comment around.

se7entyse7en:
> It requires some practice to replace pandas with numpy

exactly

> just leave some FIXME comment around.

done: https://github.com/athenianco/athenian-api/pull/1337/commits/73d7af5ea3505e27e91e20ab7a93cf1368a519a6

se7entyse7en:
[Fixed typos](https://github.com/athenianco/athenian-api/pull/1337/commits/e81aa7f18738b4641e73c6142a11194ba5b4020c) and rebased on `master`.

vmarkovtsev (approved):

PR #1392, opened 2021-05-04
Healthz endpoint

se7entyse7en:
Closes [DEV-2165](https://athenianco.atlassian.net/browse/DEV-2165).

### DIFF
--- a/server/tests/controllers/test_status_controller.py
+++ b/server/tests/controllers/test_status_controller.py
@@ -61,3 +61,10 @@ async def test_objgraph_400(client, headers, query):
         method="GET", path="/objgraph?" + query, headers=headers, json={},
     )
     assert response.status == 400
+
+
+async def test_healthz_200(client, headers):
+    response = await client.request(
+        method="GET", path="/v1/healthz", headers=headers, json={},
+    )
+    assert response.status == 200
--- a/server/athenian/api/__main__.py
+++ b/server/athenian/api/__main__.py
@@ -294,6 +294,7 @@ def main() -> Optional[AthenianApp]:
         sdb_conn=args.state_db,
         pdb_conn=args.precomputed_db,
         rdb_conn=args.persistentdata_db,
+        preloading_enabled=args.preload_dataframes,
         **compose_db_options(args.metadata_db,
                              args.state_db,
                              args.precomputed_db,
--- a/server/athenian/api/connexion.py
+++ b/server/athenian/api/connexion.py
@@ -169,6 +169,7 @@ class AthenianApp(connexion.AioHttpApp):
                  ui: bool,
                  client_max_size: int,
                  max_load: float,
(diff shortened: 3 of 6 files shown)


server/tests/controllers/test_status_controller.py
```diff
@@ -61,3 +61,10 @@ async def test_objgraph_400(client, headers, query):
         method="GET", path="/objgraph?" + query, headers=headers, json={},
     )
     assert response.status == 400
+
+
+async def test_healthz_200(client, headers):
```
se7entyse7en, line 66: I'll add a test for 425 once [this](https://github.com/athenianco/athenian-api/pull/1337) gets merged since it already has some utilities in place for testing with preloading enabled.

vmarkovtsev (changes requested):
Sorry for not catching this later. We should do this differently, I think.

We already have `/status` that reports various information to Prometheus. We would rather add another section there related to preloading. This is much better than a new endpoint because:

- No new code, no new endpoint. The best code is no code.
- Better extensibility. No need to change the spec to add something.
- We can append other useful stats about the preloader and plot everything in Grafana with low coding overhead.

Checking the preloading status in the script will change from checking 425 to `grep athenian.preloading.status` in the response or smth similar.

se7entyse7en:
> We already have /status that reports various information to Prometheus. We would rather add another section there related to preloading. 

I totally agree with you, which is indeed something that is tracked in [DEV-2167](https://athenianco.atlassian.net/browse/DEV-2167). It's also related to [DEV-2127](https://athenianco.atlassian.net/browse/DEV-2127), but this is more about the memory usage "at rest" which can be computed with the script that I wrote last week.

> Checking the preloading status in the script will change from checking 425 to grep athenian.preloading.status in the response or smth similar.

Ok, I just re-read the docs for liveness and readiness probes, I totally forgot that those can also be commands rather than calls to endpoints. So what we can do is the following:
- we add this tracking to `/status` so that we also close [DEV-2167](https://athenianco.atlassian.net/browse/DEV-2167),
- we change the readiness probe of both the [API](https://github.com/athenianco/infrastructure/blob/master/helm/charts/api/templates/deployment-api.yml#L88) and [Nginx](https://github.com/athenianco/infrastructure/blob/master/helm/charts/api/templates/deployment-api.yml#L237) to call a script that will basically to something like:

```
curl /status | grep athenian.preloading.status | whatever
```

\cc @lwsanty

lwsanty:
https://athenian.slack.com/archives/CRJC5VDLL/p1620290452040600?thread_ts=1620286624.035100&cid=CRJC5VDLL

se7entyse7en:
Closing this in favor of https://athenian.slack.com/archives/CRJC5VDLL/p1620292831055600?thread_ts=1620286624.035100&cid=CRJC5VDLL

PR #1394, opened 2021-05-05
[DEV-2110] Add the spec of /metrics/code_checks

vmarkovtsev:


server/athenian/api/openapi/openapi.yaml
```diff
@@ -7320,3 +7376,235 @@ components:
...
+        granularities:
+          items:
+            $ref: '#/components/schemas/Granularity'
+          type: array
+        quantiles:
+          $ref: '#/components/schemas/Quantiles'
+        account:
+          description: Session account ID.
+          type: integer
+        split_by_check_runs:
```
vmarkovtsev, line 7436: > option to get the distribution by number of checks run in the suites 
This way means that we will have to make two separate API calls to fetch the CI metrics if we want both overall values and the distribution by suite size.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -7320,3 +7376,235 @@ components:
...
+          description: |-
+            Check runs must be triggered by commits authored by these people.
+            We aggregate by each group so that you can request metrics of several teams at once.
+            We treat `commit_authors` as another group, if specified.
+          items:
+            items:
+              $ref: '#/components/schemas/DeveloperID'
+            type: array
+          type: array
+        jira:
```
vmarkovtsev, line 7480: JIRA filtering comes for free.

server/athenian/api/openapi/openapi.yaml
warenlg: Distribution by number of checks run in the suites (aka suite size) will be used only for:
- number of suites
- suite run time

server/athenian/api/openapi/openapi.yaml
```diff
@@ -7320,3 +7377,242 @@ components:
...
+          $ref: '#/components/schemas/Quantiles'
+        account:
+          description: Session account ID.
+          type: integer
+        split_by_check_runs:
+          description: Calculate metrics separately for each number of check runs in suite.
+          type: boolean
+      required:
+      - account
+      - date_from
```
warenlg, line 7442: date_from and date_to define an interval. A set of checks belongs to that interval according to their starting date. However, for the checks that map to PRs, we need to go backward and include all the checks run into those PRs to be able to compute accurately metrics such as:
- suites_per_pr
- suite_time_per_pr
- prs_with_checks_count

server/athenian/api/openapi/openapi.yaml
```diff
@@ -7320,3 +7377,242 @@ components:
...
+      type: object
+    CodeCheckMetricID:
+      description: |-
+        * `chk-suites-count` - number of executed check suites.
+        * `chk-successful-suites-count` - number of successfully executed check suites.
+        * `chk-failed-suites-count` - number of failed check suites.
+        * `chk-cancelled-suites-count` - number of cancelled check suites.
+        * `chk-success-ratio` - number of successful check suites divided by `chk-suites-count`.
+        * `chk-suite-time` - average check suite execution time.
+        * `chk-suites-per-pr` - average number of times check suites executed in a pull request.
```
warenlg, line 7497: Let's agree on what to consider "executed" in general. I'd say `status=completed` and `conclusion!=skipped`

server/athenian/api/openapi/openapi.yaml
```diff
@@ -7320,3 +7377,242 @@ components:
...
+        * `chk-suite-time-per-pr` - average check suite execution time by pull request. The difference with `chk-suite-time` is that we don't consider checks triggered by direct pushes, e.g. by merge commits.
```
warenlg, line 7498: The main difference to me with `chk-suite-time` is more that `chk-suite-time-pr` can be relatively high is `chk-suite-per-pr` is high

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: Good point.

vmarkovtsev:
We've probably missed some details here, but this is just too big. We'll figure on the go.

PR #1424, opened 2021-05-12
Refactor MemoryCache to remove string constant IDs

vmarkovtsev:
Rationale: if I hold Ctrl and click `prs` in `MCID`, I immediately jump to the usages of that cached table in PyCharm.


se7entyse7en (approved):
Thanks! Much cleaner!

PR #1420, opened 2021-05-11
Preload release match timespan

se7entyse7en:
Closes [DEV-2306](https://athenianco.atlassian.net/browse/DEV-2306).

Depends on [this](https://github.com/athenianco/athenian-api/pull/1419).

PR #1428, opened 2021-05-13
Preload open prs facts

se7entyse7en:
Closes [DEV-2311](https://athenianco.atlassian.net/browse/DEV-2311).

Depends on [this](https://github.com/athenianco/athenian-api/pull/1420).

### DIFF
--- a/server/athenian/api/experiments/preloading/entries.py
+++ b/server/athenian/api/experiments/preloading/entries.py
@@ -13,6 +13,7 @@ from athenian.api.controllers.features.github.unfresh_pull_request_metrics impor
     UnfreshPullRequestFactsFetcher
 from athenian.api.controllers.miners.filters import JIRAFilter, LabelFilter
 from athenian.api.controllers.miners.github.branches import BranchMiner
+from athenian.api.controllers.miners.github.precomputed_prs import OpenPRFactsLoader
 from athenian.api.controllers.miners.github.pull_request import PullRequestMiner
 from athenian.api.controllers.miners.github.release_load import \
     match_groups_to_conditions, ReleaseLoader
@@ -106,10 +107,17 @@ class PreloadedReleaseToPullRequestMapper(ReleaseToPullRequestMapper):
     release_loader = PreloadedReleaseLoader
 
 
+class PreloadedOpenPRFactsLoader(OpenPRFactsLoader):
+    """Loader for preloaded open PRs facts."""
+
+    pass
+
+
 class PreloadedUnfreshPullRequestFactsFetcher(UnfreshPullRequestFactsFetcher):
     """Fetcher for preloaded unfresh pull requests facts."""
 
     release_loader = PreloadedReleaseLoader
+    open_prs_facts_loader = PreloadedOpenPRFactsLoader
 
 
 class PreloadedBranchMiner(BranchMiner):
--- a/server/athenian/api/experiments/preloading/entries.py
+++ b/server/athenian/api/experiments/preloading/entries.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 import functools
 import operator
-from typing import Collection, Dict, Iterable, List, Optional, Sequence, Set, Tuple, Union
+from typing import Collection, Dict, Iterable, List, Mapping, Optional, Sequence, Set, Tuple, Union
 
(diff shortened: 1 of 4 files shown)


server/athenian/api/experiments/preloading/entries.py
```diff
@@ -73,10 +105,60 @@ class PreloadedReleaseToPullRequestMapper(ReleaseToPullRequestMapper):
...
+            date_range = build_days_range(time_from, time_to)
+            mask &= df["activity_days"].apply(lambda v: bool(set(v).intersection(date_range)))
+
+        open_prs_facts = cached_df.filter(mask, columns=[
+            "pr_node_id", "repository_full_name", "data"])
+        if open_prs_facts.empty:
+            return {}
+
+        facts = {}
+        for row in open_prs_facts.iterrows():
```
vmarkovtsev, line 147: Same here - `iterrows()` considered harmful!

vmarkovtsev (changes requested):

server/athenian/api/experiments/preloading/entries.py
se7entyse7en: done

server/athenian/api/experiments/preloading/entries.py
```diff
@@ -73,10 +109,63 @@ class PreloadedReleaseToPullRequestMapper(ReleaseToPullRequestMapper):
...
+        for pr_node_id, repository_full_name, data in zip(
+                open_prs_facts["pr_node_id"].values,
+                open_prs_facts["repository_full_name"].values,
+                open_prs_facts["data"].values,
+        ):
+            node_id = pr_node_id.rstrip()
+            facts[node_id] = PullRequestFacts(
+                data=data,
+                repository_full_name=repository_full_name,
+                author=authors[node_id])
```
vmarkovtsev, line 160: In that case,
```suggestion
        facts = {
            facts[pr_node_id] = PullRequestFacts(
                data=data,
                repository_full_name=repository_full_name,
                author=authors[pr_node_id])
            for pr_node_id, repository_full_name, data in zip(
                open_prs_facts["pr_node_id"].str.rstrip().values,
                open_prs_facts["repository_full_name"].values,
                open_prs_facts["data"].values,
            )
        }        
```

If we didn't have `str.rstrip()` in Pandas, I would propose this:
```python
        facts = {
            facts[node_id := pr_node_id.rstrip()] = PullRequestFacts(
                data=data,
                repository_full_name=repository_full_name,
                author=authors[node_id])
            for pr_node_id, repository_full_name, data in zip(
                open_prs_facts["pr_node_id"].values,
                open_prs_facts["repository_full_name"].values,
                open_prs_facts["data"].values,
            )
        }  
```

vmarkovtsev (changes requested):

server/athenian/api/preloading/cache.py
```diff
@@ -452,5 +459,48 @@ def get_memory_cache_options() -> Dict[str, Dict[str, Dict[str, List[Instrumente
                     PrecomputedGitHubReleaseMatchTimespan.acc_id,
                 ],
             },
+            PCID.done_pr_facts.value: {
```
se7entyse7en, line 462: Interesting, I've just realized that I preloaded the table for done facts, but I actually need open facts, but test didn't fail... 🤔

se7entyse7en: Ah ok, it's because the tests are calling the classmethod directly which is always pointing to the non-preloaded version, and probably there's no test that calls the `/metrics/prs` endpoint that makes it call the preloaded implementation.

vmarkovtsev: We should cover this with tests then. This stuff is critical.

se7entyse7en: Totally agree, do you mind if I fix this and then I address all the tests separately? I guess this is not the only case.

vmarkovtsev: Sure.

vmarkovtsev (approved):

server/athenian/api/preloading/cache.py
se7entyse7en: Thanks!

PR #1429, opened 2021-05-13
Refactor precomputed prs module

se7entyse7en:
Splits the `miners.github.precomputed_prs` module that was massive into `{done,merged,open}_prs` + wrap loader functions into classes that will be needed for next preloadings.

Depends on [this](https://github.com/athenianco/athenian-api/pull/1428).


se7entyse7en:
I'll rework this on top of #1433

PR #1433, opened 2021-05-14
Preloading fix tests

se7entyse7en:
- Removes all the re-exported backward-compatible class methods as functions in favor of explicit call,
- All tests previously calling the re-exported functions are now calling the corresponding class method, preloaded or not, depending on the `with_preloading` fixture,
- Adds miners to CI preloading tests.

### DIFF
--- a/server/athenian/api/experiments/preloading/entries.py
+++ b/server/athenian/api/experiments/preloading/entries.py
@@ -1,10 +1,11 @@
-from datetime import datetime
+from datetime import datetime, timezone
 import functools
 import operator
 from typing import Collection, Dict, Iterable, List, Mapping, Optional, Sequence, Set, Tuple, Union
 
 import aiomcache
 import databases
+import numpy as np
 import pandas as pd
 from sqlalchemy.sql.elements import BinaryExpression
 
@@ -13,8 +14,7 @@ from athenian.api.controllers.features.github.unfresh_pull_request_metrics impor
     UnfreshPullRequestFactsFetcher
 from athenian.api.controllers.miners.filters import JIRAFilter, LabelFilter
 from athenian.api.controllers.miners.github.branches import BranchMiner
-from athenian.api.controllers.miners.github.precomputed_prs import \
-    build_days_range, OpenPRFactsLoader
+from athenian.api.controllers.miners.github.precomputed_prs import OpenPRFactsLoader
 from athenian.api.controllers.miners.github.pull_request import PullRequestMiner
 from athenian.api.controllers.miners.github.release_load import \
     match_groups_to_conditions, ReleaseLoader
@@ -74,8 +74,8 @@ class PreloadedReleaseLoader(ReleaseLoader):
         release_match_spans = cached_df.filter(mask)
         spans = {}
         for time_from, time_to, release_match, repository_full_name in zip(
-                release_match_spans["time_from"].values,
-                release_match_spans["time_to"].values,
+                release_match_spans["time_from"],
+                release_match_spans["time_to"],
                 release_match_spans["release_match"].values,
                 release_match_spans["repository_full_name"].values,
         ):
@@ -83,6 +83,7 @@ class PreloadedReleaseLoader(ReleaseLoader):
                 release_match = ReleaseMatch.tag
             else:
                 release_match = ReleaseMatch.branch
+
             times = time_from, time_to
             spans.setdefault(repository_full_name, {})[release_match] = times
 
@@ -100,7 +101,8 @@ class PreloadedReleaseLoader(ReleaseLoader):
             for cond in or_conditions
         ]
 
-        return functools.reduce(operator.or_, or_masks)
+        return (functools.reduce(operator.or_, or_masks) if or_masks
+                else np.ones(len(df), dtype=bool))
 
 
 class PreloadedReleaseToPullRequestMapper(ReleaseToPullRequestMapper):
@@ -135,12 +137,22 @@ class PreloadedOpenPRFactsLoader(OpenPRFactsLoader):
         df = cached_df.df
         mask = (
             (df["acc_id"] == account)
-            & df["pr_node_id"].isin(prs)
+            & df["pr_node_id"].isin(pr.encode() for pr in prs)
         )
 
         if exclude_inactive:
-            date_range = build_days_range(time_from, time_to)
-            mask &= df["activity_days"].apply(lambda v: bool(set(v).intersection(date_range)))
+            activity_days = np.concatenate(df["activity_days"])
+            activity_mask = np.full(len(df["activity_days"]), False)
+            activity_days_in_range = (
+                (time_from.replace(tzinfo=timezone.utc) <= activity_days) &
+                (activity_days < time_to.replace(tzinfo=timezone.utc))
+            )
+            activity_offsets = np.cumsum(df["activity_days"].apply(len).values)
+            indexes = np.searchsorted(activity_offsets, np.nonzero(activity_days_in_range)[0],
+                                      side="right")
+            activity_mask[indexes] = 1
+
+            mask &= activity_mask
 
         open_prs_facts = cached_df.filter(mask, columns=[
             "pr_node_id", "repository_full_name", "data"])
--- a/server/tests/controllers/miners/github/test_branches.py
+++ b/server/tests/controllers/miners/github/test_branches.py
@@ -3,14 +3,11 @@ from random import random
 import pytest
 from sqlalchemy import delete, insert, select, update
 
-from athenian.api.controllers.miners.github.branches import BranchMiner
 from athenian.api.defer import wait_deferred, with_defer
-from athenian.api.experiments.preloading.entries import PreloadedBranchMiner
 from athenian.api.models.metadata.github import Branch
 
 
-async def test_extract_branches_zero(mdb, with_preloading):
-    branch_miner = PreloadedBranchMiner if with_preloading else BranchMiner
+async def test_extract_branches_zero(mdb, branch_miner):
     branches, defaults = await branch_miner.extract_branches(
         ["src-d/gitbase"], (6366825,), mdb, None)
     assert branches.empty
@@ -22,8 +19,7 @@ async def test_extract_branches_zero(mdb, with_preloading):
         assert defaults == {"src-d/gitbase": "master"}
 
 
-async def test_extract_branches_trash(mdb, with_preloading):
-    branch_miner = PreloadedBranchMiner if with_preloading else BranchMiner
+async def test_extract_branches_trash(mdb, branch_miner):
     branches, defaults = await branch_miner.extract_branches(
         ["src-d/whatever"], (6366825,), mdb, None)
     assert branches.empty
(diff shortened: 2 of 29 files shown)


server/tests/controllers/miners/github/test_branches.py
```diff
@@ -3,49 +3,60 @@
...
-from athenian.api.controllers.miners.github.branches import extract_branches
+from athenian.api.controllers.miners.github.branches import BranchMiner
 from athenian.api.defer import wait_deferred, with_defer
+from athenian.api.experiments.preloading.entries import PreloadedBranchMiner
 from athenian.api.models.metadata.github import Branch
 
 
-async def test_extract_branches_zero(mdb):
-    branches, defaults = await extract_branches(["src-d/gitbase"], (6366825,), mdb, None)
+async def test_extract_branches_zero(mdb, with_preloading):
```
vmarkovtsev, line 12: Let's have a fixture.
```python
@pytest.fixture(scope="module")
def branch_miner(with_preloading):
    return PreloadedBranchMiner if with_preloading else BranchMiner
```

server/tests/controllers/miners/github/test_precomputed_prs.py
```diff
@@ -599,14 +600,15 @@ def with_repository_full_name(s):
 
 @with_defer
 async def test_discover_update_unreleased_prs_smoke(
-        mdb, pdb, rdb, default_branches, release_match_setting_tag):
+        mdb, pdb, rdb, default_branches, release_match_setting_tag, with_preloading):
+    release_loader = PreloadedReleaseLoader if with_preloading else ReleaseLoader
```
vmarkovtsev, line 604: Same here

vmarkovtsev (changes requested):

vmarkovtsev:
Sorry for the conflicts. I refactored `MemoryCachePreloader` a bit, so the code has become simpler. In this case it should be safe to overwrite with my changes in `conftest.py`.

se7entyse7en:
No worries, the conflicts were easy, I'm glad you simplified the preloader 👍.

vmarkovtsev (approved):

se7entyse7en:
@vmarkovtsev you may want to have a look at [this last commit] after your approval. Since now the tests were indeed running properly for the miners, some bugs came out 🎉

vmarkovtsev (changes requested):
Yeah, the stuff requires some changes now. Let me review.

server/athenian/api/experiments/preloading/entries.py
```diff
@@ -135,11 +143,11 @@ async def load_open_pull_request_facts_unfresh(cls,
...
         mask = (
             (df["acc_id"] == account)
-            & df["pr_node_id"].isin(prs)
+            & df["pr_node_id"].isin(pr.encode() for pr in prs)
         )
 
         if exclude_inactive:
-            date_range = build_days_range(time_from, time_to)
+            date_range = set(d.strftime("%Y-%m-%d") for d in build_days_range(time_from, time_to))
             mask &= df["activity_days"].apply(lambda v: bool(set(v).intersection(date_range)))
```
vmarkovtsev, line 151: This will work very slow. There is vectorized code using `np.searchsorted`, for example https://github.com/athenianco/athenian-api/blob/master/server/athenian/api/controllers/features/github/pull_request_metrics.py#L538

vmarkovtsev: Anyways, `df["activity_days"]` should not have the string type. It should be an object, a list of datetimes.

server/tests/controllers/miners/github/test_release.py
```diff
@@ -1403,15 +1412,18 @@ async def test_precomputed_releases_ambiguous(
...
         release_match_setting_branch, 1, (6366825,), mdb, pdb, rdb, None)
     await wait_deferred()
+    if with_preloading_enabled:
+        await pdb.cache.refresh()
     prels = await release_loader._fetch_precomputed_releases(
         {ReleaseMatch.tag: {".*": ["src-d/go-git"]},
          ReleaseMatch.branch: {"master": ["src-d/go-git"]}},
         time_from, time_to, 1, pdb)
-    prels = prels[releases_tag.columns]
+    cols = set(releases_tag.columns).intersection(prels.columns)
```
vmarkovtsev, line 1421: The columns should be exactly the same, hence the check. If they don't then there is a bug.

server/tests/controllers/miners/github/test_release.py
```diff
@@ -1375,25 +1379,30 @@ async def test_mine_releases_cache(
...
         ["src-d/go-git"], branches, default_branches, time_from, time_to,
         settings, 1, (6366825,), mdb, pdb, rdb, None)
     await wait_deferred()
+    if with_preloading_enabled:
+        await pdb.cache.refresh()
     prels = await release_loader._fetch_precomputed_releases(
         {ReleaseMatch(settings_index): {["master", ".*"][settings_index]: ["src-d/go-git"]}},
         time_from, time_to, 1, pdb)
-    prels = prels[releases.columns]
+    cols = set(releases.columns).intersection(prels.columns)
```
vmarkovtsev, line 1396: Same here.

vmarkovtsev (changes requested):

server/athenian/api/experiments/preloading/entries.py
se7entyse7en: Indeed, I just realized with the tests failing for PG in CI, it should be like this actually:
```
        if exclude_inactive:
            date_range = build_days_range(time_from, time_to)
            if pdb.url.dialect == "sqlite":
                date_range = set(d.strftime("%Y-%m-%d") for d in date_range)
            mask &= df["activity_days"].apply(lambda v: bool(set(v).intersection(date_range)))
```

I'll have a look at your vectorized implementation.

server/tests/controllers/miners/github/test_release.py
se7entyse7en: When I was testing the releases table everything worked so when building the options for the `MemoryCache` I just skipped all the columns that were unused. The reasons being to save memory ofc, especially given that many of the columns skipped are identifiers. Or maybe I just missed where they're used? At least for the unfresh path it didn't break.
What I mean is that we can easily just preload all the columns, but it will be costly. Does it make sense to change this check only when preloading is enabled then?

server/athenian/api/experiments/preloading/entries.py
```diff
@@ -83,7 +89,8 @@ async def fetch_precomputed_release_match_spans(
                 release_match = ReleaseMatch.tag
             else:
                 release_match = ReleaseMatch.branch
-            times = time_from, time_to
+
+            times = _parse(time_from), _parse(time_to)
```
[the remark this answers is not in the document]
se7entyse7en: There's a test that explicitly asserts the equality on `datetime`, but since we're `zip`-ping on the `values` the type is a numpy datetime.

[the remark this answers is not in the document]
vmarkovtsev: I see, then it's another time when we have to remove `values`. It damages the performance a little bit, but not as much as `groupby` or `apply` hehe.

server/athenian/api/experiments/preloading/entries.py
vmarkovtsev: I assume `df["activity_days"]` with SQLite should be coerced to the same format as Postgres inside the preloader.

server/tests/controllers/miners/github/test_release.py
vmarkovtsev: As I remember, `_fetch_precomputed_releases` should return quite specific columns that much the "live-loaded" alternative. The code really relies on all of them. What are the exact unused columns?

se7entyse7en: The only columns that are being preloaded are the followings:
```
                    PrecomputedRelease.id,
                    PrecomputedRelease.release_match,
                    PrecomputedRelease.repository_full_name,
                    PrecomputedRelease.published_at,
                    PrecomputedRelease.acc_id,
```

But probably I know why: https://github.com/athenianco/athenian-api/blob/master/server/athenian/api/controllers/features/github/unfresh_pull_request_metrics.py#L93

Here `releases` should be the processed data frame that comes from `_fetch_precomputed_releases`. But inside `load_releases` the columns that I mentioned above are enough, and nothing else is needed. What I'm saying is based on how the `releases` variable [here](https://github.com/athenianco/athenian-api/blob/master/server/athenian/api/controllers/miners/github/release_load.py#L103) is used inside the method.

server/athenian/api/experiments/preloading/entries.py
se7entyse7en: Ah, good point! coercing in the preloader is the best solution, thanks!

se7entyse7en: done

server/athenian/api/experiments/preloading/entries.py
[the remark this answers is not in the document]
se7entyse7en: done

server/tests/controllers/miners/github/test_release.py
se7entyse7en: Changed conditionally on `with_preloading_enabled`.

server/tests/controllers/miners/github/test_release.py
se7entyse7en: Changed conditionally on `with_preloading_enabled`.

server/tests/controllers/miners/github/test_release.py
vmarkovtsev: This is the thing. There are code branches that require the rest of the columns. For example, filtering PRs breaks here: https://github.com/athenianco/athenian-api/blob/master/server/athenian/api/controllers/features/github/pull_request_filter.py#L256
Or filtering JIRA and `/get/pull_requests` break here: https://github.com/athenianco/athenian-api/blob/master/server/athenian/api/controllers/features/github/pull_request_filter.py#L787
The tests do not break because the property values are not checked, probably...

vmarkovtsev: So we should preload everything, really.

vmarkovtsev: I understood: we are only running "fat" tests so some controller tests do not execute with preloading.

vmarkovtsev: Let's do the following then. Explicitly assert that the rest of the columns do not exist under `with-preloading`. This way, when we extend to the other endpoints, these tests will fail and make us update them. Add a comment that explains why we assert and how to update - to remove the `if` whatsoever.

se7entyse7en: It makes a lot of sense, thanks for checking this out 🙇

se7entyse7en:
Make tests for preloaded precomputed releases more specific + comment: https://github.com/athenianco/athenian-api/pull/1433/commits/2e2e6382b6206a0c781baa838d3faa377b7f42f7

se7entyse7en:
> sqlite3.OperationalError: database is locked

flaky?

vmarkovtsev (approved):

PR #1447, opened 2021-05-17
[DEV-2289] Cancel previous workflow runs

vmarkovtsev:


vmarkovtsev:
Works like a charm @se7entyse7en !

PR #1448, opened 2021-05-18
[DEV-2115] Add the spec of /filter/code_checks

vmarkovtsev:


server/athenian/api/models/web/filter_code_checks_request.py
```diff
@@ -0,0 +1,86 @@
...
+            raise ValueError("Invalid value for `in_`, must not be `None`")
+
+        self._in_ = in_
+
+    @property
+    def triggered_by(self) -> List[str]:
+        """Gets the triggered_by of this FilterCodeChecksRequest.
+
+        Check runs must be triggered by commits pushed by these people. When it is impossible
+        to determine who pushed, e.g. in legacy API based checks, they are committers.
```
warenlg, line 65: Ok, so no issue here on filtering by users, right?

server/athenian/api/models/web/filtered_code_check_run.py
```diff
@@ -0,0 +1,188 @@
...
+
+        self._title = title
+
+    @property
+    def repository(self) -> str:
+        """Gets the repository of this FilteredCodeCheckRun.
+
+        Repository name which uniquely identifies any repository in any service provider.
+        The format matches the repository URL without the protocol part. No \".git\" should be
+        appended. We support a special syntax for repository sets: \"{reposet id}\" adds all
```
warenlg, line 83: Are we safe wrt GHE pointers? We expect webapp to reconstruct the url of the repo.

server/athenian/api/models/web/filtered_code_check_run.py
```diff
@@ -0,0 +1,188 @@
...
+
+        :param repository: The repository of this FilteredCodeCheckRun.
+        """
+        if repository is None:
+            raise ValueError("Invalid value for `repository`, must not be `None`")
+
+        self._repository = repository
+
+    @property
+    def last_execution(self) -> datetime:
```
warenlg, line 107: Could we add the url pointing to this last execution?

server/athenian/api/models/web/filtered_code_check_runs.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        self._timeline = timeline
+        self._items = items
+
+    @property
+    def timeline(self) -> List[date]:
+        """Gets the timeline of this FilteredCodeCheckRuns.
+
+        Sequence of dates from `date_from` till `date_to`. We choose such an interval that
+        the number of items is approximately 50.
```
warenlg, line 38: I don't understand what is this? What is the difference with granularities?

server/athenian/api/models/web/filter_code_checks_request.py
```diff
@@ -0,0 +1,86 @@
...
+    @triggered_by.setter
+    def triggered_by(self, triggered_by: List[str]):
+        """Sets the triggered_by of this FilterCodeChecksRequest.
+
+        Check runs must be triggered by commits pushed by these people. When it is impossible
+        to determine who pushed, e.g. in legacy API based checks, they are committers.
+
+        :param triggered_by: The triggered_by of this FilterCodeChecksRequest.
+        """
+        self._triggered_by = triggered_by
```
warenlg, line 80: Alright, so we allow passing an empty list or None here.

server/athenian/api/openapi/openapi.yaml
```diff
@@ -7841,3 +7890,375 @@ components:
...
+          repository: github.com/src-d/athenian-api
+          last_execution: 2021-05-18
+          size_groups:
+          - 3
+          - 5
+          - 8
+          total_stats:
+            count: 100
+            successes: 50
+            mean_execution_time: 200s
```
warenlg, line 7958: We'll need quantiles applied to this one.

server/athenian/api/models/web/filtered_code_check_run.py
```diff
@@ -0,0 +1,188 @@
...
+
+        :param total_stats: The total_stats of this FilteredCodeCheckRun.
+        """
+        if total_stats is None:
+            raise ValueError("Invalid value for `total_stats`, must not be `None`")
+
+        self._total_stats = total_stats
+
+    @property
+    def prs_stats(self) -> CodeCheckRunStatistics:
```
warenlg, line 149: Ah ok so prs_stats is the stats for checks run on_pull_requests, and total_stats for checks run on_pull_requests and on_push, right?

server/athenian/api/models/web/code_check_run_statistics.py
```diff
@@ -0,0 +1,321 @@
...
+
+from athenian.api.models.web.base_model_ import Model
+
+
+class CodeCheckRunStatistics(Model):
+    """Gathered statistics about a group of check runs."""
+
+    openapi_types = {
+        "count": int,
+        "successes": int,
```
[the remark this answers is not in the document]
vmarkovtsev: Yes, because I'd rather send the raw counts and let the webapp do the division. In case we'll need the counts in the future.

server/athenian/api/models/web/code_check_run_statistics.py
```diff
@@ -0,0 +1,321 @@
...
+    def count(self, count: int):
+        """Sets the count of this CodeCheckRunStatistics.
+
+        Number of executions with respect to `date_from` and `date_to`.
+
+        :param count: The count of this CodeCheckRunStatistics.
+        """
+        if count is None:
+            raise ValueError("Invalid value for `count`, must not be `None`")
+        if count < 1:
```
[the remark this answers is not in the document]
vmarkovtsev: This is just by definition: we cannot return statistics about 0 check runs.

server/athenian/api/models/web/filter_code_checks_request.py
vmarkovtsev: No problem at all.

server/athenian/api/models/web/filtered_code_check_run.py
vmarkovtsev: Yes, we should be fine here.

server/athenian/api/models/web/filtered_code_check_run.py
vmarkovtsev: Sure.

server/athenian/api/models/web/filtered_code_check_runs.py
vmarkovtsev: Consider this the autogranularity that lets us plot everything on X axis. Do you think that the webapp should select how many timestamps? We could agree on a reasonable default and reduce the complexity.

server/athenian/api/models/web/filter_code_checks_request.py
vmarkovtsev: Here - yes.

server/athenian/api/models/web/filtered_code_check_run.py
vmarkovtsev: Correct.

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: Can I hardcode 0.95 maybe? How about the timelines?

warenlg: Hardcoding 0.95 would work for a pretty long time I believe. Yes please do.
For timelines it should work as the regular timelines in the product ie. apply quantization overall then expand the values through the timeline. And not apply quantization separately to each data point eg. each day

server/athenian/api/models/web/filtered_code_check_runs.py
warenlg: I see the goal here is to save us from a massive implementation of the granularities logic, right?
Let's discuss it tomorrow but indeed letting the webapp choosing the number of timestamps might let us approx in a cheap way, the behavior of granularities eg. if date_to - date_from = delta > 28 days the webapp could request delta / 7 timestamps, etc

server/athenian/api/openapi/openapi.yaml
vmarkovtsev: I realized that hardcoding 0.95 is almost equivalent to exposing `quantiles`. Added `quantiles`.

warenlg (approved):

PR #1452, opened 2021-05-20
Ensure that PreloadedPullRequestMiner is used when testing with preloading enabled

se7entyse7en:
I realized I missed fixing the tests for `PullRequestMiner`.

### DIFF
--- a/server/athenian/api/preloading/cache.py
+++ b/server/athenian/api/preloading/cache.py
@@ -470,18 +470,24 @@ def get_memory_cache_options() -> Dict[str, Dict[str, Dict[str, List[Instrumente
                 },
                 "cols": [
                     PrecomputedRelease.id,
+                    PrecomputedRelease.sha,
                     PrecomputedRelease.release_match,
                     PrecomputedRelease.repository_full_name,
                     PrecomputedRelease.published_at,
                     PrecomputedRelease.acc_id,
+                    PrecomputedRelease.author,
+                    PrecomputedRelease.url,
                 ],
                 "categorical_cols": [
                     PrecomputedRelease.acc_id,
                     PrecomputedRelease.repository_full_name,
                     PrecomputedRelease.release_match,
+                    PrecomputedRelease.author,
                 ],
                 "identifier_cols": [
                     PrecomputedRelease.id,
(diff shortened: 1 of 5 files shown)


vmarkovtsev (approved):

vmarkovtsev:
@se7entyse7en conflicts

se7entyse7en:
@vmarkovtsev tests were passing already, but if you could PTAL at [this](https://github.com/athenianco/athenian-api/pull/1452/commits/7fff5199f65978416895d2e9b3594b3810d9c0a7). Basically, what is happening is that in that flaky test, when it's not "ready", the commit id column has `np.nan`s and breaks the string formatting with `TypeError`.

server/athenian/api/controllers/miners/github/commit.py
```diff
@@ -200,6 +200,7 @@ async def fetch_repository_commits(repos: Dict[str, DAG],
     missed_counter = 0
     repo_heads = {}
     sha_col, id_col, dt_col, repo_col = columns
+    branches = branches[branches[id_col].notnull()]
```
vmarkovtsev, line 203: Well, this is impossible because of https://github.com/athenianco/athenian-api/blob/7fff5199f65978416895d2e9b3594b3810d9c0a7/server/athenian/api/controllers/miners/github/branches.py#L127 I know that releases may pretend as branches, too, but this is still a huge violation of the input constraints. In other words, if this happens and some branches appear without commits, we must assert and panic instead of silently filtering them away.

vmarkovtsev (changes requested):

vmarkovtsev:
What's the exact test that fails sometimes?

se7entyse7en:
This one: https://github.com/athenianco/athenian-api/blob/master/server/tests/controllers/miners/github/test_pull_request.py#L865

Maybe you know how to better address this.

vmarkovtsev:
thanks, I'll take a look :heavy_check_mark:

vmarkovtsev:
Taking a look.

vmarkovtsev:
I think I fixed the test yesterday, once and for all.

se7entyse7en:
Rebased and removed the hack 👍 thanks!

PR #1449, opened 2021-05-19
Preload prs merged facts

se7entyse7en:
Closes [DEV-2312](https://athenianco.atlassian.net/browse/DEV-2312).

Depends on #1434.

PR #1454, opened 2021-05-20
Preload prs done facts

se7entyse7en:
Closes [DEV-2307](https://athenianco.atlassian.net/browse/DEV-2307).

Depends on #1449.

### DIFF
--- a/server/athenian/api/experiments/preloading/entries.py
+++ b/server/athenian/api/experiments/preloading/entries.py
@@ -9,6 +9,7 @@ import aiomcache
 import databases
 import numpy as np
 import pandas as pd
+from sqlalchemy.ext.declarative.api import Base as SQLABase
 from sqlalchemy.sql.elements import BinaryExpression
 
 from athenian.api import metadata
--- a/server/athenian/api/experiments/preloading/entries.py
+++ b/server/athenian/api/experiments/preloading/entries.py
@@ -238,15 +238,8 @@ class PreloadedDonePRFactsLoader(DonePRFactsLoader):
         with sentry_sdk.start_span(op="_load_precomputed_done_filters/triage"):
             result = {}
             ambiguous = {ReleaseMatch.tag.name: {}, ReleaseMatch.branch.name: {}}
-
-            # TODO: this `to_dict` call can probably be removed. The reason why it's here
-            # is because `dump` values requires to be `dict`, but iterating over the numpy
-            # arrays and doing the following:
-            #
-            #     dump[pr_node_id] = dict(done_pr_facts.iloc[i])
-            #
-            # is super slow because of the `dict` conversion.
-            i = 0
+            # we cannot iterate with zip(column1, column2, ...) because the rest of the code
+            # expects dicts and to_dict() is faster than making individual dicts.
             for row in done_pr_facts.to_dict(orient="records"):
                 dump = triage_by_release_match(
                     row[model.repository_full_name.key],
@@ -259,7 +252,6 @@ class PreloadedDonePRFactsLoader(DonePRFactsLoader):
                 if dump is None:
                     continue
                 dump[row[model.pr_node_id.key]] = row
-                i += 1
 
         return cls._post_process_ambiguous_done_prs(result, ambiguous)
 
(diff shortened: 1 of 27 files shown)


server/athenian/api/experiments/preloading/entries.py
```diff
@@ -116,6 +119,148 @@ class PreloadedReleaseToPullRequestMapper(ReleaseToPullRequestMapper):
...
+                    row["release_match"],
+                    release_settings,
+                    default_branches,
+                    result,
+                    ambiguous,
+                )
+                if dump is None:
+                    continue
+                dump[row["pr_node_id"]] = row
+                i += 1
```
vmarkovtsev, line 229:
```suggestion
            for repository_full_name, release_match, pr_node_id in zip(done_pr_facts[DonePullRequestFacts.repository_full_name.key].values, done_pr_facts[DonePullRequestFacts.release_match.key].values, done_pr_facts[DonePullRequestFacts.pr_node_id].values):
```
This is both the fastest and avoids hardcoding column names.

server/athenian/api/experiments/preloading/entries.py
```diff
@@ -116,6 +119,148 @@ class PreloadedReleaseToPullRequestMapper(ReleaseToPullRequestMapper):
...
+            match_groups[ReleaseMatch.rejected] = match_groups[
+                ReleaseMatch.force_push_drop
+            ] = {"": repos}
+            if event_repos:
+                match_groups[ReleaseMatch.event] = {"": event_repos}
+
+            mask &= cls._match_groups_to_mask(df, match_groups)
+
+        done_pr_facts = cached_df.filter(
+            mask, columns=["pr_node_id", "repository_full_name", "release_match"] + [
```
vmarkovtsev, line 201:
```suggestion
            mask, columns=[DonePullRequestFacts.pr_node_id.key, "repository_full_name", "release_match"] + [
```
There should be zero hardcoded strings like these in the code because if need a refactoring with a column you can click "Find Usages" in PyCharm and simply go through the list.

server/athenian/api/experiments/preloading/entries.py
```diff
@@ -116,6 +119,148 @@ class PreloadedReleaseToPullRequestMapper(ReleaseToPullRequestMapper):
...
+                else np.ones(len(df), dtype=bool))
+
+    @classmethod
+    def _match_groups_to_mask(
+            cls,
+            df: pd.DataFrame,
+            match_groups: Dict[ReleaseMatch, Dict[str, Iterable[str]]]) -> pd.Series:
+        or_conditions, _ = match_groups_to_conditions(match_groups)
+        or_masks = [
+            (df["release_match"] == cond["release_match"]) &
```
vmarkovtsev, line 255:
```suggestion
            (df[DonePullRequestFacts.release_match.key] == cond["release_match"]) &
```

vmarkovtsev (changes requested):

server/athenian/api/experiments/preloading/entries.py
se7entyse7en: I'll do a separate PR before this and rebase where I fix all the names only.

server/athenian/api/experiments/preloading/entries.py
se7entyse7en: I initially did this, but since `dump` values should be dictionaries of the whole row, it seemed that building the `dict` in the loop was taking a lot of time, that's why I came up with `to_dict(orient="records")`.

vmarkovtsev: Damn. OK, please leave a comment that explains this.

se7entyse7en: It's here: https://github.com/athenianco/athenian-api/pull/1454/files#diff-0ad06f21667ae6b7fa99290cd35ebd84fdcab18d79d7711e6c505449bb328e9fR209 😄

se7entyse7en:
Rebased on top of #1468.

server/athenian/api/experiments/preloading/entries.py
vmarkovtsev: Changed the comment.

vmarkovtsev:
The reason why this fails is because our preloading engine does not support Postgres' HSTORE.

vmarkovtsev (approved):

PR #1469, opened 2021-05-26
[DEV-2383] Add "only_default_branch" option to the commits filter

vmarkovtsev:


vmarkovtsev:
@se7entyse7en `test_pr_miner_unreleased_facts` is not passing now.
This happened when we merged #1452: https://github.com/athenianco/athenian-api/runs/2675960625
I have to mark the test as expected failure with preloading.

PR #1509, opened 2021-06-08
[DEV-2451] Check if None appears in join_dags()

vmarkovtsev:


vmarkovtsev:
Merging as-is due to `pull_request_target`, hoping for the best.

PR #1530, opened 2021-06-16
[DEV-2239] Beat shit out of aiocron

vmarkovtsev:


vmarkovtsev:
@se7entyse7en It would be great to add a few unit tests for aiocron. Maybe launching something every second.

PR #1532, opened 2021-06-17
Preloading refresh

se7entyse7en:
Closes [DEV-2239](https://athenianco.atlassian.net/browse/DEV-2239).

### DIFF
--- a/server/athenian/api/__main__.py
+++ b/server/athenian/api/__main__.py
@@ -105,6 +105,8 @@ def parse_args() -> argparse.Namespace:
                              "for example, 0.0.0.0:11211")
     parser.add_argument("--preload-dataframes", required=False, action="store_true",
                         help="Whether to preload DB tables in memory and refresh periodically.")
+    parser.add_argument("--preload-refresh-frequency", type=int, default=60, required=False,
+                        help="Frequency at which to refresh the preloaded tables in minutes.")
     parser.add_argument("--ui", action="store_true", help="Enable the REST UI.")
     parser.add_argument("--no-google-kms", action="store_true",
                         help="Skip Google Key Management Service initialization. Personal Access "
@@ -298,11 +300,13 @@ def create_segment() -> Optional[SegmentClient]:
 PRELOADER_VAR_NAME = "mc_preloader"
 
 
-def setup_preloading(app: AthenianApp, log: logging.Logger) -> None:
+def setup_preloading(app: AthenianApp, preload_refresh_frequency: int,
+                     log: logging.Logger) -> None:
     """Initialize the memory cache and schedule loading the DB tables."""
     log.info("Preloading DB tables to memory is enabled")
     app.app[PRELOADER_VAR_NAME] = mc_preloader = MemoryCachePreloader(
-        app.app[PROMETHEUS_REGISTRY_VAR_NAME])
+        preload_refresh_frequency,
+        prometheus_registry=app.app[PROMETHEUS_REGISTRY_VAR_NAME])
     app.on_dbs_connected(mc_preloader.preload)
 
 
@@ -334,7 +338,7 @@ def main() -> Optional[AthenianApp]:
         segment=create_segment(),
     )
     if args.preload_dataframes:
-        setup_preloading(app, log)
+        setup_preloading(app, args.preload_refresh_frequency, log)
     app.run(host=args.host, port=args.port, print=lambda s: log.info("\n" + s))
     return app
 
--- a/server/athenian/api/__main__.py
+++ b/server/athenian/api/__main__.py
@@ -309,6 +309,11 @@ def setup_preloading(app: AthenianApp, preload_refresh_frequency: int,
         prometheus_registry=app.app[PROMETHEUS_REGISTRY_VAR_NAME])
     app.on_dbs_connected(mc_preloader.preload)
 
+    async def shutdown(self, app: Optional[aiohttp.web.Application] = None) -> None:
+        await mc_preloader.stop()
+
+    app.app.on_shutdown.insert(0, shutdown)
+
 
 def main() -> Optional[AthenianApp]:
     """Server's entry point."""
--- a/server/athenian/api/preloading/cache.py
+++ b/server/athenian/api/preloading/cache.py
@@ -12,6 +12,7 @@ import sqlalchemy as sa
 from sqlalchemy import select
 from sqlalchemy.orm.attributes import InstrumentedAttribute
 
+from athenian.api import aiocron
 from athenian.api import metadata
 from athenian.api.async_utils import gather, read_sql_query
 from athenian.api.models.metadata.github import Branch, NodePullRequestJiraIssues, \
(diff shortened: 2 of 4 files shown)


server/athenian/api/__main__.py
```diff
@@ -105,6 +105,8 @@ def level_from_msg(msg: str) -> Optional[str]:
                              "for example, 0.0.0.0:11211")
     parser.add_argument("--preload-dataframes", required=False, action="store_true",
                         help="Whether to preload DB tables in memory and refresh periodically.")
+    parser.add_argument("--preload-refresh-frequency", type=int, default=60, required=False,
+                        help="Frequency at which to refresh the preloaded tables.")
```
vmarkovtsev, line 109: Can you please add the units to the help message: seconds, minutes, etc.

server/athenian/api/preloading/cache.py
```diff
@@ -333,9 +334,15 @@ class MemoryCachePreloader:
...
 
     def __init__(self,
+                 refresh_frequency_minutes: int,
+                 max_early_refresh_minutes: Optional[int] = None,
                  prometheus_registry: Optional[prometheus_client.CollectorRegistry] = None,
                  debug_memory: Optional[bool] = None):
         """Initialize a `MemoryCachePreloader`."""
+        self._refresh_frequency_minutes = refresh_frequency_minutes
+        self._max_early_refresh_minutes = (
+            max_early_refresh_minutes or int(refresh_frequency_minutes / 5)
```
vmarkovtsev, line 344:
```suggestion
            max_early_refresh_minutes or refresh_frequency_minutes // 5
```

server/athenian/api/preloading/cache.py
```diff
@@ -366,6 +373,20 @@ async def preload(self, **dbs: databases.Database) -> None:
...
 
+        self._log.debug("Scheduling refresh for preloaded DB tables in memory")
+        for db_name in get_memory_cache_options().keys():
+            job = aiocron.crontab(
+                f"*/{self._refresh_frequency_minutes} * * * *",
+                func=_refresh, args=(dbs[db_name], ), start=False,
+                max_early_expiration_seconds=self._max_early_refresh_minutes)
+            job.start()
+
+        self._log.info("Finished scheduling refresh for preloaded DB tables in memory")
```
vmarkovtsev, line 384:
```suggestion
        self._log.info("Scheduled refresh for preloaded DB tables in memory each %d min", self._refresh_frequency_minutes)
```

server/athenian/api/preloading/cache.py
```diff
@@ -366,6 +373,20 @@ async def preload(self, **dbs: databases.Database) -> None:
         await gather(*tasks)
         self._log.info("Finished preloading DB tables in memory")
 
+        self._log.debug("Scheduling refresh for preloaded DB tables in memory")
+        for db_name in get_memory_cache_options().keys():
+            job = aiocron.crontab(
+                f"*/{self._refresh_frequency_minutes} * * * *",
+                func=_refresh, args=(dbs[db_name], ), start=False,
+                max_early_expiration_seconds=self._max_early_refresh_minutes)
+            job.start()
```
vmarkovtsev, line 382: 1. What if we set `start=True`, we won't have to call `job.start()`.
2. We should save to `self._job` and `close()` it in our higher-level `close()`.

server/athenian/api/preloading/cache.py
```diff
@@ -366,6 +373,20 @@ async def preload(self, **dbs: databases.Database) -> None:
         await gather(*tasks)
         self._log.info("Finished preloading DB tables in memory")
 
+        self._log.debug("Scheduling refresh for preloaded DB tables in memory")
+        for db_name in get_memory_cache_options().keys():
```
vmarkovtsev, line 377: This will work, too - Go style:
```suggestion
        for db_name in get_memory_cache_options():
```

server/tests/conftest.py
```diff
@@ -311,7 +311,7 @@ async def with_preloading(sdb, mdb, mdb_rw, pdb, rdb, with_preloading_enabled):
     if not with_preloading_enabled:
         return False
 
-    mc_preloader = MemoryCachePreloader()
+    mc_preloader = MemoryCachePreloader(60)
```
vmarkovtsev, line 314: Yeah, we definitely need unit tests of the refresh. Maybe in a separate PR, as you prefer.

vmarkovtsev (changes requested):

server/tests/conftest.py
se7entyse7en: Indeed, I created a separate ticket: https://athenianco.atlassian.net/browse/DEV-2499

server/athenian/api/__main__.py
```diff
@@ -298,12 +300,15 @@ def create_segment() -> Optional[SegmentClient]:
...
+def setup_preloading(app: AthenianApp, preload_refresh_frequency: int,
+                     log: logging.Logger) -> None:
     """Initialize the memory cache and schedule loading the DB tables."""
     log.info("Preloading DB tables to memory is enabled")
     app.app[PRELOADER_VAR_NAME] = mc_preloader = MemoryCachePreloader(
-        app.app[PROMETHEUS_REGISTRY_VAR_NAME])
+        preload_refresh_frequency,
+        prometheus_registry=app.app[PROMETHEUS_REGISTRY_VAR_NAME])
     app.on_dbs_connected(mc_preloader.preload)
+    app.on_shutdown(mc_preloader.stop)
```
vmarkovtsev, line 311:
```suggestion
    app.app.on_shutdown.append(mc_preloader.stop)
```
and we don't need our custom code in `connexion.py` :+1:

vmarkovtsev (changes requested):

server/athenian/api/__main__.py
se7entyse7en: The reason why I did it is to have all the shutdown operations centralized in the current `_shutdown` method. In this case, for example, these callbacks will be called after the check of `self._shutting_down` and after calling `self._set_unready()`.

vmarkovtsev: Too much code for a single use-case. Let's iterate over `self.app.on_shutdown` inside our `shutdown()` then and remove `self.app.on_shutdown.append(self.shutdown)`.

se7entyse7en: And how our `shutdown()` is triggered then if we remove `self.app.on_shutdown.append(self.shutdown)`?

vmarkovtsev: The thing is, we actually disable the built-in shutdown mechanism, the signal handlers, etc. because they do not work good enough: https://github.com/athenianco/athenian-api/blob/master/server/athenian/api/connexion.py#L648
I think I inserted `self.app.on_shutdown.append(self.shutdown)` during a brain freeze.

se7entyse7en: I think there's something that I'm clearly missing. Even if you remove and add the signal handlers, I don't see any other place for calling `self.shutdown` other than being appended into `self.app.on_shutdown`.

vmarkovtsev: In short, we intercept SIGTERM and SIGINT and do not let aiohttp do the standard job. We further re-raise GracefulExit if needed. It works quite twisted, but super reliable.

vmarkovtsev: Now I see. Yes, GracefulExit in turn triggers the pass over `app.on_shutdown`, so we shouldn't remove it. OK! Then the only thing to do is `self.app.on_shutdown.insert(0, callback)` and the rest will happen automatically.

What I initially thought is that we overrode the `shutdown()` that's called before the signal processing. No, it's the other way round.

se7entyse7en: done

vmarkovtsev (approved):

PR #1548, opened 2021-06-21
No preloading if not ready

se7entyse7en:
Closes [DEV-2513](https://athenianco.atlassian.net/browse/DEV-2513).

Depends on [this](https://github.com/athenianco/athenian-api/pull/1536).

### DIFF
--- a/server/athenian/api/controllers/calculator_selector.py
+++ b/server/athenian/api/controllers/calculator_selector.py
@@ -4,7 +4,8 @@ import aiomcache
 from sqlalchemy import and_, select
 
 from athenian.api.async_utils import gather
-from athenian.api.controllers.features.entries import make_calculator, MetricEntriesCalculator
+from athenian.api.controllers.features.entries import CalculatorNotReadyException, \
+    make_calculator, MetricEntriesCalculator
 from athenian.api.db import DatabaseLike
 from athenian.api.models.state.models import AccountFeature, Feature, FeatureComponent
 from athenian.api.tracing import sentry_span
@@ -46,14 +47,23 @@ async def _get_calculator_for_account(
     base_module: Optional[str] = "athenian.api.experiments",
 ) -> MetricEntriesCalculator:
     def _make_calculator(variation=None):
+        assert service == "github", "we don't support others"
+
+        try:
+            calculator = make_calculator(
+                account_id, meta_ids, mdb, pdb, rdb, cache,
+                variation=variation, base_module=base_module,
+            )
+        except CalculatorNotReadyException:
+            variation = "default"
+            calculator = make_calculator(
+                account_id, meta_ids, mdb, pdb, rdb, cache,
+                variation=variation, base_module=base_module,
+            )
+
         if instrument is not None:
             instrument[service] = variation or "default"
-
-        assert service == "github", "we don't support others"
-        return make_calculator(
-            account_id, meta_ids, mdb, pdb, rdb, cache,
-            variation=variation, base_module=base_module,
-        )
+        return calculator
 
     feature_name_prefix = METRIC_ENTRIES_VARIATIONS_PREFIX.get(service)
     if not feature_name_prefix:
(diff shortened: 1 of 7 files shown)


server/tests/experiments/test_preloading.py
```diff
@@ -1,9 +1,14 @@
+import pytest
+
 from athenian.api.controllers.features.entries import make_calculator
 from athenian.api.experiments.preloading import MetricEntriesCalculator
 
 
+# TODO: this fails because precomputed data are not loaded for unittest
```
se7entyse7en, line 7: Tbh I'd rather not do this in this PR, I don't even know how much work it could require.

server/athenian/api/controllers/features/entries.py
```diff
@@ -719,7 +723,12 @@ def make_calculator(
     """Get the metrics calculator according to the account's features."""
 
     def build_calculator(cls=MetricEntriesCalculator):
-        return cls(account, meta_ids, mdb, pdb, rdb, cache)
+        calculator = cls(account, meta_ids, mdb, pdb, rdb, cache)
+        if not calculator.is_ready_for(account, meta_ids):
+            log.error("Cannot make calculator for variation '%s'", variation)
+            raise RuntimeError(f"Calculator not ready for account '{account}' and "
```
vmarkovtsev, line 729: Let's have our very own `CalculatorNotReady` exception.

server/athenian/api/controllers/calculator_selector.py
```diff
@@ -46,14 +46,23 @@ async def _get_calculator_for_account(
...
 ) -> MetricEntriesCalculator:
     def _make_calculator(variation=None):
+        assert service == "github", "we don't support others"
+
+        try:
+            calculator = make_calculator(
+                account_id, meta_ids, mdb, pdb, rdb, cache,
+                variation=variation, base_module=base_module,
+            )
+        except Exception:
```
vmarkovtsev, line 56:
```suggestion
        except CalculatorNotReady:
```

server/tests/experiments/test_preloading.py
vmarkovtsev: What's happening here?

vmarkovtsev (changes requested):
So this is to avoid using the preloaded data while it is being refreshed, correct?

server/tests/experiments/test_preloading.py
se7entyse7en: The problem is that the `is_ready_for` checks for the data being indeed preloaded to avoid querying an empty df. This check is implemented by checking whether there's the corresponding shard loaded. The problem is that in the unittests we only load `metadata`, but not `precomputed`, so this check always fails. We need to also load for tests the `precomputed` data.

se7entyse7en: In production those data would be ready since we don't open access until the heater runs.

vmarkovtsev: We've got the `app` fixture. We can extend it with `is_preloading_enabled` and await the preloader if it is True. Should be easy.

se7entyse7en: I think I'm missing something, the problem is not the preloader being ready, but that we don't have any data in the `precomputed` database. So AFAIU to solve this we should either create the fixture and load those the same way we do for `metadata` or run the heater on top of the loaded `metadata` data.

vmarkovtsev: Got it 👍

server/athenian/api/controllers/features/entries.py
se7entyse7en: done

server/athenian/api/controllers/calculator_selector.py
se7entyse7en: done

vmarkovtsev (approved):

PR #1580, opened 2021-07-01
[DEV-2557] Disable release missing_high mechanism when preloading

se7entyse7en:
Closes [DEV-2557](https://athenianco.atlassian.net/browse/DEV-2557).

### DIFF
--- a/server/athenian/api/controllers/miners/github/release_load.py
+++ b/server/athenian/api/controllers/miners/github/release_load.py
@@ -4,7 +4,7 @@ from itertools import chain
 import logging
 import pickle
 import re
-from typing import Dict, Iterable, List, Optional, Sequence, Tuple, Union
+from typing import Dict, Iterable, List, NamedTuple, Optional, Sequence, Set, Tuple, Union
 
 import aiomcache
 import asyncpg
@@ -41,6 +41,15 @@ from athenian.api.tracing import sentry_span
 
 tag_by_branch_probe_lookaround = timedelta(weeks=4)
 unfresh_releases_threshold = 50
+
+
+class MissingReleases(NamedTuple):
+    """Missing releases."""
+
+    missing_high: List[Tuple[datetime, Tuple[str, ReleaseMatch]]]
+    missing_low: List[Tuple[datetime, Tuple[str, ReleaseMatch]]]
+    missing_all: List[Tuple[str, ReleaseMatch]]
+    ambiguous_branches_scanned: Set[str]
 
 
 class ReleaseLoader:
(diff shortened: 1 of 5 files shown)


server/athenian/api/controllers/miners/github/release_load.py
```diff
@@ -41,6 +41,9 @@
 tag_by_branch_probe_lookaround = timedelta(weeks=4)
 unfresh_releases_threshold = 50
 
+MissingReleases = namedtuple("MissingReleases", ["missing_high", "missing_low", "missing_all",
```
vmarkovtsev, line 44: Can you please use the typed `typing.NamedTuple`

server/athenian/api/experiments/preloading/entries.py
```diff
@@ -159,6 +159,20 @@ def _match_groups_to_mask(
...
+                               repos: Iterable[str],
+                               time_from: datetime,
+                               time_to: datetime,
+                               max_time_to: datetime,
+                               applied_matches: Dict[str, ReleaseMatch],
+                               spans: Dict[str, Dict[str, Tuple[datetime, datetime]]],
+                               pdb: databases.Database) -> MissingReleases:
+        missing_releases = super(PreloadedReleaseLoader, cls)._find_missing_releases(
+            repos, time_from, time_to, max_time_to, applied_matches, spans, pdb)
+        return missing_releases._replace(missing_high=[])
```
vmarkovtsev, line 174: This is nasty, tuples are supposed to be immutable + we are calling the protected method.
```suggestion
        return MissingReleases([], [], [], missing_releases.ambiguous_branches_scanned)
```

server/athenian/api/controllers/miners/github/release_load.py
```diff
@@ -142,47 +145,12 @@ def gather_applied_matches() -> Dict[str, ReleaseMatch]:
...
-                    missed = True
-                if not missed:
-                    hits += 1
-        add_pdb_hits(pdb, "releases", hits)
+
+        missing_releases = cls._find_missing_releases(repos, time_from, time_to, max_time_to,
+                                                      applied_matches, spans, pdb)
+        missing_high = missing_releases.missing_high
+        missing_low = missing_releases.missing_low
+        missing_all = missing_releases.missing_all
```
vmarkovtsev, line 153:
```suggestion
        missing_high, missing_low, missing_all, ambiguous_branches_scanned = \
            cls._find_missing_releases(repos, time_from, time_to, max_time_to,
                                       applied_matches, spans, pdb)
```

vmarkovtsev (changes requested):

server/athenian/api/experiments/preloading/entries.py
se7entyse7en: Indeed the method name is quite misleading unfortunately: despite being called `_replace` it actually creates a new instance with that value replaced. WRT calling a protected method that's another hiccup, indeed quoting the doc:

> In addition to the methods inherited from tuples, named tuples support three additional methods and two attributes. To prevent conflicts with field names, the method and attribute names start with an underscore.

For sure not the best module of the python stdlib 😅

se7entyse7en:
@vmarkovtsev as confirmation, according to your review so you think we should disable everything and not only `missing_high`?

se7entyse7en:
If I understand correctly how the `missing_*` thing works, only `missing_high` should be affected since it's the one that depends on the "fresh" update and preloaded tables are not fresh.

vmarkovtsev:
> @vmarkovtsev as confirmation, according to your review so you think we should disable everything and not only missing_high?
Yes. We will fetch others otherwise in some cases. The heater makes sure that the rest are always empty, BUT there may be problems that will break that promise and we'll do unpreloaded queries.

se7entyse7en:
> there may be problems that will break the promise and we'll do unpreloaded queries.

I'm just wondering whether we're sacrificing too much accuracy in favor of performance. There are other places that haven't been preloaded, but they're fine performance-wise. I don't know though how much heavy are the SQL queries for fixing `missing_low` and `missing_all`.

vmarkovtsev:
We are not sacrificing anything at all here. Just hardening the logic and making it more robust to failures.

vmarkovtsev (approved):

PR #1625, opened 2021-07-22
[DEV-2645] Prefill pdb with data if WITH_PRELOADING=1

vmarkovtsev:


se7entyse7en:
@vmarkovtsev I debugged one of the failing tests, this specifically `tests/controllers/test_metrics_controller.py::test_calc_metrics_prs_labels_include`.

The test fails to assert equality with the expected value simply because the input data are different. Since preloading now has filled `pdb` data as well. In this specific test, the fetch of PRs is executed by also passing a non-empty `blacklist` parameter that depends on the done facts. Ofc in the case of non-preloaded mode, `pdb` is not filled, so there's no blacklist. I think there are other subtle changes like this in the input since I tried hardcoding the blacklist to `None` and still it failed, but you got the point.

I don't know what would be for you the best way to address this. I came up only with a couple of solutions:
1. change the expected values conditionally depending on whether we're using preloading or not, as done [here](https://github.com/athenianco/athenian-api/pull/1580/files#diff-54331dea039d4b904a84e1e43d9ad460990149ae0524c31554b76401fb2c1a7bR654),
2. fill `pdb` always, but this would require checking ALL the tests... but I guess it's nearer to the real usage since customers do not use the API before at least a precomputation is performed. In this case, the assertion values should be the same for preloading and non-preloading.
3. not fill `pdb` by default, but with a separate flag. This would let the current tests unchanged and not failing, and at the same time, I can enable the filling specifically for testing [this PR](https://github.com/athenianco/athenian-api/pull/1580).

se7entyse7en:
To me the first option seems to be the most "correct one", but the third option seems a good tradeoff.

se7entyse7en:
I'll start with option 3 then by rebasing [my PR](https://github.com/athenianco/athenian-api/pull/1580) on top of this. In case of disagreement, since we don't wanna focus work on performance unless necessary, we can wrap up this PR and just disable loading the data, but at least we have the logic and we review it again in the future.

se7entyse7en:
mhmmm it's actually trickier as it seems there are two groups:
- tests for things that work the same with `mdb` only and with `mdb` + `pdb`
- tests for things that work differently when `mdb` only and with `mdb` + `pdb`

and there's no clear division. Still need to think about it.

PR #2055, opened 2022-01-27
docker-compose -> V2 compatible

lwsanty:
https://docs.docker.com/compose/compose-file/compose-file-v2/#entrypoint

related to [[DEV-3558]]

[DEV-3558]: https://athenianco.atlassian.net/browse/DEV-3558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -3,27 +3,28 @@ services:
   api:
     image: ${API_IMAGE:-gcr.io/athenian-1/api:latest}
     restart: unless-stopped
-    entrypoint: ['/bin/sh']
-    command: >
-      -c "python3 -m athenian.api.models.state \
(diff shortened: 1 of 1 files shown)


vmarkovtsev (approved):
I am not happy that we pushed everything inside `entrypoint`, but whatever it takes to make the config work.

lwsanty:
@vmarkovtsev well, that's why I referred to the doc
> The entrypoint can also be a list, in a manner similar to dockerfile:
entrypoint: ["php", "-d", "memory_limit=-1", "vendor/bin/phpunit"]

so I would not consider it as something really bad

PR #2257, opened 2022-04-11
[DEV-3995] set account expires to far future

gaetano-guerriero:


server/athenian/api/serialization.py
```diff
@@ -90,37 +89,52 @@ def _deserialize_object(value: T) -> T:
     return value
 
 
-def deserialize_date(string: str) -> datetime.date:
-    """Deserializes string to date.
+# default bounds for deserialize_date and deserialize_datetime
+_DATETIME_MIN = datetime(2000, 1, 1)
+_DATE_MIN = _DATETIME_MIN.date()
+_DATETIME_MAX_FUTURE = timedelta(days=365)
```
vmarkovtsev, line 95: This suggests a datetime but the type is timedelta, let's rename to _?_MAX_FUTURE

server/athenian/api/controllers/user_controller.py
```diff
@@ -167,7 +166,9 @@ async def set_account_features(request: AthenianWebRequest, id: int, body: dict)
         for i, feature in enumerate(features):
             if feature.name == DBAccount.expires_at.name:
                 try:
-                    expires_at = deserialize_datetime(feature.parameters)
+                    expires_at = deserialize_datetime(
+                        feature.parameters, min_=None, max_future=None,
```
vmarkovtsev, line 170: I would leave the default minimum here because I've seen the client sending us weird dates Before Christ which we don't want to save in the DB.

vmarkovtsev (changes requested):

server/athenian/api/serialization.py
gaetano-guerriero: renamed as `_DATETIME_MAX_FUTURE_DELTA`, which is "maximum time delta in the future for datetime deserialization"

server/athenian/api/controllers/user_controller.py
gaetano-guerriero: done

vmarkovtsev (approved):

PR #2274, opened 2022-04-21
[DEV-4004] Add planner hints to optimize releases fetch

gaetano-guerriero:
Add some hints to the fetch tag releases query.

### DIFF
--- a/server/athenian/api/controllers/miners/github/release_load.py
+++ b/server/athenian/api/controllers/miners/github/release_load.py
@@ -865,15 +865,20 @@ class ReleaseMatcher:
         """Return the releases matched by tag."""
         if releases is None:
             with sentry_sdk.start_span(op="fetch_tags"):
-                releases = await read_sql_query(
-                    select([Release])
-                    .where(and_(
-                        Release.acc_id.in_(self._meta_ids),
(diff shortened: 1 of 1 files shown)


server/athenian/api/controllers/miners/github/release_load.py
```diff
@@ -865,15 +865,20 @@ async def match_releases_by_tag(self,
...
-                    self._mdb, Release)
+                query = select([Release]).where(and_(
+                    Release.acc_id.in_(self._meta_ids),
+                    Release.published_at.between(time_from, time_to),
+                    Release.repository_full_name.in_(coerce_logical_repos(repos)),
+                )).order_by(desc(Release.published_at))
+                for hint in (
+                    "Rows(ref_2 repo_2 *1000)",
+                    "Rows(ref_3 repo_3 *1000)",
+                    "Rows(rel rrs *1000)",
```
vmarkovtsev, line 876: Maybe let's try a more conservative `*100` instead? I guess the plan won't change much, but we will not overestimate for other accounts...

gaetano-guerriero: After a bit of bisection the lowest magic number is 250, I'm putting this.

PR #2278, opened 2022-04-22
[DEV-4044] precomputer discover new accounts separately

gaetano-guerriero:


server/athenian/api/precompute/discover_accounts.py
```diff
@@ -28,5 +38,39 @@ async def main(context: PrecomputeContext, args: argparse.Namespace) -> List[int
...
+
+        discovered.append((account, precomputed))
+
+    if args.partition:
+        result: Dict[str, List[int]] = {"precomputed": [], "fresh": []}
+        for precomputed, group_accounts in groupby(
+            sorted(discovered, key=itemgetter(1)), key=itemgetter(1),
+        ):
+            result_field = "precomputed" if precomputed else "fresh"
+            result[result_field] = [account for account, _ in group_accounts]
```
vmarkovtsev, line 50: I know that this code is correct and works, but here is a much simpler to comprehend alternative:
```suggestion
        result: Dict[str, List[int]] = {"precomputed": [], "fresh": []}
        for account, precomputed in discovered:
            result["precomputed" if precomputed else "fresh"].append(account)
        for accounts in result.values():
            accounts.sort()
```

server/athenian/api/precompute/discover_accounts.py
```diff
@@ -28,5 +38,39 @@ async def main(context: PrecomputeContext, args: argparse.Namespace) -> List[int
...
+
+async def _get_accounts(sdb: Database) -> List[Tuple[int, bool]]:
+    """Return the existing account IDs, each with a precomputed flag."""
+    # already precomputed accounts have at least one repo set and all their
+    # repository sets are precomputed
+    if sdb.url.dialect == "sqlite":
+        and_aggr = func.min
+    else:
+        and_aggr = func.bool_and
+    precomputed_expr = func.coalesce(and_aggr(RepositorySet.precomputed), False)
```
vmarkovtsev, line 64: We consider some account precomputed if the "all" reposet is precomputed. The other reposets can only be subsets of the main one. So we should add the condition `.name == RepositorySet.ALL` and remove the aggregation.

vmarkovtsev (changes requested):

server/athenian/api/precompute/discover_accounts.py
gaetano-guerriero: done

server/athenian/api/precompute/discover_accounts.py
gaetano-guerriero: intention here was to preserve accounts order in both cases.
Rewritten as you suggested.
Anyway now we can also remove `ORDER BY` from query ?

vmarkovtsev: We can remove it but then we should sort client-side if `partition = False`.

The goal of this sorting was to start precomputing from the oldest account to the newest, so that if we crash on a new one, all the rest are precomputed. If we `partition = True`, we don't care about the order anymore, but it's still nice for determinism.

vmarkovtsev: I forgot that we sort the result now, so yeah, let's remove.

server/athenian/api/precompute/discover_accounts.py
```diff
@@ -28,5 +37,30 @@ async def main(context: PrecomputeContext, args: argparse.Namespace) -> List[int
...
+async def _get_accounts(sdb: Database) -> List[Tuple[int, bool]]:
+    """Return the existing account IDs, each with a precomputed flag."""
+    # already precomputed accounts have an ALL repository set that is precomputed
+    # outer join will return one row per account, since (owner_id, name) is unique in RepositorySet
+    join_cond = sa.and_(
+        Account.id == RepositorySet.owner_id,
+        RepositorySet.name == RepositorySet.ALL,
+    )
+    stmt = sa.select(
+        [Account.id, sa.func.coalesce(RepositorySet.precomputed, False).label("precomputed")],
```
vmarkovtsev, line 58: `RepositorySet.precomputed` is `nullable=False`, we have a strong guarantee here.

server/athenian/api/precompute/discover_accounts.py
```diff
@@ -28,5 +37,30 @@ async def main(context: PrecomputeContext, args: argparse.Namespace) -> List[int
...
+        return sorted(chain(*discovered.values()))
+
+
+async def _get_accounts(sdb: Database) -> List[Tuple[int, bool]]:
+    """Return the existing account IDs, each with a precomputed flag."""
+    # already precomputed accounts have an ALL repository set that is precomputed
+    # outer join will return one row per account, since (owner_id, name) is unique in RepositorySet
+    join_cond = sa.and_(
+        Account.id == RepositorySet.owner_id,
+        RepositorySet.name == RepositorySet.ALL,
```
vmarkovtsev, line 55: We should move this inside `WHERE`. Postgres doesn't really like when you put constant conditions inside JOIN, it produces unexpected, bad plans. `WHERE` is always bullet-proof, on the other hand, and always leads to good plans.

vmarkovtsev (changes requested):

server/athenian/api/precompute/discover_accounts.py
gaetano-guerriero: order by removed

server/athenian/api/precompute/discover_accounts.py
gaetano-guerriero: need condition in `ON` to have accounts with reposets

vmarkovtsev: But wait, we've got a LEFT JOIN specifically because we may have accounts without reposets.

vmarkovtsev: True

vmarkovtsev (approved):

PR #2313, opened 2022-05-02
[DEV-4051] Define Align goal tables

gaetano-guerriero:


server/athenian/api/models/state/versions/69aed10c7ce1_goals.py
```diff
@@ -0,0 +1,117 @@
...
+            nullable=False,
+        ),
+        sa.Column("goal_id", sa.BigInteger().with_variant(sa.Integer(), "sqlite"), nullable=False),
+        sa.Column("team_id", sa.Integer(), nullable=False),
+        sa.Column("target", sa.JSON(), nullable=False),
+        sa.ForeignKeyConstraint(["goal_id"], ["goals.id"], name="fk_team_goal_goal"),
+        sa.ForeignKeyConstraint(["team_id"], ["teams.id"], name="fk_team_goal_team"),
+        sa.PrimaryKeyConstraint("goal_id", "team_id"),
+    )
+    op.drop_column("account_features", "created_at")
```
gaetano-guerriero, line 64: are these expected ? Should I remove them from migration ?

gaetano-guerriero: I found out the sqlite in 20.04 is too old to drop columns, so I'm removing these

server/athenian/api/models/state/models.py
```diff
@@ -284,3 +283,39 @@ class LogicalRepository(create_time_mixin(created_at=True, updated_at=True), Bas
...
+
+    id = Column(Integer(), primary_key=True)
+    account_id = Column(
+        Integer(),
+        ForeignKey("accounts.id", name="fk_goal_account"),
+        nullable=False,
+    )
+    template_id = Column(Integer(), nullable=False)
+    valid_from = Column(Date, nullable=False)
+    expires_at = Column(Date, nullable=False)
```
vmarkovtsev, line 301:
```suggestion
    valid_from = Column(TIMESTAMP(timezone=True), nullable=False)
    expires_at = Column(TIMESTAMP(timezone=True), nullable=False)
```
(we always require the explicit UTC timezone everywhere)

vmarkovtsev (changes requested):
Otherwise, LGTM!

server/athenian/api/models/state/models.py
gaetano-guerriero: I chose the "date" type since that is what the UI handles, and that has no timezone.
Should we have a timestamp instead ? Some related decisions:
- in API we truncate to day anyway ?
- when we receive `{validFrom: 2000-01-01, expiresAt: 2001-04-01}`, should we save `expires_at: 2001-04-01T23:59:59.99999` ? that would be conceptually right

vmarkovtsev: We kinda implicitly expect UTC in the UI. @akbarik @alex-athenian note: all the times that the webapp sends us must be in UTC.
Regarding `expires_at`, we can save `2001-04-02 00:00:00` (+1 day). That's how our filters work, I think.

gaetano-guerriero: So `2001-04-02 00:00:00` in DB is "2001-04-01" in API (for `expiresAt`) or API will express the full timestamp ?

vmarkovtsev: The first one is correct.

gaetano-guerriero: done, changed fields as timestamp with time zone

PR #2364, opened 2022-05-17
[DEV-4051] align update goal mutation

gaetano-guerriero:


server/athenian/api/align/goals/dbaccess.py
```diff
@@ -54,6 +64,49 @@ async def delete_goal(account_id: int, goal_id: int, sdb_conn: Connection) -> No
...
+async def delete_team_goals(
+    account_id: int, goal_id: int, team_ids: Sequence[int], sdb_conn: Connection,
+) -> None:
+    """Delete a set of TeamGoal-s from DB."""
+    assert team_ids
+    assert conn_in_transaction(sdb_conn)
+    await _validate_team_goal_deletions(account_id, goal_id, team_ids, sdb_conn)
+
+    where_clause = sa.and_(TeamGoal.goal_id == goal_id, TeamGoal.team_id.in_(team_ids))
+    delete_stmt = sa.delete(TeamGoal).where(where_clause)
```
vmarkovtsev, line 76: Formatting:
```suggestion
    where_clause =
    delete_stmt = sa.delete(TeamGoal).where(sa.and_(
        TeamGoal.goal_id == goal_id,
        TeamGoal.team_id.in_(team_ids),
    ))
```
we will never-ever inspect the where condition separately, so the variable is not needed. Also, this is how the existing code is formatted.

server/athenian/api/align/goals/dbaccess.py
```diff
@@ -54,6 +64,49 @@ async def delete_goal(account_id: int, goal_id: int, sdb_conn: Connection) -> No
...
+    await _validate_team_goal_assignments(account_id, goal_id, assignments, sdb_conn)
+
+    now = datetime.now(timezone.utc)
+    values = [
+        {
+            "goal_id": goal_id,
+            "team_id": assign.team_id,
+            "target": assign.target,
+            "created_at": now,
+            "updated_at": now,
```
vmarkovtsev, line 96:
```suggestion
            TeamGoal.goal_id.name: goal_id,
            TeamGoal.team_id.name: assign.team_id,
            TeamGoal.target.name: assign.target,
            TeamGoal.created_at.name: now,
            TeamGoal.updated_at.name: now,
```
We never hardcode column names because it is a nightmare to refactor them otherwise. You can grep the whole codebase for, say, `PullRequest.repository_full_name.name` and you'll get precisely all the usages.

server/athenian/api/align/goals/dbaccess.py
```diff
@@ -54,6 +64,49 @@ async def delete_goal(account_id: int, goal_id: int, sdb_conn: Connection) -> No
...
+        }
+        for assign in assignments
+    ]
+
+    insert = dialect_specific_insert(sdb_conn)
+    stmt = insert(TeamGoal)
+    upsert_stmt = stmt.on_conflict_do_update(
+        index_elements=["goal_id", "team_id"],
+        set_={"target": stmt.excluded.target, "updated_at": stmt.excluded.updated_at},
```
vmarkovtsev, line 105: same here: change to TeamGoal.*.name

server/athenian/api/align/goals/dbaccess.py
```diff
@@ -54,6 +64,49 @@ async def delete_goal(account_id: int, goal_id: int, sdb_conn: Connection) -> No
...
+            "created_at": now,
+            "updated_at": now,
+        }
+        for assign in assignments
+    ]
+
+    insert = dialect_specific_insert(sdb_conn)
+    stmt = insert(TeamGoal)
+    upsert_stmt = stmt.on_conflict_do_update(
+        index_elements=["goal_id", "team_id"],
```
vmarkovtsev, line 104: Regarding the indexed elements: do we have an index for those two fields?

server/athenian/api/align/goals/dbaccess.py
```diff
@@ -71,3 +124,49 @@ async def _validate_goal_creation_info(
...
+) -> None:
+    where_clause = sa.and_(
+        Goal.account_id == account_id,
+        Team.owner_id == account_id,
+        TeamGoal.goal_id == goal_id,
+        TeamGoal.team_id.in_(team_ids),
+    )
+    select_stmt = sa.select(
+        [TeamGoal.team_id],
+    ).join_from(TeamGoal, Goal).join(Team).where(where_clause)
```
vmarkovtsev, line 140: It's not obvious by which fields we join + we specify the same account_id condition twice. This code should perform better:
``` python
sa.select([TeamGoal.team_id]).select_from(join(join(TeamGoal, Goal, TeamGoal.goal_id == Goal.id)), Team, sa.and_(Goal.account_id == Team.owner_id, TeamGoal.team_id == Team.id)).where(sa.and_(Goal.account_id == account_id, TeamGoal.goal_id == goal_id, TeamGoal.team_id.in_(team_ids)))
```
Maybe we need a few indexes added for `account_id`/`owner_id`. We need them in any case.

server/athenian/api/align/goals/dbaccess.py
```diff
@@ -71,3 +124,49 @@ async def _validate_goal_creation_info(
...
+    )))
+    if not goal_exists:
+        raise GoalMutationError(
+            f"Goal {goal_id} doesn't exist or access denied", HTTPStatus.NOT_FOUND,
+        )
+
+    teams_where_clause = sa.and_(
+        Team.id.in_(a.team_id for a in assignments), Team.owner_id == account_id,
+    )
+    teams_stmt = sa.select([Team.id]).where(teams_where_clause)
```
vmarkovtsev, line 166:
```suggestion
    teams_stmt = sa.select([Team.id]).where(sa.and_(
        Team.id.in_(a.team_id for a in assignments),
        Team.owner_id == account_id,
    ))
```

In general, try to compress lines, it is easier to read and review more compact code. Here we avoid the extra mental pressure on the reviewer to remember the "teams_where_clause" variable.

server/athenian/api/align/mutations/goals.py
```diff
@@ -32,20 +38,49 @@ async def resolve_create_goal(
...
+async def resolve_update_goal(
+    _: Any,
+    info: GraphQLResolveInfo,
+    accountId: int,
+    input: dict,
+) -> dict:
+    """Update an existing Goal."""
+    update = _parse_update_goal_input(input, accountId)
+    goal_id = input["goalId"]
+    result = {"goal": {"id": goal_id}}
```
vmarkovtsev, line 67: Let's define the models for both the input and the output and use them here.

server/athenian/api/align/mutations/goals.py
```diff
@@ -32,20 +38,49 @@ async def resolve_create_goal(
...
+                await delete_team_goals(accountId, goal_id, deletions, sdb_conn)
+            if assignments:
+                await assign_team_goals(accountId, goal_id, assignments, sdb_conn)
+
+    return result
+
+
 def _parse_create_goal_input(input: Dict[str, Any], account_id: int) -> GoalCreationInfo:
     """Parse CreateGoalInput into GoalCreationInfo."""
     template_id = input["templateId"]
```
vmarkovtsev, line 86: Same here, let's define an input model.

server/athenian/api/align/mutations/goals.py
```diff
@@ -73,18 +108,76 @@ def _parse_create_goal_input(input: Dict[str, Any], account_id: int) -> GoalCrea
...
+
+def _parse_update_goal_input(input: Dict[str, Any], account_id: int) -> GoalUpdateInfo:
+    deletions = []
+    assignments = []
+
+    duplicates = []
+    both = []
+    invalid_targets = []
+    seen = set()
+    for change in input.get("teamGoalChanges", ()):
```
vmarkovtsev, line 148: Same here, we shouldn't hardcode "teamGoalChanges" string, it should exist inside a model.

server/athenian/api/align/mutations/goals.py
```diff
@@ -73,18 +108,76 @@ def _parse_create_goal_input(input: Dict[str, Any], account_id: int) -> GoalCrea
...
+        team_id = change["teamId"]
+        if team_id in seen:
+            duplicates.append(team_id)
+        if change.get("remove") and change.get("target"):
```
vmarkovtsev, line 152: same here: we should work with a model

server/athenian/api/db.py
```diff
@@ -399,3 +402,12 @@ def _conn_backend_in_transaction(conn_backend: ConnectionBackend) -> bool:
     if isinstance(raw_connection, aiosqlite.Connection):
         return raw_connection.in_transaction
     raise AssertionError(f"Unhandled db connection type {type(raw_connection)}")
+
+
+def dialect_specific_insert(conn: Connection) -> Callable:
+    """Return the dialect specific insert method for the connection."""
```
vmarkovtsev, line 408:
```suggestion
    """Return the specific insertion function for the connection's SQL dialect."""
```

vmarkovtsev (changes requested):

server/athenian/api/align/goals/dbaccess.py
gaetano-guerriero: yes this is the primary key of the table

server/athenian/api/align/goals/dbaccess.py
gaetano-guerriero: Done.
I would never intentionally put more mental pressure on the poor reviewer :)
By naming and outlining things, even if used once, I hope to help the reader by avoiding long statements and decomposing complexity in multiple statements.

server/athenian/api/align/goals/dbaccess.py
gaetano-guerriero: In your version the condition is moved from `WHERE` to `JOIN` `ON`, but the plan it's the same being an inner join:
```

Comment (gaetano-guerriero) [reply server/athenian/api/align/goals/dbaccess.py:140]: SQLAlchemy can detect join condition automatically given fk declarations but if you prefer I can write explicit conditions.

File: server/athenian/api/align/goals/dbaccess.py (hunk 6 above)
Comment (vmarkovtsev) [reply server/athenian/api/align/goals/dbaccess.py:166]: Haha, maybe I am too worn off by Sentry here. Variables are nice indeed, I fully agree. It's just that sometimes, this situation happens:
1. We crash on `sa.select`.
2. Error gets reported to Sentry.
3. Sentry takes the closest 10 variables to the line where the exception raised.
4. It orders it alphabetically and repr-s until it exceeds the size limit (4 or 8KB per stack frame, I think, it's not configurable).
5. There are garbage variables on top and they eat the whole quota so that, for example, `assignments` is not shown.
6. We cannot easily debug the error directly from Sentry.

image

File: server/athenian/api/align/mutations/goals.py (hunk 7 above)
Comment (gaetano-guerriero) [reply server/athenian/api/align/mutations/goals.py:67]: defined model for output, declared fields for input

File: server/athenian/api/align/goals/dbaccess.py [hunk 12]
```diff
@@ -54,6 +64,53 @@ async def delete_goal(account_id: int, goal_id: int, sdb_conn: Connection) -> No
...
+            TeamGoal.created_at.name: now,
+            TeamGoal.updated_at.name: now,
+        }
+        for assign in assignments
+    ]
+
+    insert = dialect_specific_insert(sdb_conn)
+    stmt = insert(TeamGoal)
+    upsert_stmt = stmt.on_conflict_do_update(
+        index_elements=["goal_id", "team_id"],
```
vmarkovtsev, line 105:
```suggestion
        index_elements=[TeamGoal.goal_id.name, TeamGoal.team_id.name],
```

server/athenian/api/align/models.py
```diff
@@ -16,3 +16,68 @@ def __init__(self, success: bool):
...
+        """Init the MutateGoalResultGoal."""
+        self._goal = goal
+
+    @property
+    def goal(self) -> MutateGoalResultGoal:
+        """Get the id of the goal."""
+        return self._goal
+
+
+class CreateGoalInputFields:
```
vmarkovtsev, line 55:
```suggestion
from enum import Enum

class CreateGoalInputFields(Enum):
```

vmarkovtsev (changes requested):

server/athenian/api/align/models.py
gaetano-guerriero: `Enum` requires the bloat `.value` when using the field, is this ok ?

server/athenian/api/align/goals/dbaccess.py
gaetano-guerriero: done

server/athenian/api/align/models.py
vmarkovtsev: Ok, then
```python
from athenian.api.models.web.base_model_ import Enum
class CreateGoalInputFields(metaclass=Enum):
```

gaetano-guerriero: done

vmarkovtsev (approved):

PR #2370, opened 2022-05-20
[DEV-4192] Fix bot team creation with unknown gh user login

gaetano-guerriero:


server/athenian/api/precompute/accounts.py
```diff
@@ -285,14 +285,39 @@ async def create_teams(account: int,
...
                                                Team.owner_id == account)))
     if bot_team is not None:
-        return num_teams, len(bot_team[Team.members.name])
+        return len(bot_team[Team.members.name])
+
     bots -= await fetch_bots.extra(mdb)
-    bots = set(chain.from_iterable(prefixer.user_login_to_node.get(u) for u in bots)) - {None}
+    bot_ids = set(chain.from_iterable(prefixer.user_login_to_node.get(u, ()) for u in bots))
     await sdb.execute(insert(Team).values(
-        Team(id=account, name=Team.BOTS, owner_id=account, members=sorted(bots))
```
gaetano-guerriero, line 296: id=account here was wrong but harmless

server/athenian/api/precompute/accounts.py
```diff
@@ -285,14 +285,39 @@ async def create_teams(account: int,
...
     bot_team = await sdb.fetch_one(select([Team.id, Team.members])
                                    .where(and_(Team.name == Team.BOTS,
                                                Team.owner_id == account)))
     if bot_team is not None:
-        return num_teams, len(bot_team[Team.members.name])
+        return len(bot_team[Team.members.name])
+
     bots -= await fetch_bots.extra(mdb)
-    bots = set(chain.from_iterable(prefixer.user_login_to_node.get(u) for u in bots)) - {None}
+    bot_ids = set(chain.from_iterable(prefixer.user_login_to_node.get(u, ()) for u in bots))
```
gaetano-guerriero, line 307: should we instead do `prefixer.user_login_to_node[u]` ? `u` comes from `node_bot` table and prefixer uses `api_users` so a KeyError is unexpected

vmarkovtsev: I had `[]` in the past and it raised a KeyError, I think. There are crazy mdb inconsistencies sometimes that fire in unexpected places.

Here the bots teams gets loaded from sdb and the user can add any crazy people inside.

vmarkovtsev (approved):

PR #2400, opened 2022-05-26
Add codeball to prove that every model has false positives

vmarkovtsev:


vmarkovtsev:
Haha it would have been weird if you didn't approve yourself :smile:

PR #2409, opened 2022-05-30
[DEV-4050] Consider full lineage in flatten_teams

gaetano-guerriero:


vmarkovtsev:
Does this work as if only_root is always true now?

gaetano-guerriero:
Yes it flattens all teams found in rows now, but the result for the root teams are the same as they were with only_root=True

vmarkovtsev (approved):

PR #2465, opened 2022-06-14
Dev 4193 black isort sample

gaetano-guerriero:
Just to check some diffs, CI is still missing


server/athenian/api/ariadne.py
```diff
@@ -1,14 +1,14 @@
-from pathlib import Path
 import traceback
+from pathlib import Path
 from typing import Any, Callable, Iterable, Optional
 
 import aiohttp.web
+import sentry_sdk
```
vmarkovtsev, line 6: Why did it put sentry_sdk here?

server/requirements-lint.txt
```diff
@@ -1,16 +1,19 @@
...
-flake8-import-order==0.18.1
-flake8-quotes==3.3.1
 flake8-commas==2.1.0
 flake8-debugger==4.1.2
+flake8-docstrings==1.6.0
+flake8-quotes==3.3.1
 flake8-sfs==0.0.3
-pydocstyle==6.1.1
+flake8==4.0.1
+git+ssh://git@github.com:/athenianco/chorny@v1.4.0#egg=chorny==v1.4.0
```
vmarkovtsev, line 11: It is a public Pypi package, no need for this

vmarkovtsev (changes requested):

server/athenian/api/ariadne.py
gaetano-guerriero: by default pure import before from ..., configured as before now

vmarkovtsev (approved):

PR #2476, opened 2022-06-16
[DEV-4366] fix expire notify uninstalled account

gaetano-guerriero:


.github/workflows/push.yml
```diff
@@ -211,9 +211,10 @@ jobs:
         MDB: postgresql://postgres:postgres@0.0.0.0:5432/metadata_%s?min_size=2&max_size=8
         PDB: postgresql://postgres:postgres@0.0.0.0:5432/precomputed_%s?min_size=2&max_size=6
         RDB: postgresql://postgres:postgres@0.0.0.0:5432/persistentdata_%s?min_size=2&max_size=5
+        MDB_EPHEMERAL: postgresql://postgres:postgres@0.0.0.0:5432/metadata_ephemeral_%s?min_size=2&max_size=8
```
gaetano-guerriero, line 214: when approved do not merge, I'll make gha changes in a separate pr first

.github/workflows/push.yml
```diff
@@ -211,9 +211,10 @@ jobs:
...
       run: |
         set -x
-        for db in state metadata precomputed persistentdata; do
+        for db in state metadata precomputed persistentdata metadata_ephemeral; do
```
vmarkovtsev, line 217: Can you please avoid this ephemeral change or at least move to a separate PR, because it needs a further discussion. I am against this overcomplication, for example.

server/athenian/api/precompute/notify_almost_expired_accounts.py
```diff
@@ -36,12 +36,15 @@ async def main(context: PrecomputeContext, args: argparse.Namespace) -> None:
     )
     users = dict(user_rows)
     name_rows = await mdb.fetch_all(
-        select([GitHubAccount.id, GitHubAccount.name]).where(
+        select(GitHubAccount.id, GitHubAccount.name).where(
```
vmarkovtsev, line 39: I think sqla 1.4 select wants a list just like 1.3 did, are you sure?

vmarkovtsev (changes requested):

.github/workflows/push.yml
gaetano-guerriero: My target here was to have mdb in tests that is empty and writable

server/athenian/api/precompute/notify_almost_expired_accounts.py
gaetano-guerriero: Yes 1.4 accepts both formats
https://docs.sqlalchemy.org/en/14/changelog/migration_14.html?highlight=select#select-case-now-accept-positional-expressions

vmarkovtsev (changes requested):
I'll be able to review as soon as there is no ephemeral DB

vmarkovtsev:
Wait... why do you need a mutable mdb in the first place? Shouldn't you change the mdb fixture instead? Add your test case in `fill_metadata_session` in `sample_db_data.py`, it is only called once.

The reason why I am so strongly against the ephemeral DB is because 4 DBs is more than enough for the testing. Requiring 5 DBs increases the overall complexity, and I am doing everything to reduce the complexity vs. growing it.

gaetano-guerriero:
I think a mutable and not persistent mdb is useful to have a custom and isolated context for the test.

I'm removing this and I hope you can reason about a future sustainable, non-complex, solution to satisfy this use case.

Otherwise you have more complexity in tests, since either

- you have to write/restore the DB, cf. test_extract_branches_main(), test_filter_jira_deleted_repositories() for example; that
  of course make less probable that the poor dev takes the effort to make custom context for the test, he will probably more than  satisfied with what the code does with the monolithic fixture. The problem comes when a new bug is not replicable with the monolothic fixture.
 
- you have to refer to a global context created elsewhere using magic numbers; for example in test_account_not_found_in_mdb() I only need a github account number not existing in mdb and it's easy, but can become complex quite fast

vmarkovtsev (approved):
Thank you!

vmarkovtsev:
Let's remember the issue with the DBs, I agree that it's not ideal atm. A good point for the next meeting.

PR #2493, opened 2022-06-21
[DEV-4282] batched gh metrics computation

gaetano-guerriero:


server/tests/internal/features/test_entries.py
```diff
@@ -0,0 +1,123 @@
...
+            time_intervals=all_intervals,
+            lines=[],
+            environments=[],
+            repositories=[{"src-d/go-git"}],
+            participants=[{}],
+            **base_kwargs,
+        )
+
+        await wait_deferred()
+        batch_calc_res = await calculator.batch_calc_pull_request_metrics_line_github(
```
gaetano-guerriero, line 87: in this test this is 10% faster than simple calc...()

server/athenian/api/internal/features/entries.py
```diff
@@ -1450,3 +1545,32 @@ def make_calculator(
...
+    """Information for a single pull request metrics request."""
+
+    metrics: Sequence[str]
+    time_intervals: Sequence[Sequence[datetime]]
+    lines: Sequence[int]
+    environments: Sequence[str]
+    repositories: Sequence[Collection[str]]
+    participants: list[PRParticipants]
+
+    def __str__(self) -> str:
```
vmarkovtsev, line 1561: Can you please also define `__sentry_repr__` with basic stats - how many metrics, repos, participants, etc. in each group.

server/athenian/api/internal/features/entries.py
```diff
@@ -206,13 +209,113 @@ def _align_time_min_max(
...
+            all_participants,
+            labels,
+            jira,
+            exclude_inactive,
+            bots,
+            release_settings,
+            logical_settings,
+            prefixer,
+            fresh,
+            need_jira_mapping(all_metrics),
```
vmarkovtsev, line 275: it is worth renaming `from x import need_jira_mapping as pr_metrics_need_jira_mapping`

vmarkovtsev (approved):
Yes, something like this

server/athenian/api/internal/features/entries.py
gaetano-guerriero: done in the same style of `PRDataFrames`

PR #2499, opened 2022-06-22
[DEV-4282] batched metrics computation

gaetano-guerriero:


server/athenian/api/align/queries/metrics.py
```diff
@@ -192,185 +335,91 @@ def _jirafy_teams(
...
-        for i, metric in enumerate(jira_metrics):
-            metric_teams = result[interval][metric] = {}
-            for team, team_metric_values in zip(teams, jira_metric_values):
-                metric_teams[team] = team_metric_values[0][interval_idx][0][i].value
+    def add_request(self, request: _MetricsLineRequestGenT, team_ids: Sequence[int]) -> None:
+        self._requests.append((request, team_ids))
 
-    return result
+    def get_requests(self) -> Sequence[_MetricsLineRequestGenT]:
+        return tuple(req[0] for req in self._requests)
```
vmarkovtsev, line 369: We should make as few calls to `calc()` as possible.

Every request is:
- One metric.
- Two time intervals (previous and current). Metrics calculated on wrong time intervals are a pure waste of resources.
- People. The actual participants are in different formats but always derive from the same set. Metrics calculated on wrong people are useless. `people = set[str]` 

Build the two-level dictionary (two time intervals) -> people -> list[metric]

Deduplicate list[metric].

Turn each list to a tuple, then inverse the bottom level: (two time intervals) -> tuple[metric] -> list[people]

Deduplicate list[people]. We've reached a reasonably good grouping. Each tree path is a separate call to `calc()`.

vmarkovtsev (changes requested):
I suggest to split the PR into two:
1. Remaining batch functions in `entries.py` + related code
2. Changes in Align

gaetano-guerriero:
Opened the GraphQL independent part separately here
https://github.com/athenianco/athenian-api/pull/2500

server/athenian/api/align/queries/metrics.py
gaetano-guerriero: done

server/athenian/api/align/queries/metrics.py
```diff
@@ -192,185 +331,112 @@ def _jirafy_teams(
...
-    )
+    # intervals => metrics => team ids
+    simplified_tree: dict[Sequence[Interval], dict[tuple[str, ...], set[int]]] = {}
+    for intervals, intervals_tree in requests_tree.items():
+        simplified_tree[intervals] = {}
 
+        for team_id, metrics in intervals_tree.items():
+            sorted_metrics = tuple(sorted(metrics))
+            simplified_tree[intervals].setdefault(sorted_metrics, set())
+            simplified_tree[intervals][sorted_metrics].add(team_id)
```
vmarkovtsev, line 360:
```suggestion
            simplified_tree[intervals].setdefault(sorted_metrics, set()).add(team_id)
```

vmarkovtsev (approved):

PR #2517, opened 2022-06-24
[DEV-4483] goals query only with targets param

gaetano-guerriero:


server/athenian/api/align/models.py
```diff
@@ -84,22 +84,46 @@ def __init__(
...
+            total_members_count = len(total_members)
+        self._total_members_count = total_members_count
+
+    def with_children(self, children: list[TeamTree]) -> TeamTree:
+        """Return a copy of the object with children property replaced.
+
+        Properties depending from `children` retain the original value of the object.
+
+        """
+        copy = cast(TeamTree, self.copy())
```
vmarkovtsev, line 124: I think it is possible to change `copy()` to return the correct type https://stackoverflow.com/questions/68082661/typing-hinting-in-parent-class-for-a-method-that-returns-object-of-any-child-cla

Otherwise, we are heading towards a parody on Java/C#.

server/athenian/api/align/queries/goals.py
```diff
@@ -145,3 +189,30 @@ def _team_tree_to_team_goal_tree(
...
+) -> Optional[TeamTree]:
+    """Remove unwanted teams from a TeamTree.
+
+    Empty branches are pruned.
+    """
+    kept_children = [
+        kept
+        for kept in (_team_tree_prune_empty_branches(c, keep_team_fn) for c in team_tree.children)
+        if kept is not None
+    ]
```
vmarkovtsev, line 214:
```suggestion
    kept_children = [
        kept
        for c in team_tree.children
        if (kept := _team_tree_prune_empty_branches(c, keep_team_fn)) is not None
    ]
```
without the second iterator

server/athenian/api/align/queries/goals.py
```diff
@@ -145,3 +189,30 @@ def _team_tree_to_team_goal_tree(
...
+    if keep_team_fn(team_tree.id) or kept_children:
```
vmarkovtsev, line 215:
```suggestion
    if kept_children or keep_team_fn(team_tree.id):
```
(performance, lazy computation)

server/athenian/api/align/queries/goals.py
```diff
@@ -81,18 +70,73 @@ async def resolve_goals(
...
+        if only_with_targets:
+            # if the flag is true
+            # 1 - prune team tree to keep only branches with assigned teams
+            keep_team = _TeamFilterFromTeamGoalRows(team_goal_rows)
+            goal_team_tree = _team_tree_prune_empty_branches(team_tree, keep_team)
+            # assigned_teams is a non empty subset of all the ids in team_tree,
+            # so the pruned tree is never empty
+            assert goal_team_tree is not None
+            # 2 - in the metrics requests only ask for teams present in the pruned tree
+            goal_tree_team_ids = goal_team_tree.get_flatten_team_ids()
```
vmarkovtsev, line 127:
```suggestion
            goal_tree_team_ids = goal_team_tree.flatten_team_ids()
```

or

```suggestion
            goal_tree_team_ids = goal_team_tree.get_flat_team_ids()
```

I like the first variant more.

vmarkovtsev (changes requested):

PR #2528, opened 2022-06-27
[DEV-4372] Track resolution prometheus counters in precomputer

gaetano-guerriero:


server/athenian/api/precompute/accounts.py
```diff
@@ -72,17 +72,30 @@ async def main(context: PrecomputeContext, args: argparse.Namespace) -> Optional
...
+            try:
+                await precompute_reposet(
+                    reposet, meta_ids, context, args, time_to, no_time_from, time_from,
+                )
+            except Exception:
+                track_time()
+                track_failure()
+            else:
+                track_time()
+                track_success()
```
vmarkovtsev, line 98:
```suggestion
            try:
                await precompute_reposet(
                    reposet, meta_ids, context, args, time_to, no_time_from, time_from,
                )
                track_success()
            except Exception:
                track_failure()
            finally:
                track_time()
```

server/athenian/api/precompute/accounts.py
```diff
@@ -116,6 +129,9 @@ async def callback(context: PrecomputeContext):
                         reposet.owner_id,
                         status[1],
                     )
+                    track_failure()
+                else:
+                    track_success()
             track_time()
```
vmarkovtsev, line 135: I think we should move `track_time()` inside `finally:`, otherwise, we don't measure the time if there is a nasty system exception.

server/athenian/api/precompute/accounts.py
```diff
@@ -487,8 +501,56 @@ def track(self, account: int, meta_ids: Sequence[int], is_fresh: bool) -> None:
...
-            github_account=",".join(map(str, sorted(meta_ids))),
+            github_account=_github_account_tracking_label(meta_ids),
             is_fresh=is_fresh,
         ).observe(elapsed)
         self._log.info("tracking precompute duration: %.3f seconds", elapsed)
-        push_metrics(self._gateway, self._PROMETHEUS_JOB)
+        push_metrics(self._gateway)
+
+
+class _ResolutionTracker:
```
vmarkovtsev, line 511:
```suggestion
class _StatusTracker:
```

I recommend renaming "resolution" to "status" because it's a more common term in this context. "Resolution" is commonly used with timers, e.g. Python's default timer's resolution is 1 microsecond. Another meaning is: this problem's or issue's resolution is "Won't fix".

GHA defines "status" and "conclusion" of a check run, where "status" is `SUCCESS`, `FAILURE`, etc.

vmarkovtsev (changes requested):

server/athenian/api/precompute/accounts.py
gaetano-guerriero: done

PR #2548, opened 2022-07-05
[DEV-4568] Use JOIN VALUES in commits query in mine_releases

gaetano-guerriero:
This is faster than ANY VALUES ...


vmarkovtsev (commented):
@gaetano-guerriero Have you compared the plans? I very much doubt that this will be faster because:
- Smaller batches perform worse. This is obvious. 
- PG does not predict the correct number of rows without hints, hence chooses a wrong join method.

I have spent months on optimization of this particular query, and dug really deep in the performance here. I know that PG can easily fool you by caching pages so that it may seem that one way is faster than the other, but it's not faster. I strongly recommend comparing the plans and eventually revert this.

gaetano-guerriero:
Without updated statistics JOIN VALUES was always able to finish and this allowed account to be precomputed.
IN ... or = ANY (VALUES couldn't finish also with small batches

PR #2568, opened 2022-07-18
[DEV-4616] precomputer resync bots team

gaetano-guerriero:


server/athenian/api/precompute/accounts.py
```diff
@@ -212,26 +217,27 @@ async def precompute_reposet(
     try:
         prefixer, bots, new_items = await gather(
             Prefixer.load(meta_ids, mdb, cache),
-            fetch_bots(reposet.owner_id, meta_ids, mdb, sdb, None),
+            fetch_bots.get_account_bots(reposet.owner_id, meta_ids, mdb, sdb, None),
```
vmarkovtsev, line 220: Should stay the old code

server/athenian/api/internal/team.py
```diff
@@ -223,3 +229,30 @@ async def delete_team(team: Row, sdb_conn: Connection) -> None:
...
+    db_members = team[Team.members.name]
+    new_members = members_set.difference(db_members)
+
+    if new_members:
+        all_members = sorted(chain(new_members, db_members))
+        await sdb_conn.execute(
+            sa.update(Team)
+            .where(Team.id == team[Team.id.name])
+            .values({Team.updated_at: datetime.now(timezone.utc), Team.members: all_members}),
+        )
```
vmarkovtsev, line 256: We should rewrite this to write the new team members without any special logic inside.

Who are the new team members is the caller's responsibility.

server/athenian/api/precompute/accounts.py
```diff
@@ -483,6 +484,33 @@ async def _ensure_root_team(account: int, sdb: Database) -> int:
...
+
+    async with sdb.connection() as sdb_conn:
+        async with sdb_conn.transaction():
+            bot_team = await get_bots_team(account, sdb)
+            if bot_team is None:
+                return
+            log.info("syncing bots team members")
+            new_members = await sync_team_members(bot_team, bot_ids, sdb_conn)
+            if new_members:
+                log.info("added members to Bots team: %s", ",".join(map(str, new_members)))
```
vmarkovtsev, line 511: The flow should be:

- Load the current bots team members.
- Find the new bots team members minus "extra". This is exactly the code inside `_ensure_bot_team` down until we insert into the DB. We factor that code out.
- If the members are the same, return. Otherwise, issue a SQL update.

According to how I see it, all the old code lives as-is. Also, the new code becomes simpler.

vmarkovtsev (changes requested):

server/athenian/api/internal/miners/github/bots.py
```diff
@@ -56,17 +63,18 @@ async def extra(self, mdb: morcilla.Database) -> FrozenSet[str]:
...
         cache: Optional[aiomcache.Client],
-    ) -> FrozenSet[str]:
+    ) -> frozenset[str]:
         """
         Return the bot logins.
 
         There are two parts: global bots in mdb and local bots in the Bots team in sdb.
         """
         await self._ensure_fetched(mdb)
+        assert self._bots is not None
```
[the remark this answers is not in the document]
gaetano-guerriero: mypy doesn't infer instance variable type across method calls, bots would still be `Optional[dict[int, frozenset[str]]]` here as declared

server/athenian/api/internal/team.py
gaetano-guerriero: What if I call the function `add_team_members` ? 
If needed I can split another _set_team_members without logic inside

server/athenian/api/internal/miners/github/bots.py
```diff
@@ -85,13 +93,43 @@ async def __call__(
...
+@dataclasses.dataclass(frozen=True)
+class AccountBots:
+    """The collection of bots for an account.
+
+    Bots are partitioned into global (common to every account)
+    and local (specific of this account) groups
+
+    """
+
+    global_bots: frozenset[str]
```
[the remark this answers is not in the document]
gaetano-guerriero: IMHO it is an improvement, not an alternative behavior. If we sometimes need the "account specific" bots separated from "extra" bots it's natural to have this API instead of doing:
```
bots = fetch_bots()
account_specific_bots = bots - fetch_bots.extra(mdb)
```

But of course I can easily delete  the new method.

server/athenian/api/internal/miners/github/bots.py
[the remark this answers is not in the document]
vmarkovtsev: Got it. Yes, I had the same experience in databases/morcilla

vmarkovtsev (approved):
To speed up the work, let's merge this as-is. I will read the code myself and will see how we can delete the new lines, if it's possible.

PR #2577, opened 2022-07-18
[DEV-4624] datetime parsing support custom timezones

gaetano-guerriero:


vmarkovtsev (approved):

gaetano-guerriero:
This has been reworked and should be re-reviewed:
* it's not feasible to reject date time strings without timezone, we emit them ourselves in tests with sqlite3 and we expect to parse them back
* now `deserialize_datetime` checks for "2022-07-05 23:14:11 CEST" or "2022-07-05 23:14:11 MET" and rejects them (and that was the bug to fix)
* `/events/deployments/` explicitly checks that the dates have the timezone

server/athenian/api/models/web/deployment_notification.py
```diff
@@ -66,6 +66,12 @@ def __init__(
 
     def validate_timestamps(self) -> None:
         """Post-check the request data."""
+        if self.date_finished.tzinfo is None or self.date_started.tzinfo is None:
+            raise ResponseError(
+                InvalidRequestError(
+                    "`date_finished` and `date_started` must include the timezone",
+                ),
+            )
```
vmarkovtsev, line 74: It would be nice to raise independent errors:
```
for attr in ("date_finished", "date_started"):
    if getattr(self, attr).tzinfo is None:
       raise f"`{attr}` must include the time zone"
```

gaetano-guerriero: done

PR #2586, opened 2022-07-21
[DEV-4616] Refactor adding bots in precomputer

vmarkovtsev:


server/tests/controllers/miners/github/test_bots.py
```diff
@@ -1,19 +0,0 @@
-from athenian.api.internal.miners.github.bots import bots
-
-
-class TestBots:
-    async def test_call(self, mdb, sdb):
```
gaetano-guerriero, line 5: this test maybe can stay

server/athenian/api/precompute/accounts.py
```diff
@@ -418,47 +404,99 @@ async def report_precompute_success():
...
-        insert(Team).values(
-            Team(name=Team.BOTS, owner_id=account, parent_id=root_team_id, members=sorted(bot_ids))
-            .create_defaults()
-            .explode(),
-        ),
-    )
-    return len(bot_ids)
+    async with sdb.connection() as sdb_conn:
+        async with sdb_conn.transaction():
+            old_team = await fetch_bots.team(account, sdb_conn)
```
gaetano-guerriero, line 466: we're making this query twice now, right ?

server/athenian/api/precompute/accounts.py
```diff
@@ -418,47 +404,99 @@ async def report_precompute_success():
...
+            if await is_postgresql(sdb):
+                sql = sql.on_conflict_do_update(
+                    constraint=Team.__table__.primary_key,
+                    set_={
+                        col.name: getattr(sql.excluded, col.name)
+                        for col in (Team.members, Team.updated_at)
+                    },
+                )
+            else:
+                sql = sql.prefix_with("OR REPLACE")
```
gaetano-guerriero, line 486: on_conflict_do_update should work also with sqlite, see `assign_team_goals`

server/tests/controllers/miners/github/test_bots.py
vmarkovtsev: Yep, good point

server/athenian/api/precompute/accounts.py
vmarkovtsev: Twice. This one is made inside a transaction, so that we don't have any contradictions.

server/athenian/api/precompute/accounts.py
vmarkovtsev: Jeez, we have quite a few similar places in the existing code. Now I understand why you factored the dialect-specific insert but left the rest.

PR #2599, opened 2022-07-26
[DEV-4673] Use api_teams view instead of node_team

gaetano-guerriero:
api_views view automatically filters out deleted teams.


vmarkovtsev (approved):

vmarkovtsev:
@gaetano-guerriero Ideally, we would bump the schema version here: https://github.com/athenianco/athenian-api/blob/master/server/athenian/api/models/metadata/__init__.py#L15

gaetano-guerriero:
That should match the number of the last migration applied in metadata DB  ?

PR #2609, opened 2022-07-27
[DEV-4663] align sdb models to migrations

gaetano-guerriero:
Run CI with pull_request event here
https://github.com/athenianco/athenian-api/runs/7547122214?check_suite_focus=true
https://github.com/athenianco/athenian-api/runs/7574761678?check_suite_focus=true

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -68,6 +68,22 @@ jobs:
     name: custom checks
     if: "!contains(github.event.head_commit.message, 'Bump version') || github.event_name != 'push'"
     runs-on: ubuntu-22.04
+    services:
+      postgres:
+        image: postgres:11.8
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: postgres
+        ports:
+        - 5432:5432
+        options: >-
+          --mount type=tmpfs,destination=/var/lib/postgresql/data
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 1s
+          --health-retries 10
+          --health-start-period 2s
     steps:
     - uses: tspascoal/get-user-teams-membership@v1
       id: membership
@@ -93,7 +109,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: ${{ env.PIP_CACHE }}
-        key: ubuntu-22.04-pip-custom-checks-${{ hashFiles('server/requirements.txt', 'server/requirements-lint.txt') }}
+        key: ubuntu-22.04-pip-custom-checks-${{ hashFiles('server/requirements.txt', 'server/requirements-lint.txt', 'server/requirements-test.txt') }}
         restore-keys: ubuntu-22.04-pip-custom-checks-
     - name: pip
       run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/push.yml
```diff
@@ -68,6 +68,22 @@ jobs:
     name: custom checks
     if: "!contains(github.event.head_commit.message, 'Bump version') || github.event_name != 'push'"
     runs-on: ubuntu-22.04
+    services:
+      postgres:
+        image: postgres:11.8
```
vmarkovtsev, line 73: We are using 13 in prod, it is better to test with 13 in the CI

server/athenian/api/models/state/env.py
```diff
@@ -27,6 +29,10 @@
 # ... etc.
 
 
+def _sqlalchemy_url_override() -> Optional[str]:
+    return os.environ.get("ALEMBIC_SDB", None)
```
vmarkovtsev, line 33: We shouldn't modify the env file. It is trivial to patch the config file with `sed` in the CI as needed.

server/athenian/api/models/state/models.py
```diff
@@ -325,6 +327,7 @@ class MappedJIRAIdentity(create_time_mixin(created_at=True, updated_at=True), Ba
     """JIRA identity mapping."""
 
     __tablename__ = "jira_identity_mapping"
+    __table_args__ = (Index("jira_identity_mapping_jira_user_id", "account_id", "jira_user_id"),)
```
vmarkovtsev, line 330: No-no, indexes are too complex and advanced to mention in the models. We should ignore their existence.

server/scripts/check_sdb_migrations.sh
```diff
@@ -0,0 +1,21 @@
...
+dbname=check_sdb_migrations
+
+export ATHENIAN_INVITATION_KEY=777 # needed by migrations
+
+psql -c "DROP DATABASE IF EXISTS $dbname" -h 0.0.0.0 -p 5432  -d postgres
+psql -c "CREATE DATABASE $dbname TEMPLATE 'template0' LC_COLLATE 'C.UTF-8'" -h 0.0.0.0 -p 5432
+
+python3 -m athenian.api.models.state "postgresql://$PGUSER:$PGPASSWORD@0.0.0.0:5432/$dbname"
+# a spurious backup table is left by migrations
+psql -c "DROP TABLE jira_identity_mapping_old" -h 0.0.0.0 -p 5432  -d $dbname
```
vmarkovtsev, line 16: We should drop this table in a migration, it is an old artifact.

server/scripts/check_sdb_migrations.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+
+set -eu
+
+echo $PGUSER > /dev/null
+echo $PGPASSWORD > /dev/null
```
vmarkovtsev, line 6: This doesn't make sense :smile:

vmarkovtsev (changes requested):

server/scripts/check_sdb_migrations.sh
gaetano-guerriero: This tests that the variables are set, with set -u

server/athenian/api/models/state/models.py
gaetano-guerriero: I don't understand, should we only manually create them in migrations ? Do we have examples of indexes we were not able to declare in models ? We already have indexes on single columns, unique constraints, check constraints with a custom expression in the models.

gaetano-guerriero: I missed the INCLUDE

vmarkovtsev (approved):

PR #2624, opened 2022-08-01
[DEV-4469] improve metric entries calculator cache hit

gaetano-guerriero:
I will  replicate for other calc_* methods if approach  is right


server/athenian/api/internal/features/entries.py
```diff
@@ -130,6 +130,38 @@ def _postprocess_cached_facts(
...
+    def _sort_level(lst, depth: int, sort_indexes: list):
+        if depth <= 0:
+            elements = [lst[idx] for idx in sort_indexes]
+            if isinstance(lst, np.ndarray):  # keep np array as np array
+                return np.array(elements)
+            return elements
+        return [_sort_level(sublst, depth - 1, sort_indexes) for sublst in lst]
+
+    # list of metrics to sort is the sixth dimension
+    return _sort_level(result, 5, metrics_sorted_indexes)
```
vmarkovtsev, line 162: Won't this work?
```suggestion
    return result[: ,:, :, :, :, metrics_sorted_indexes]
```

server/athenian/api/internal/features/entries.py
```diff
@@ -130,6 +130,38 @@ def _postprocess_cached_facts(
...
+            sort_indexes = [metrics.index(m) for m in sorted_metrics]
+            return _sorted_result_metrics(result, sort_indexes)
+
+        return result
+
+    @staticmethod
+    def preprocess(result: np.ndarray, *, metrics: Sequence[str], **kwargs: Any) -> np.ndarray:
+        # cache value must be stored with sorted(metrics), adapt if needed
+        if (sorted_metrics := sorted(metrics)) != metrics:
+            sort_indexes = [sorted_metrics.index(m) for m in metrics]
```
vmarkovtsev, line 147:
```suggestion
            sort_indexes = np.searchsorted(sorted_metrics, metrics)
```
O(n log n) vs. O(n^2)

vmarkovtsev (changes requested):

server/athenian/api/internal/features/entries.py
gaetano-guerriero: done

server/athenian/api/internal/features/entries.py
gaetano-guerriero: I tried to do everything with numpy but problem is that ndarray has 4 axis, the last two dimensions are python lists. This is because "time intervals" for each granularity is variable.

vmarkovtsev: I see. Then this code:
```python
for obj in result.ravel():
    for m in obj:
        m[:] = (m[idx] for idx in sort_indexes)
```

gaetano-guerriero: tried a variant since I need to preserve result

server/athenian/api/internal/features/entries.py
```diff
@@ -121,15 +121,41 @@
...
         raise CancelCache()
     return result
 
 
+class _PRMetricsLineGHCache:
+    @staticmethod
+    def postprocess(result: np.ndarray, *, metrics: Sequence[str], **kwargs: Any) -> np.ndarray:
+        # cache value has been stored with sorted(metrics), adapt before returning to caller
+        if (sorted_metrics := sorted(metrics)) != metrics:
+            sort_indexes = [metrics.index(m) for m in sorted_metrics]
```
vmarkovtsev, line 138: Sorry, I've just realized that this code is not symmetric.
```python
sort_indexes = np.argsort(metrics)
```

server/athenian/api/internal/features/entries.py
```diff
@@ -121,15 +121,41 @@
...
+    def preprocess(result: np.ndarray, *, metrics: Sequence[str], **kwargs: Any) -> np.ndarray:
+        # cache value must be stored with sorted(metrics), adapt if needed
+        if (sorted_metrics := sorted(metrics)) != metrics:
+            sort_indexes = np.searchsorted(sorted_metrics, metrics)
+            return _sorted_result_metrics(result, sort_indexes)
+        return result
+
+
+def _sorted_result_metrics(result: np.ndarray, metrics_sort_indexes: Iterable[int]) -> np.ndarray:
+    copy = result.reshape(-1).copy()
```
vmarkovtsev, line 153: Can you please use `ravel()` - it is more idiomatic than remembering the magic argument values.

vmarkovtsev (changes requested):

vmarkovtsev (approved):

PR #2641, opened 2022-08-04
[DEV-4716] hide already precomputed outlier first deploy

gaetano-guerriero:


server/athenian/api/internal/miners/github/deployment.py
```diff
@@ -2674,31 +2667,33 @@ async def hide_outlier_first_deployments(
...
 
-    log.info("Clearing %d outlier first deployments", len(toclear_deploys))
-
-    stmts: list[Executable] = []
+    # group outlier deploys by name and consider only deploys still to be hidden
+    async def deploy_to_be_hidden(name: str, repositories: Sequence[str]) -> bool:
+        stmt = sa.select(1).where(
+            GitHubCommitDeployment.acc_id == account,
+            GitHubCommitDeployment.deployment_name == name,
+            GitHubCommitDeployment.repository_full_name.in_(repositories),
```
vmarkovtsev, line 2680:
```suggestion
```
better to remove this condition, will be more reliable

vmarkovtsev: Since you'll have to touch this, let's also union with similar checks in `GitHubPullRequestDeployment` and `GitHubReleaseDeployment`. It's impossible to have the latter without the former ofc, but who knows, maybe the process crashed somewhere on line 2704 where we clear those tables.

vmarkovtsev (changes requested):

server/athenian/api/internal/miners/github/deployment.py
```diff
@@ -2674,31 +2667,33 @@ async def hide_outlier_first_deployments(
...
-
-    if not toclear_deploys:
-        return
 
-    log.info("Clearing %d outlier first deployments", len(toclear_deploys))
+    tables = (GitHubCommitDeployment, GitHubPullRequestDeployment, GitHubReleaseDeployment)
+    # group outlier deploys by name and consider only deploys still to be hidden
 
-    stmts: list[Executable] = []
+    async def deploy_to_be_hidden(name: str, repositories: Sequence[str]) -> bool:
```
vmarkovtsev, line 2678:
```suggestion
    async def deploy_to_be_hidden(name: str) -> bool:
```

vmarkovtsev (changes requested):

vmarkovtsev (approved):

PR #2643, opened 2022-08-04
[DEV-4721] preserve logical repos on reposet creation

gaetano-guerriero:


server/athenian/api/internal/reposet.py
```diff
@@ -354,6 +354,26 @@ async def load_prefixer():
...
+                    LogicalRepository.account_id == account,
+                ),
+            )
+            for name, physical_repo_id in logical_repos:
+                try:
+                    physical_name = prefixer.repo_node_to_prefixed_name[physical_repo_id]
+                except KeyError:
+                    log.error(
+                        "logical repo %s refers to not existing repo %d", name, physical_repo_id,
+                    )
```
vmarkovtsev, line 370: Let's accumulate all the missing logicals in a list and log them only once after the cycle.

vmarkovtsev (changes requested):

server/athenian/api/internal/reposet.py
gaetano-guerriero: done

vmarkovtsev (approved):

PR #2667, opened 2022-08-10
[DEV-4751] goal templates oa read operations

gaetano-guerriero:


server/athenian/api/align/goals/templates.py
```diff
@@ -1,37 +1,58 @@
+from __future__ import annotations
+
+from typing import TypedDict
+
 from athenian.api.models.web import JIRAMetricID, PullRequestMetricID, ReleaseMetricID
+from athenian.api.models.web.goal_template import GoalTemplateMetricID
 
-TEMPLATES_COLLECTION = {
+TEMPLATES_COLLECTION: dict[int, _TemplateDefinition] = {
```
vmarkovtsev, line 8:
```suggestion
TEMPLATES_COLLECTION: dict[int, TemplateDefinition] = {
```

Let's make it a fully-featured public class.

server/athenian/api/controllers/goal_controller.py
```diff
@@ -0,0 +1,40 @@
...
+    """Retrieve a goal template.
+
+    :param id: Numeric identifier of the goal template.
+    :type id: int
+
+    """
+    try:
+        template_def = TEMPLATES_COLLECTION[id]
+    except KeyError:
+        return ResponseError(NotFoundError("Template %d was not found." % id)).response
```
vmarkovtsev, line 22: Let's add a big TODO to check the account ID <> template ID access.

vmarkovtsev (approved):
I don't see anything critical, these are some points to consider

server/athenian/api/models/web/goal_template.py
```diff
@@ -0,0 +1,70 @@
...
+        """Gets the metric of this GoalTemplate."""
+        return self._metric
+
+    @metric.setter
+    def metric(self, metric: GoalTemplateMetricID) -> None:
+        """Sets the metric of this GoalTemplate."""
+        if metric is None:
+            raise ValueError("Invalid value for `metric`, must not be `None`")
+
+        self._metric = metric
```
vmarkovtsev, line 70: We must check if the value is in `JIRAMetricID | PullRequestMetricID | ReleaseMetricID`

server/athenian/api/controllers/goal_controller.py
gaetano-guerriero: Next PR will add retrieve from DB, and there we check account ownership

server/athenian/api/models/web/goal_template.py
gaetano-guerriero: Added.
I just noticed that typing with `base_model_.Enum` is not working as expected, i.e. `metric` is really string.
This is another global issue we must resolve before introducing mypy

vmarkovtsev: The string goes directly from aiohttp/connexion, the JSONSchema engine checks enums but doesn't do anything with types, I think. On the other hand, the code is simple now and it's important.

PR #2677, opened 2022-08-15
[DEV-4777] graphql goal metric name

gaetano-guerriero:


server/athenian/api/models/state/versions/bbafc5dfcf1d_goal_name_and_metric.py
```diff
@@ -0,0 +1,68 @@
...
+        metric = template["metric"]
+        conn.execute(
+            "UPDATE goals SET name = %s, metric = %s WHERE id = %s", name, metric, row["id"],
+        )
+
+    with op.batch_alter_table("goals") as bop:
+        bop.alter_column("name", nullable=False)
+        bop.alter_column("metric", nullable=False)
+        bop.create_unique_constraint(
+            "uc_goal_name", ["account_id", "name", "valid_from", "expires_at"],
```
vmarkovtsev, line 61: Why do we include `valid_from` and `expires_at` into the unique constraint?

vmarkovtsev (approved):

PR #2744, opened 2022-08-31
[DEV-4804] gql goals mutations add jira filters

gaetano-guerriero:


server/athenian/api/align/mutations/goals.py
```diff
@@ -259,11 +288,65 @@ def _ids_repr(ids: Sequence[int]) -> str:
...
+
+def _parse_request_jira_issue_types(input: dict[str, Any]) -> Optional[list[str]]:
+    if (issue_types := input.get(CreateGoalInputFields.jiraIssueTypes)) is None:
+        return None
+    else:
+        return _parse_request_jira_issue_types_value(issue_types)
+
+
+def _parse_request_jira_issue_types_value(value: list[str]) -> list[str]:
+    return list(dict.fromkeys([normalize_issue_type(t) for t in value]))
```
vmarkovtsev, line 352:
```suggestion
    return list({normalize_issue_type(t) for t in value})
```

server/athenian/api/align/mutations/goals.py
```diff
@@ -259,11 +288,65 @@ def _ids_repr(ids: Sequence[int]) -> str:
...
+
+def _parse_request_jira_priorities(input: dict[str, Any]) -> Optional[list[str]]:
+    if (priorities := input.get(CreateGoalInputFields.jiraPriorities)) is None:
+        return priorities
+    else:
+        return _parse_request_jira_priorities_value(priorities)
+
+
+def _parse_request_jira_priorities_value(value: list[str]) -> list[str]:
+    return list(dict.fromkeys([normalize_priority(p) for p in value]))
```
vmarkovtsev, line 341:
```suggestion
    return list({normalize_priority(p) for p in value})
```

server/athenian/api/models/web/base_model_.py
```diff
@@ -138,6 +138,9 @@ def __init__(cls, name, bases, dikt):
         """Initialize a new enumeration class type."""
         super().__init__(name, bases, dikt)
         cls.__members = set(v for k, v in dikt.items() if not k.startswith("__"))
+        cls.__members.update(
+            chain.from_iterable(getattr(base, "_Enum__members", ()) for base in bases),
```
vmarkovtsev, line 142: What if the base name is not "Enum"? It will be named `_BaseName__members`...

vmarkovtsev (changes requested):

server/athenian/api/align/mutations/goals.py
gaetano-guerriero: `fromkeys` was used to preserve while dropping duplicates, i.e. if the user selects [Tasks, Bugs] it is not good if he sees [Bugs,Tasks] after server round-trip

server/athenian/api/models/web/base_model_.py
gaetano-guerriero: since `__members` is only defined in Enum class definition it will always called this way, even if code is run by a subclass; so this works:
```
In [1]: from athenian.api.models.web.base_model_ import Enum

In [2]: class A(metaclass=Enum):
   ...:     a = '1'
   ...: class B(A):
   ...:     b = '2'
   ...: class C(B):
   ...:     c = '3'
   ...: list(C)
   ...: 
Out[2]: ['3', '2', '1']
```

server/athenian/api/align/mutations/goals.py
vmarkovtsev: We should sort the values alphabetically before returning them to the user, good observation. As for the original order, it's not important here.

gaetano-guerriero: changed to `sorted(set(...` on when writing

vmarkovtsev (approved):

PR #2747, opened 2022-09-01
[DEV-4919] drop custom cache in calc pr facts GitHub

gaetano-guerriero:


server/athenian/api/internal/features/entries.py
```diff
@@ -114,24 +114,14 @@
...
 unfresh_prs_threshold = 1000
 unfresh_participants_threshold = 50
 
 
-def _postprocess_cached_facts(
-    result: tuple[dict[str, list[PullRequestFacts]], bool],
-    with_jira_map: bool,
-    **_,
-) -> tuple[dict[str, list[PullRequestFacts]], bool]:
-    if with_jira_map and not result[1]:
```
gaetano-guerriero, line 130: not possible

vmarkovtsev (changes requested):
See DM in Slack:
It is possible to call that function with with_jira_map True or False and we don't include it in the cache key. Therefore,
- User requests with_jira_map=True and the cache has it with False => reject
- User requests with_jira_map=False and the cache has it with True => allow, leave the fetched jira stuff coz we don't care
- Otherwise, True-True and False-False are both OK

We use this in many cached functions across the project.

PR #2753, opened 2022-09-05
[DEV-4919] calc pr facts extend jira info

gaetano-guerriero:


server/athenian/api/internal/miners/jira/issue.py
```diff
@@ -812,74 +816,150 @@ def _validate_and_clean_issues(df: pd.DataFrame, acc_id: int) -> pd.DataFrame:
...
-                ),
-            )
-            .where(
-                sql.and_(
-                    node_id_cond,
-                    nprji.node_acc.in_(meta_ids),
-                ),
-            ),
+        cols = [nprji.node_id, Issue.key, Issue.priority_name, Issue.type, Issue.project_id]
+        from_clause = sa.outerjoin(
```
gaetano-guerriero, line 884: question: outerjoin here is maybe not needed ? we have a FK (nprji.jira_acc, nprji.jira_id) => (Issue.acc_id, Issue.id)

server/athenian/api/experiments/preloading/entries.py
```diff
@@ -49,7 +49,6 @@
     set_matched_by_from_release_match,
```
vmarkovtsev, line 49: FYI: the code inside `experiments` is legacy. We should keep it as-is. It's not used anywhere.

server/athenian/api/internal/features/entries.py
```diff
@@ -1690,7 +1688,6 @@ async def _calc_pull_request_facts_github(
         else:
             miner, unreleased_facts, matched_bys, unreleased_prs_event = await tasks[0]
         new_deps = miner.dfs.deployments.copy()
-        new_jira = miner.dfs.jiras.index.copy()
```
vmarkovtsev, line 1693: This df can mutate asynchronously. It is super important to make a copy here and use it later. I remember spending several hours on debugging the problem last year.

server/athenian/api/internal/features/entries.py
```diff
@@ -1772,20 +1769,16 @@ async def _calc_pull_request_facts_github(
...
-                    f.jira_ids = jira_map[node_id]
-                except KeyError:
-                    continue
+            jira_mappings = done_jira_map[0]
+            mined_jira_mappings = PullRequestJiraMapper.load_from_df(miner.dfs.jiras)
+            for pr_id, mined_jira_mapping in mined_jira_mappings.items():
+                if pr_id not in jira_mappings:
+                    jira_mappings[pr_id] = mined_jira_mapping
+                else:
+                    jira_mappings[pr_id] = jira_mappings[pr_id].merge(mined_jira_mapping)
```
vmarkovtsev, line 1778: This will be a performance bottleneck. Pandas does merges very slowly, we have to rewrite to pure numpy.

server/athenian/api/internal/miners/jira/issue.py
vmarkovtsev: As a rule of thumb, don't change queries because there is a high chance that they were optimized for performance. In this particular case, see "The hidden advantage of left join" in https://towardsdatascience.com/how-we-optimized-postgresql-queries-100x-ff52555eabe

server/athenian/api/internal/miners/jira/issue.py
```diff
@@ -812,74 +816,150 @@ def _validate_and_clean_issues(df: pd.DataFrame, acc_id: int) -> pd.DataFrame:
...
+
+    @classmethod
+    def load_from_df(self, df: pd.Dataframe) -> dict[int, PRJIRAMapping]:
+        """Load JIRA mappings from a Dataframe as found in PRDataFrames.jiras."""
+        mapping_builders: dict[int, _PRJIRAMappingBuilder] = defaultdict(_PRJIRAMappingBuilder)
+        for r in df.itertuples():
+            pr_id, issue_key = r[0]  # df index is (pr_id, issue_key)
+            mapping_builders[pr_id].import_values(issue_key, r.project_id, r.priority_name, r.type)
+        return {pr_id: builder.finalize() for pr_id, builder in mapping_builders.items()}
+
```
vmarkovtsev, line 904: OK, this is prohibitively slow and should be rewritten.
- itertuples is extremely slow
- Python object manipulation is slow
- this part of the code stand on the critical path and is performance critical

Uff, this is tough. All the code should be rewritten, but it's tricky to explain how.

vmarkovtsev (changes requested):
Let's discuss in DM

vmarkovtsev:
@gaetano-guerriero This is a raw draft of what I expect: https://github.com/athenianco/athenian-api/pull/2769 Can you please take that PR, polish the code, pass the CI, and finish it. Feel free to push straight there.

Important notes:
- Issue.id is S12 now - a collateral change.
- Issue.key is S24 - this will explode everything, must update the code where we expect strings. Better to change the other code to work with bytes vs. writing `.decode()` everywhere.
- Completely untested indirect unpacking in df_from_structs().

PR #2797, opened 2022-09-13
[DEV-4921] gql goals metrics fix repositories

gaetano-guerriero:
Fix GraphQL `goals()` metrics values retrieval with two team goals differing only for repositories.

### DIFF
--- a/server/athenian/api/align/queries/metrics.py
+++ b/server/athenian/api/align/queries/metrics.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
 import asyncio
+from collections import defaultdict
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from itertools import chain
-from typing import Any, Collection, Iterable, Mapping, Optional, Sequence, cast
+from typing import Any, Collection, Iterable, Mapping, Optional, Sequence
 
 import aiomcache
 from ariadne import ObjectType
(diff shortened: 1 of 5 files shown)


server/athenian/api/align/queries/metrics.py
```diff
@@ -280,64 +316,55 @@ async def calculate_team_metrics(
...
 
         if pr_metrics:
+            teams_filters = []
+            for (team_id, td), members in zip(request.teams.items(), team_members):
+                repos = _RepositoriesWrapper.from_requested_team_details(td, all_repos)
+                teams_filters.append(
+                    TeamSpecificFilters(
+                        team_id=team_id,
+                        participants={PRParticipationKind.AUTHOR: members},
+                        repositories=repos,
```
vmarkovtsev, line 326:
```suggestion
                        repositories=_RepositoriesWrapper.from_requested_team_details(td, all_repos),
```
(repos is a very common variable, don't want to pollute the locals)

server/athenian/api/align/queries/metrics.py
```diff
@@ -386,11 +413,11 @@ async def calculate_team_metrics(
...
     team_metrics_res: TeamMetricsResult = {}
     if pr_requests:
-        pr_collector.collect(calc_results.pop(0), team_metrics_res)
+        team_metrics_res.update(pr_collector.collect(calc_results.pop(0)))
     if release_requests:
-        release_collector.collect(calc_results.pop(0), team_metrics_res)
+        team_metrics_res.update(release_collector.collect(calc_results.pop(0)))
     if jira_requests:
-        jira_collector.collect(calc_results[0], team_metrics_res)
+        team_metrics_res.update(jira_collector.collect(calc_results[0]))
```
vmarkovtsev, line 421: Instead of calling one function, we now do 2x more operations and increase the cognitive load + the code complexity. It is better to return back the old code. Simpler >>> complex.

vmarkovtsev (changes requested):
I am not sure about the problem solved, but I don't like this refactoring at all so far: much more complex code, no obvious benefit.

server/athenian/api/align/queries/metrics.py
gaetano-guerriero: `collect` now is a pure function and this should be usually perceived as a huge code simplification. reverting.

PR #2807, opened 2022-09-15
[DEV-4922] gql goals query jira grouping

gaetano-guerriero:


server/athenian/api/align/queries/goals.py
```diff
@@ -66,14 +69,22 @@ async def resolve_goals(
 
     goals_to_serve = []
 
+    get_jira_config = _LazyCachedCoroutine(
+        get_jira_installation, accountId, info.context.sdb, info.context.mdb, info.context.cache,
+    )
```
vmarkovtsev, line 74: We should unconditionally load this earlier inside `gather()` at line 51.

vmarkovtsev (changes requested):
(unfinished)

vmarkovtsev (approved):
LGTM, I haven't found anything suspicious. Good job :+1:

PR #2844, opened 2022-09-23
[DEV-4987] gql goals openapi port

gaetano-guerriero:


server/athenian/api/controllers/goal_controller.py
```diff
@@ -6,17 +8,29 @@
...
     delete_goal_template_from_db,
     dump_goal_repositories,
+    fetch_team_goals,
     get_goal_template_from_db,
     get_goal_templates_from_db,
     insert_goal_template,
     parse_goal_repositories,
     update_goal_template_in_db,
 )
+from athenian.api.align.queries.goals import GoalToServe, GoalTreeGenerator
```
gaetano-guerriero, line 18: better to move relocate those when goals() query is removed

server/tests/align/queries/test_goals.py
```diff
@@ -541,6 +541,32 @@ async def test_same_team_different_repos(self, sdb: Database) -> None:
         assert (tg := goal_21["teamGoal"])["team"]["id"] == 1
         assert tg["value"]["initial"]["int"] == 0
 
+    @pytest.mark.xfail
+    async def test_team_unassigned_multiple_goals(self, sdb: Database) -> None:
```
gaetano-guerriero, line 545: this is really a corner case

server/athenian/api/controllers/goal_controller.py
```diff
@@ -168,4 +182,60 @@ async def parse_request_repositories(
...
+        get_jira_installation_or_none(
+            goals_request.account, request.sdb, request.mdb, request.cache,
+        ),
+    )
+    team_tree = build_team_tree_from_rows(team_rows, None if team == 0 else team)
+    team_member_map = flatten_teams(team_rows)
+
+    team_ids = [row[Team.id.name] for row in team_rows]
+    team_goal_rows, prefixer = await gather(
+        fetch_team_goals(goals_request.account, team_ids, request.sdb),
```
vmarkovtsev, line 205: Please add a comment in `fetch_team_goals` that the order is mandatory. Otherwise, `groupby` will silently betray us.

server/athenian/api/controllers/goal_controller.py
```diff
@@ -168,4 +182,60 @@ async def parse_request_repositories(
...
+        pdb=request.pdb,
+        rdb=request.rdb,
+        cache=request.cache,
+        slack=request.app["slack"],
+        unchecked_jira_config=jira_config,
+    )
+
+    goal_tree_generator = GoalTreeGenerator()
+    models = [
+        to_serve.build_goal_tree(all_metric_values, goal_tree_generator).to_dict()
```
vmarkovtsev, line 238: `to_dict()` should always happen inside `model_response()`. It is easier to tune serialization in the future this way.

server/athenian/api/typing_utils.py
```diff
@@ -89,9 +89,7 @@ def is_union(klass: object) -> bool:
 
 def is_optional(klass: type):
     """Determine whether klass is an Optional."""
-    return (
-        is_union(klass) and len(klass.__args__) == 2 and issubclass(klass.__args__[1], type(None))
-    )
+    return is_union(klass) and types.NoneType in klass.__args__
```
vmarkovtsev, line 92: The fresh native code I added in `web_model_io.pyx` has a hard requirement for `None` to come second and uses this function. We should check both the expected args length (2) and that None comes second.

server/tests/align/queries/test_goals.py
vmarkovtsev: To be fixed next week, right?

vmarkovtsev (changes requested):

server/athenian/api/typing_utils.py
gaetano-guerriero: Reverted,
This is needed for `Optional[int | float]` (or `Optional[Union[int, float]]`) to work, I solved this with a workaround

vmarkovtsev (approved):

PR #2856, opened 2022-09-27
[DEV-5035] measure goals add timeseries

gaetano-guerriero:


server/athenian/api/align/queries/goals.py
```diff
@@ -124,14 +136,21 @@ def __init__(
...
         jira_config: Optional[JIRAConfig],
+        only_with_targets: bool,
+        include_series: bool,
     ):
         """Init the GoalToServe object."""
         self._team_goal_rows = team_goal_rows
         self._prefixer = prefixer
-        self._request, self._goal_team_tree = self._team_goal_rows_to_request(
-            team_goal_rows, team_tree, team_member_map, prefixer, jira_config, only_with_targets,
+        self._request, self._goal_team_tree, self._timeseries_spec = self._parse_rows(
```
vmarkovtsev, line 146:
```suggestion
        self._request, self._goal_team_tree, self._timeseries_spec = self._parse_team_goal_rows(
```
looks more obvious

server/athenian/api/models/web/goal.py
```diff
@@ -25,14 +26,23 @@ class GoalSeriesGranularity(Enum):
...
+    value: VerbatimOptional["MetricValue"]
+
+
 class GoalValue(Model):
     """The current metric values for the goal on a team."""
 
-    # do NOT change the wrapped types for Optional fields, base Model doesn't support
-    # wrapping an union with an optional format
+    # do NOT change the wrapped type "MetricValue" to a direct reference,
+    # base Model doesn't support wrapping an union with an optional form
```
vmarkovtsev, line 40:
```suggestion
    # base Model doesn't support wrapping a union with an optional form
```

server/athenian/api/serialization.py
```diff
@@ -251,6 +252,8 @@ def serialize(klass, obj):
             return obj.tolist()
         if isinstance(obj, (bytes, bytearray)):
             return obj.decode()
+        if isinstance(obj, enum.Enum):
+            return obj.value
```
vmarkovtsev, line 256: Our models cannot support enums due to the planned future native code, we should work with strings here. Take `.value` in the endpoint code pls.

server/athenian/api/align/queries/goals.py
```diff
@@ -244,20 +272,27 @@ def _team_goal_rows_to_request(
                 ),
             )
 
+        time_intervals = [initial_interval, (valid_from, expires_at)]
```
vmarkovtsev, line 275:
```suggestion
        time_intervals = [initial_interval, [valid_from, expires_at]]
```
so that we have lists everywhere

vmarkovtsev (changes requested):

server/athenian/api/align/queries/goals.py
gaetano-guerriero: we need `Intervals` (tuple) as elements of time_intervals since they will be used as keys in `TeamMetricsResult`

vmarkovtsev: OK, so we have tuples everywhere.

vmarkovtsev (approved):

PR #2875, opened 2022-10-04
[DEV-5107] web model io fix multibyte unicodestring

gaetano-guerriero:
Fix model string handling in web_model_io when python unicode representation has more than one byte per code point


vmarkovtsev (approved):
Good point :+1:

PR #2878, opened 2022-10-05
[DEV-5084] fix mine env logical repo owning no deployed components

gaetano-guerriero:


server/athenian/api/internal/miners/github/deployment_light.py
```diff
@@ -565,7 +565,7 @@ async def mine_environments(
     envs_col = components[DeploymentNotification.environment.name].values
     order = np.argsort(envs_col)
     repo_unique_envs, repo_group_counts = np.unique(envs_col[order], return_counts=True)
-    assert (repo_unique_envs == unique_envs).all()
+    assert list(repo_unique_envs) == list(unique_envs)
```
gaetano-guerriero, line 568: the fix in `split_logical_deployed_components` is enough to avoid the Sentry error, but this is more reliable given that

```
In [12]: np.array(['1', '2'], dtype=object) == np.array([], dtype=object)
<ipython-input-12-1c78974ddc15>:1: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
  np.array(['1', '2'], dtype=object) == np.array([], dtype=object)
Out[12]: False
```

vmarkovtsev: We can also use `np.testing.assert_array_equal()`

vmarkovtsev (approved):

PR #2907, opened 2022-10-11
[DEV-5044] dockerfile non root user

gaetano-guerriero:


Dockerfile
```diff
@@ -70,7 +71,12 @@ RUN apt-get update && \
     apt-key add - && \
     apt-get update && \
     apt-get install -y --no-install-suggests --no-install-recommends intel-mkl-common-c-$MKL intel-mkl-gnu-rt-$MKL intel-mkl-f95-$MKL && \
-    rm -rf /opt/intel/documentation_2019 /opt/intel/compilers_and_libraries_*/linux/mkl/{bin,tools,examples} && \
```
gaetano-guerriero, line 73: dash does not expand {..,..}

Dockerfile
```diff
@@ -115,30 +121,36 @@ lapack_libs = mkl_lapack95_lp64' >/root/.numpy-site.cfg && \
...
 ARG GKWILLIE_TOKEN
 RUN apt-get update && \
     apt-get install -y --no-install-suggests --no-install-recommends gcc g++ patch && \
     sed -i "s/git+ssh:\/\/git@/git+https:\/\/gkwillie:$GKWILLIE_TOKEN@/g" server/requirements.txt && \
+    echo "Installing Python packages" && \
     pip3 install --no-cache-dir -r /server/requirements.txt && \
     sed -i "s/git+https:\/\/gkwillie:$GKWILLIE_TOKEN@/git+ssh:\/\/git@/g" server/requirements.txt && \
     pip3 uninstall -y flask && \
     patch /usr/local/lib/python*/dist-packages/prometheus_client/exposition.py /patches/prometheus_client.patch && \
+    chown -R worker:worker /usr/local/lib/python* && \
```
vmarkovtsev, line 139: Doesn't this reduce security? How about adding the user in the end?

gaetano-guerriero: Maybe we can also avoid this chown, installed packages should have pyc files, and we are not interested in worker user creating them anyway.
What do you mean by adding the user in the end anyway ?

vmarkovtsev: I mean leave everything as-is and execute `adduser` right before `USER` and `ENTRYPOINT`.

btw, we should test `sudo` or `su` with that new user, otherwise, we cannot sacrifice debuggability.

gaetano-guerriero: We need write permission at least in `/server/` to test migrations I think.
To debug from outside it should be ok to do `docker exec --user root` so enabling sudo is not needed

vmarkovtsev: Na-nah, we are under kubernetes in GKE. What works in docker will not work there.

vmarkovtsev: The topic is live debugging in production, which I had to do numerous times and which is the most important property we cannot lose.

vmarkovtsev: Regarding the write permission in `/server`, we can make it more fine-grained in just those directories that require RW.

gaetano-guerriero: Right, `kubectl exec` doesn't allow to propagate `-u`
https://github.com/kubernetes/kubectl/blob/a0af655b7abaf06983c99ad5a4dc8f02ae5eb3e5/pkg/cmd/exec/exec.go#L100

gaetano-guerriero: Maybe we could set a root password ?

.github/workflows/push.yml
```diff
@@ -436,23 +436,27 @@ jobs:
...
         $run combine
         $run xml
         docker rmi test
-        sed -i 's/[\./]server[\./]//g' coverage.xml
-        sed -i "s\<source></source>\<source>$GITHUB_WORKSPACE/server</source>\g" coverage.xml
-        rm .coverage
+        sudo chown $(whoami) -R /tmp/io
+        sed -i 's/[\./]server[\./]//g' /tmp/io/coverage.xml
+        sed -i "s\<source></source>\<source>$GITHUB_WORKSPACE/server</source>\g" /tmp/io/coverage.xml
+        rm /tmp/io/.coverage
```
vmarkovtsev, line 454: This is not needed anymore, I had to delete it because it overrode `coverage.xml`.

vmarkovtsev (approved):
LGTM, :crossed_fingers:

.github/workflows/push.yml
gaetano-guerriero: removed

.github/workflows/push.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches:
     - master
-  pull_request_target:
+  pull_request:
```
gaetano-guerriero, line 7: @vmarkovtsev how was the workflow here ?
Check that CI is green, and then revert before merge ?

vmarkovtsev: It is needed to build dependabot, so we can return back in the next PR.

PR #2934, opened 2022-10-21
[DEV-5153] port gql remove goal mutation to openapi

gaetano-guerriero:


server/tests/controllers/goal_controller/test_goal.py
```diff
@@ -0,0 +1,94 @@
+"""Tests for the goal CRUD controllers."""
```
gaetano-guerriero, line 1: those are the new tests, other changes in tests  it's just splitting from single module to package

vmarkovtsev (approved):

PR #2964, opened 2022-10-28
Increase precomputer timeout for a single account

gaetano-guerriero:


server/athenian/api/precompute/__main__.py
```diff
@@ -127,7 +127,7 @@ class Formatter(argparse.ArgumentDefaultsHelpFormatter, argparse.RawTextHelpForm
         help="Do not sandbox each account in a separate process",
     )
     accounts_parser.add_argument(
-        "--timeout", type=int, default=20 * 60, help="Maximum processing time for one account",
+        "--timeout", type=int, default=60 * 60, help="Maximum processing time for one account",
```
vmarkovtsev, line 130: This is for every account; we need to increase only new freshly installed accounts (accounts-new job)

vmarkovtsev (changes requested):

PR #2980, opened 2022-11-01
[DEV-5225] new goal templates

gaetano-guerriero:


server/athenian/api/align/goals/dbaccess.py
```diff
@@ -265,11 +265,22 @@ async def get_goal_template_from_db(template_id: int, sdb: DatabaseLike) -> Row:
...
+    account: int,
+    sdb: DatabaseLike,
+    exclude_with_metric_params: bool,
+) -> list[Row]:
     """Return all account GoalTemplate-s, ordered by id."""
-    stmt = (
-        sa.select(GoalTemplate).where(GoalTemplate.account_id == account).order_by(GoalTemplate.id)
-    )
+    where = GoalTemplate.account_id == account
+    from sqlalchemy import JSON
```
vmarkovtsev, line 275: Let's import everything in the header

server/athenian/api/align/goals/dbaccess.py
```diff
@@ -265,11 +265,22 @@ async def get_goal_template_from_db(template_id: int, sdb: DatabaseLike) -> Row:
...
+
+    if exclude_with_metric_params:
+        # select both SQL NULL and JSON null
+        where = sa.and_(
+            where,
+            sa.or_(GoalTemplate.metric_params.is_(None), GoalTemplate.metric_params == JSON.NULL),
```
vmarkovtsev, line 281: We cannot use `OR` in Postgres queries in production because that defeats the indexes. Switch to `UNION` or better `UNION ALL`.

server/athenian/api/align/goals/dbaccess.py
```diff
@@ -265,11 +265,22 @@ async def get_goal_template_from_db(template_id: int, sdb: DatabaseLike) -> Row:
...
+        )
+    stmt = sa.select(GoalTemplate).where(where).order_by(GoalTemplate.id)
```
vmarkovtsev, line 283: Note: `ORDER BY` doesn't work with `UNION [ALL]` directly, this should save you some debugging time :slightly_smiling_face:

server/athenian/api/models/state/readd_goal_templates.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python
```
vmarkovtsev, line 1: You named the script "reaDD", the second D should be a typo.

server/tests/testutils/db.py
```diff
@@ -41,6 +42,14 @@ async def models_insert_auto_pk(db: Database, *models: BaseType) -> list[Any]:
     return results
 
 
+SKIP_MODEL_FIELD = object()
+"""Set a model field to this value avoid propagating it to DB on model insertion.
+
+For instance, this is needed to insert a SQL NULL into a JSONB column.
+
```
vmarkovtsev, line 49:
```suggestion
```
(save vertical space)

vmarkovtsev (changes requested):

server/athenian/api/models/state/readd_goal_templates.py
gaetano-guerriero: Right, 
I renamed to `reimport_goal_templates` seems more correct

server/athenian/api/align/goals/dbaccess.py
```diff
@@ -265,11 +265,22 @@ async def get_goal_template_from_db(template_id: int, sdb: DatabaseLike) -> Row:
     return template
 
 
-async def get_goal_templates_from_db(account: int, sdb: DatabaseLike) -> list[Row]:
+async def get_goal_templates_from_db(
+    account: int,
+    sdb: DatabaseLike,
+    exclude_with_metric_params: bool,
```
vmarkovtsev, line 271:
```suggestion
    exclude_with_metric_params: bool,
    sdb: DatabaseLike,
```
we tend to specify the DB args last, unless there are optional kwargs

server/tests/testutils/db.py
```diff
@@ -129,7 +138,8 @@ async def transaction_conn(db: Database) -> AsyncIterator[Connection]:
...
 
 def _build_table_where_clause(table: DeclarativeMeta, **kwargs: Any) -> ClauseElement:
-    return cast(
-        ClauseElement,
-        sa.and_(*(getattr(table, key) == val for key, val in kwargs.items())),
-    )
+    if kwargs:
+        where = sa.and_(*(getattr(table, key) == val for key, val in kwargs.items()))
+    else:
+        where = sa.and_(True)
```
vmarkovtsev, line 144: It's better to append to a specified list of filters from the function arguments instead of returning these hacks.

vmarkovtsev (changes requested):

server/tests/testutils/db.py
gaetano-guerriero: changed, now caller checks that kwargs is not empty

server/athenian/api/align/goals/dbaccess.py
gaetano-guerriero: Until we don't have an index (account_id, metric_params) in this table it should be ok

```
> explain analyze select * from goal_templates where account_id = 1 AND (goal_templates.metric_params = 'null'::jsonb OR goal_templates.metric_params IS NULL);
                                                                   QUERY PLAN                                                                    
-------------------------------------------------------------------------------------------------------------------------------------------------
 Index Scan using ix_goal_templates_account_id on goal_templates  (cost=0.28..6.05 rows=11 width=262) (actual time=0.019..0.025 rows=11 loops=1)
   Index Cond: (account_id = 1)
   Filter: ((metric_params = 'null'::jsonb) OR (metric_params IS NULL))
 Planning Time: 0.136 ms
 Execution Time: 0.044 ms
(5 rows)
```

is no different than

```
tate=> explain analyze select * from goal_templates where account_id = 1 AND goal_templates.metric_params = 'null'::jsonb;
                                                                  QUERY PLAN                                                                   
-----------------------------------------------------------------------------------------------------------------------------------------------
 Index Scan using ix_goal_templates_account_id on goal_templates  (cost=0.28..6.05 rows=1 width=262) (actual time=0.027..0.027 rows=0 loops=1)
   Index Cond: (account_id = 1)
   Filter: (metric_params = 'null'::jsonb)
   Rows Removed by Filter: 11
 Planning Time: 1.097 ms
 Execution Time: 0.091 ms
(6 rows)
```

After using the index on account_id there will always be 18 rows, the number of templates per account is locked for now.

Also this code will disappear in about one month, for sure before we will add ability for the user to add templates.

vmarkovtsev: Wow, it's good that you posted the plans. `metric_params = 'null'::jsonb` is a big no-no to me. We should never write nulls to JSON fields, this is a straight road to bugs and code complexities. Either a DB null or an empty dictionary/object, please. We should put as many constraints everywhere as possible. In fact, I don't like that we allow nullable JSON at all because it's a well-known (and well-proven by my own experience) DB usage antipattern. I remember that we had to do it, though, for the repositories filter, so it's too late.

Let's edit the code to ensure that this never happens, and therefore remove `OR` thanks to the new invariant.

gaetano-guerriero: It turned out asyncpg never writes 'null'::jsonb where it receives None
aiosqlite always writes json null so I added the OR only for sqlite

server/athenian/api/align/goals/dbaccess.py
```diff
@@ -265,11 +266,20 @@ async def get_goal_template_from_db(template_id: int, sdb: DatabaseLike) -> Row:
...
+) -> list[Row]:
     """Return all account GoalTemplate-s, ordered by id."""
-    stmt = (
-        sa.select(GoalTemplate).where(GoalTemplate.account_id == account).order_by(GoalTemplate.id)
-    )
+    where = GoalTemplate.account_id == account
+    if exclude_with_metric_params:
+        exclude_expr = GoalTemplate.metric_params.is_(None)
+        # with sqlite both SQL NULL and JSON null can be found on db
+        if not await is_postgresql(sdb):
```
vmarkovtsev, line 279: f*cking sqlite

vmarkovtsev (approved):

PR #2988, opened 2022-11-03
[DEV-5252] Remove repository names from the reposet

vmarkovtsev:


vmarkovtsev:
Data migration is TODO

server/athenian/api/precompute/accounts.py
```diff
@@ -203,16 +203,15 @@ async def precompute_reposet(
...
             Prefixer.load(meta_ids, mdb, cache),
             fetch_bots(reposet.owner_id, meta_ids, mdb, sdb, None),
-            refresh_repository_names(reposet.owner_id, meta_ids, sdb, mdb, None),
         )
     except Exception as e:
         log.error("prolog %d: %s: %s", reposet.owner_id, type(e).__name__, e)
         sentry_sdk.capture_exception(e)
         return
-    reposet.items = new_items
+    reposet.items = prefixer.dereference_repositories(reposet_items_to_refs(reposet.items))
```
gaetano-guerriero, line 214: expression has a different type than `reposet.items` and if I'm not wrong `reposet.items` is  not saved after.
Could we just use a separate variable ?

gaetano-guerriero:
Is `reposet-sync` ready ? It should handle also the new format on some events. And maybe ignore rename repo event.

server/athenian/api/precompute/accounts.py
vmarkovtsev: Sure, this is just some legacy of how it used to work.

vmarkovtsev:
reposet-sync is TODO. Any ideas on how to deploy everything? We seem to need deploy both at once, but that requires a well-tested data migration. Luckily, it should be a matter of extracting logical names.

There is one thing that we are losing with this change, however. If GitLab and GitHub are separate databases, we won't be able to understand where each node ID points at. Maybe we should extend the format to save the prefix, hmmm

vmarkovtsev:
btw I am planning a follow-up refactoring to work with `RepositoryName` everywhere. We have got a few places where we are doing wild things like node_id -> name -> node_id -> name.

gaetano-guerriero:
0 downtime is hard with two services:
A) api read both formats and writes old format
B) reposet-sync read both and writes new format
C) api read both and writes new
D) migrate
E) api and reposet-sync read only new

gaetano-guerriero:
yes with multiple providers `RepositoryReference` will be (provider: str, repo_id: int, logical_name: str)

vmarkovtsev:
Let's put reposet-sync to some downtime then, we are still small to afford this. I'll try to finish before the weekend so that we can migrate on Saturday.

PR #3013, opened 2022-11-08
[DEV-5152] openapi gql update goal mutation

gaetano-guerriero:


server/tests/controllers/test_goal_controller.py
```diff
@@ -1,341 +0,0 @@
-from typing import Any
```
gaetano-guerriero, line 1: wrongly left after test module was split into a package

vmarkovtsev (approved):

PR #3016, opened 2022-11-09
[DEV-5325] Avoid zero division in ThresholdComparisonRatioCalculator

gaetano-guerriero:


server/athenian/api/internal/features/metric_calculator.py
```diff
@@ -712,9 +712,10 @@ def _analyze(
         return out
 
     def _agg(self, samples: np.ndarray, offsets: np.ndarray) -> float:
-        lengths = np.diff(offsets.base)
+        lengths = np.diff(offsets.base).astype(float)
+        lengths[lengths == 0] = np.nan
         vec = np.add.reduceat(samples, offsets) / lengths
-        vec[lengths == 0] = 0  # fix NaNs + reduceat copies the previous value
+        vec[np.isnan(vec)] = 0  # fix NaNs + reduceat copies the previous value
```
vmarkovtsev, line 718: Why doesn't the original mask work? This is important because the new code is less efficient.

vmarkovtsev (changes requested):

server/athenian/api/internal/features/metric_calculator.py
gaetano-guerriero: tried with a simple mask

server/athenian/api/internal/features/metric_calculator.py
```diff
@@ -712,9 +712,11 @@ def _analyze(
         return out
 
     def _agg(self, samples: np.ndarray, offsets: np.ndarray) -> float:
-        lengths = np.diff(offsets.base)
+        lengths = np.diff(offsets.base).astype(float)
+        nolen_mask = lengths == 0
+        lengths[nolen_mask] = np.nan
```
vmarkovtsev, line 717: This should not be needed then

server/athenian/api/internal/features/metric_calculator.py
```diff
@@ -712,9 +712,11 @@ def _analyze(
...
         vec = np.add.reduceat(samples, offsets) / lengths
-        vec[lengths == 0] = 0  # fix NaNs + reduceat copies the previous value
+        vec[nolen_mask] = 0  # fix NaNs + reduceat copies the previous value
```
vmarkovtsev, line 719:
```suggestion
        vec[lengths == 0] = 0  # fix NaNs + reduceat copies the previous value
```

server/athenian/api/internal/features/metric_calculator.py
```diff
@@ -712,9 +712,11 @@ def _analyze(
         return out
 
     def _agg(self, samples: np.ndarray, offsets: np.ndarray) -> float:
-        lengths = np.diff(offsets.base)
+        lengths = np.diff(offsets.base).astype(float)
```
vmarkovtsev, line 715: It seems that casting to float is the only required change

vmarkovtsev (changes requested):

server/athenian/api/internal/features/metric_calculator.py
gaetano-guerriero: we have `np.seterr(divide="raise")` so `np.array([1]) / np.array([0])` crashes
offsets is not strictly increasing so lengths includes zeros if I understood correctly when there is a group selecting no facts and it's not the last one. 
Is this faster maybe ?
```
        lenghts = np.diff(offsets.base)
        reduced_samples = np.add.reduceat(samples, offsets)
        # fix NaNs + reduceat copies the previous value
        vec = np.zeros(reduced_samples.shape, dtype=self.dtype)
        np.divide(reduced_samples, lenghts, where=lenghts != 0, out=vec)
        return vec
```

vmarkovtsev: This is perfect 👍

gaetano-guerriero: pushed this version

vmarkovtsev (approved):

PR #3020, opened 2022-11-10
[DEV-5150] openapi align list team members

gaetano-guerriero:


server/athenian/api/internal/team.py
```diff
@@ -77,9 +77,12 @@ async def get_root_team(account_id: int, sdb_conn: DatabaseLike) -> Row:
 
 
 @sentry_span
-async def get_team_from_db(account_id: int, team_id: int, sdb_conn: DatabaseLike) -> Row:
-    """Return a team owned by an account."""
-    stmt = sa.select(Team).where(sa.and_(Team.owner_id == account_id, Team.id == team_id))
+async def get_team_from_db(account_id: Optional[int], team_id: int, sdb_conn: DatabaseLike) -> Row:
+    """Return a team by ID, optionally filtered by owning account."""
+    where = Team.id == team_id
```
vmarkovtsev, line 82: We have a potential security vulnerability here. Someone who is not you or me will call this function with `account_id` set to `None` and will forget to check the owner.
How about requiring the `user_id` or `account_id`. If the caller sets `user_id`, join `teams` with `account_users` by account ID.

vmarkovtsev (changes requested):

server/athenian/api/internal/team.py
gaetano-guerriero: right this way is safer
I implemented also filtering by both, this is needed to return 404 when asking for `/members/99?account=3`, the user is allowed for account 3 but team 99 belongs to another account.

vmarkovtsev (approved):

PR #3036, opened 2022-11-15
[DEV-4948] Add TLO template names

warenlg:
Make the default TLO template goals user-friendly with proper names. I also removed 2 minor TLO template goals form the default config that are minor, to reduce the noise in the UI for now.

Signed-off-by: Waren Long <waren@athenian.co>


gaetano-guerriero (approved):
Let's remember both renames and deletions must be applied manually to state DBs in staging and production

vmarkovtsev:
@gaetano-guerriero you are the owner here, please merge if everything's fine.

warenlg:
@gaetano-guerriero why do you say this? What would happen if we don't edit the state DB accordingly?

PR #3039, opened 2022-11-15
[DEV-5347] Extract PRFactsCalculator from MetricEntriesCalculator

gaetano-guerriero:
Extract `PRFactsCalculator` from `MetricsCalculator`, so that only `PRFactsCalculator` is built when you only need to call `calc_pull_request_facts_github`

### DIFF
--- a/server/athenian/api/internal/features/entries.py
+++ b/server/athenian/api/internal/features/entries.py
@@ -9,7 +9,7 @@ from itertools import chain
 import logging
 import operator
 import pickle
-from typing import Any, Collection, Generic, Iterable, Mapping, Optional, Sequence, TypeVar
+from typing import Any, Collection, Generic, Iterable, Mapping, Optional, Sequence, Type, TypeVar
 
 import aiomcache
 import numpy as np
@@ -340,7 +340,7 @@ class MetricEntriesCalculator:
         """
         assert isinstance(repositories, (tuple, list))
         all_repositories = set(chain.from_iterable(repositories))
-        all_participants = self._merge_pr_participants(participants)
+        all_participants = _merge_pr_participants(participants)
         calc = PullRequestBinnedMetricCalculator(
             metrics,
             quantiles,
@@ -349,7 +349,8 @@ class MetricEntriesCalculator:
             environments=environments,
         )
         time_from, time_to = self._align_time_min_max(time_intervals, quantiles)
-        df_facts = await self.calc_pull_request_facts_github(
+        pr_facts_calc = self._build_pr_facts_calculator()
+        df_facts = await pr_facts_calc(
             time_from,
             time_to,
             all_repositories,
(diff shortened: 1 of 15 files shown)


server/athenian/api/internal/features/entries.py
```diff
@@ -1861,156 +2061,9 @@ async def _calc_pull_request_facts_github(
...
-        """Format the filters as a stable string."""
-        # participants are fully defined by the team ID
-        return f"{self.team_id}|{','.join(self.repositories)}|{self.jira_filter}"
-
-    def __sentry_repr__(self):
-        """Format for Sentry."""
-        dikt = dataclasses.asdict(self)
-        del dikt["participants"]  # fully defined by the team ID
-        return str(dikt)
+def _merge_pr_participants(participants: Iterable[PRParticipants]) -> PRParticipants:
```
vmarkovtsev, line 2064: Can you please move this function under some class, so that we don't pollute the global scope.

server/athenian/api/internal/features/entries.py
```diff
@@ -1516,6 +1511,211 @@ def check_runs_grouper(df: pd.DataFrame) -> list[np.ndarray]:
...
+        self._pdb = pdb
+        self._rdb = rdb
+        self._pr_miner = pr_miner
+        self._branch_miner = branch_miner
+        self._done_prs_facts_loader = done_prs_facts_loader
+        self._unfresh_pr_facts_fetcher = unfresh_pr_facts_fetcher
+        self._pr_jira_mapper = pr_jira_mapper
+        self._cache = cache
+
     async def calc_pull_request_facts_github(
```
vmarkovtsev, line 1719:
```suggestion
    async def __call__(
```
This is logical... This is the only thing this class is supposed to do.

server/athenian/api/internal/features/entries.py
```diff
@@ -550,14 +544,15 @@ async def calc_pull_request_histograms_github(
...
-        all_participants = self._merge_pr_participants(participants)
+        all_participants = _merge_pr_participants(participants)
         try:
             calc = PullRequestBinnedHistogramCalculator(
                 defs.values(), quantiles, environments=[environment],
             )
         except KeyError as e:
             raise UnsupportedMetricError() from e
-        df_facts = await self.calc_pull_request_facts_github(
+        pr_facts_calc = self._build_pr_facts_falculator()
```
vmarkovtsev, line 554: typo: `falculator`

vmarkovtsev (changes requested):

server/athenian/api/internal/features/entries.py
gaetano-guerriero: moved to a new class
https://github.com/athenianco/athenian-api/pull/3039/commits/faec3c7653d9c9c0fdff06b27eb543ce06c7dba8

vmarkovtsev (approved):

PR #3042, opened 2022-11-16
[DEV-5347] prs node id to number

gaetano-guerriero:
Implement the `prs_node_id_to_number` function needed by `search_prs_controller`
Also a little refactor in `filter_controller` to reuse an already existing repo name transformation from prefixer

### DIFF
--- a/server/athenian/api/controllers/filter_controller.py
+++ b/server/athenian/api/controllers/filter_controller.py
@@ -28,7 +28,7 @@ from athenian.api.internal.jira import (
     get_jira_installation_or_none,
     load_mapped_jira_users,
 )
-from athenian.api.internal.logical_repos import coerce_logical_repos, drop_logical_repo
+from athenian.api.internal.logical_repos import coerce_logical_repos
 from athenian.api.internal.miners.access_classes import access_classes
 from athenian.api.internal.miners.filters import JIRAFilter, LabelFilter
 from athenian.api.internal.miners.github.bots import bots
@@ -453,15 +453,9 @@ def web_pr_from_struct(
     for pr in prs:
         props = dict(dataclass_asdict(pr))
         del props["node_id"]
-        repo = props["repository"]
-        physical_repo = drop_logical_repo(repo)
-        if physical_repo != repo:
-            append = repo[len(physical_repo) :]
-        else:
-            append = ""
-        try:
-            props["repository"] = prefixer.repo_name_to_prefixed_name[physical_repo] + append
-        except KeyError:
+
+        props["repository"] = prefixer.prefix_logical_repo(props["repository"])
+        if props["repository"] is None:
             # deleted repository
             continue
         if pr.events_time_machine is not None:
--- a/server/athenian/api/internal/prefixer.py
+++ b/server/athenian/api/internal/prefixer.py
@@ -204,12 +204,12 @@ class Prefixer:
         ]
 
     def prefix_logical_repo(self, repo: str) -> Optional[str]:
-        """Lookup the repository name prefix for the given logical repository."""
-        *physical_repo, logical_name = repo.split("/", 2)
+        """Lookup the repository name prefix for the given repository, logical or not."""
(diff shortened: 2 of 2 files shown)


server/athenian/api/internal/miners/github/pull_request.py
```diff
@@ -3103,3 +3106,28 @@ def _validate(self, facts: PullRequestFacts, url: str) -> None:
...
+    """
+    selects = []
+    node_id_cond, hints = column_values_condition(NodePullRequest.node_id, node_ids)
+    selects = [
+        sa.select(NodePullRequest.node_id, NodePullRequest.number).where(
+            NodePullRequest.acc_id == meta_id, node_id_cond,
+        )
+        for meta_id in meta_ids
+    ]
+    stmt = sa.union_all(*selects)
```
vmarkovtsev, line 3128:
```suggestion
    stmt = selects[0] if len(selects) == 1 else sa.union_all(*selects)
```
(this used to generate broken code sometimes, maybe SA fixed it but IDK. anyways, a bit simpler from the logic PoV)

server/athenian/api/internal/miners/github/pull_request.py
```diff
@@ -3103,3 +3106,28 @@ def _validate(self, facts: PullRequestFacts, url: str) -> None:
...
+
+    for hint in hints:
+        stmt = stmt.with_statement_hint(hint)
+    rows = await mdb.fetch_all(stmt)
```
vmarkovtsev, line 3132: Let's use `read_sql_query` with defined column dtypes. It will work faster.

server/athenian/api/internal/miners/github/pull_request.py
```diff
@@ -3103,3 +3106,28 @@ def _validate(self, facts: PullRequestFacts, url: str) -> None:
...
+    return dict(rows)
```
vmarkovtsev, line 3133: therefore,
```suggestion
    return dict(df[NodePullRequest.node_id.name].values, df[NodePullRequest.number.name].values)
```

vmarkovtsev (changes requested):

server/athenian/api/internal/miners/github/pull_request.py
gaetano-guerriero: yes SA is already doing the expected thing here, so it's a matter of preference

server/athenian/api/internal/miners/github/pull_request.py
gaetano-guerriero: do you mean dict(zip(... values, ...values)) ?
I've benchmarked this locally with ~600 rows and
```
async def simple(mdb):
    rows = await mdb.fetch_all(stmt)
    return dict(rows)
```
outperforms
```
async def complex_(mdb):
    df = await read_sql_query(stmt, mdb, columns)
    return dict(
        zip(df[NodePullRequest.node_id.name].values, df[NodePullRequest.number.name].values),
    )
```
7.5 ms simple, 8.5 ms complex.
I've also tried `return dict(df.itertuples(index=False, name=None))`
Is there maybe a faster  way to do the df => dict conversion ?

vmarkovtsev: Nah, local benchmarks are all lies. The vCPUs we are using in GCP make the profile much more skewed toward CPU. As a rule of thumb, if the returned columns are types, always try to prefer `read_sql_query` instead of the vanilla asyncpg rows.

vmarkovtsev: BTW, I've just realized: you shouldn't make the dict at all! The logic must be vectorized downstream using `np.searchsorted`

vmarkovtsev: Another rule of thumb: avoid Python objects like a plague, preferring vectorized numpy arrays whereever possible.

vmarkovtsev (approved):

PR #3072, opened 2022-11-24
[DEV-5347] search prs basic

gaetano-guerriero:
The "basic" implementation of /search/pull_requests.
Propagate already supported filters to PRFactsCalculator and compose the result.

### DIFF
--- a/server/athenian/api/controllers/search_controller/search_prs.py
+++ b/server/athenian/api/controllers/search_controller/search_prs.py
@@ -1,0 +1,222 @@
+from __future__ import annotations
+
+from collections.abc import Collection
+import dataclasses
+from datetime import datetime, timedelta, timezone
+import logging
+from typing import Optional
+
+from aiohttp import web
+import aiomcache
+import sentry_sdk
+
+from athenian.api.async_utils import gather
+from athenian.api.db import DatabaseLike
+from athenian.api.internal.account import get_metadata_account_ids
+from athenian.api.internal.features.entries import PRFactsCalculator
+from athenian.api.internal.jira import get_jira_installation
+from athenian.api.internal.miners.filters import JIRAFilter, LabelFilter
+from athenian.api.internal.miners.github.bots import bots
+from athenian.api.internal.miners.github.pull_request import fetch_prs_numbers
+from athenian.api.internal.miners.types import (
+    JIRAEntityToFetch,
+    PRParticipants,
+    PRParticipationKind,
+    PullRequestFacts,
+)
+from athenian.api.internal.prefixer import Prefixer, RepositoryName
+from athenian.api.internal.reposet import resolve_repos_with_request
+from athenian.api.internal.settings import LogicalRepositorySettings, ReleaseSettings, Settings
+from athenian.api.internal.with_ import resolve_withgroups
+from athenian.api.models.web import (
+    PullRequestDigest,
+    SearchPullRequestsRequest,
+    SearchPullRequestsResponse,
+)
+from athenian.api.request import AthenianWebRequest, model_from_body
+from athenian.api.response import model_response
+from athenian.api.tracing import sentry_span
+
+log = logging.getLogger(__name__)
+
+
+async def search_prs(request: AthenianWebRequest, body: dict) -> web.Response:
+    """Search pull requests that satisfy the query."""
+    search_request = model_from_body(SearchPullRequestsRequest, body)
+    connectors = _SearchPRsConnectors(request.mdb, request.pdb, request.rdb, request.cache)
+    account_info = await _build_account_info(search_request.account, request)
+    search_filter = await _build_filter(search_request, account_info, request)
+    repos_settings = await _build_repos_settings(account_info, search_filter.repositories)
+    pr_digests = await _search_pr_digests(search_filter, account_info, repos_settings, connectors)
+    return model_response(SearchPullRequestsResponse(pull_requests=pr_digests))
+
+
+@sentry_span
+async def _build_account_info(account: int, request: AthenianWebRequest) -> _SearchPRsAccountInfo:
+    meta_ids = await get_metadata_account_ids(account, request.sdb, request.cache)
+    prefixer = await Prefixer.load(meta_ids, request.mdb, request.cache)
+    settings = Settings.from_account(
+        account, prefixer, request.sdb, request.mdb, request.cache, request.app["slack"],
+    )
+    account_bots = await bots(account, meta_ids, request.mdb, request.sdb, request.cache)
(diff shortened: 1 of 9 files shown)


server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -0,0 +1,202 @@
...
+async def _build_account_info(account: int, request: AthenianWebRequest) -> _SearchPRsAccountInfo:
+    meta_ids = await get_metadata_account_ids(account, request.sdb, request.cache)
+    prefixer = await Prefixer.load(meta_ids, request.mdb, request.cache)
+    settings = Settings.from_account(
+        account, prefixer, request.sdb, request.mdb, request.cache, request.app["slack"],
+    )
+    account_bots, release_settings, logical_settings = await gather(
+        bots(account, meta_ids, request.mdb, request.sdb, request.cache),
+        settings.list_release_matches(),
+        settings.list_logical_repositories(),
```
vmarkovtsev, line 59: We should ideally specify only repositories from the filter, if they exist (they will always exist in practice).

server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -0,0 +1,202 @@
...
+    if tasks:
+        results = await gather(*tasks.values())
+        for task_name, result in zip(tasks, results):
+            match task_name:
+                case "repositories":
+                    kwargs["repositories"] = {r.unprefixed for r in result[0]}
+                case "participants":
+                    kwargs["participants"] = result[0] if result else {}
+                case "jira_conf":
+                    kwargs["jira"] = JIRAFilter.from_web(search_req.jira, result)
```
vmarkovtsev, line 118: We'd rather do this:
```python
kwargs = {}

async def _resolve_repos():
    if search_req.repositories is None:
        return
    kwargs["repositories"] = {
        r.unprefixed for r in await resolve_repos_with_request(
            search_req.repositories,
            account_info.account,
            request,
            meta_ids=account_info.meta_ids,
            prefixer=account_info.prefixer,
            pointer=".repositories",
        )
    }
...
await gather(_resolve_repos(), _resolve_withgroups(), _resolve_jira())
```

server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -0,0 +1,202 @@
...
+    )
+    pr_digests = [
+        PullRequestDigest(number=number, repository=repository)
+        for node_id, repository_full_name, number in zip(
+            pr_facts[PullRequestFacts.f.node_id].values,
+            pr_facts[PullRequestFacts.f.repository_full_name].values,
+            prs_numbers,
+        )
+        if number != 0
+        and (repository := account_info.prefixer.prefix_logical_repo(repository_full_name))
```
vmarkovtsev, line 191: I am worried that we will do this for each pull request. We'd better optimize like this:
```
repo_order = np.argsort(pr_facts[PullRequestFacts.f.repository_full_name].values)
unique_repos, repo_lengths = np.unique(pr_facts[PullRequestFacts.f.repository_full_name].values[repo_order], return_counts=True)
unique_repos = [prefix_logical_repo(r) for r in unique_repos]
prefixed_repos = np.repeat(unique_repos, repo_lengths)[repo_order]  # the order will match pr_facts
```

vmarkovtsev (changes requested):

server/athenian/api/request.py
```diff
@@ -34,3 +37,11 @@ class AthenianWebRequest(web.Request):
     uid: str
     account: Optional[int]
     is_default_user: bool
+
+
+def model_from_body(model_class: Type[Model], body: dict):
```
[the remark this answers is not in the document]
gaetano-guerriero: typing fixed,
`base_model_` imports `serialization` so that creates a circular import for `Model`

server/athenian/api/controllers/search_controller/search_prs.py
gaetano-guerriero: not always, repo filter is optional in goals and so optional in the initial use case of this API call, anyway optimization done.

server/athenian/api/controllers/search_controller/search_prs.py
gaetano-guerriero: Right, this could become slow with many logical repos.
Given that we know what repos we put in, can we also build the mapping outside the loop ? This is faster, and also avoids complexity in this higher level code:
```
repo_mapping = {r: prefix_logical_repo(r) for r in repos}
PullRequestDigest(number=number, repository=repo_mapping[repository_full_name])
```
This is 5.5 ms against 6.5ms of current loop (with no logical repos).

vmarkovtsev: ok good

server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -0,0 +1,220 @@
...
+    prefix_logical_repo = account_info.prefixer.prefix_logical_repo
+    repo_mapping = {r: prefix_logical_repo(r) for r in repos}
+    pr_digests = [
+        PullRequestDigest(number=number, repository=repo_mapping[repository_full_name])
+        for node_id, repository_full_name, number in zip(
+            pr_facts[PullRequestFacts.f.node_id].values,
+            pr_facts[PullRequestFacts.f.repository_full_name].values,
+            prs_numbers,
+        )
+        if number != 0
```
vmarkovtsev, line 210:
```python
mask = prs_numbers != 0
for node_id, ... in zip(
    pr_facts[PullRequestFacts.f.node_id].values[mask],
    pr_facts[PullRequestFacts.f.repository_full_name].values[mask],
    prs_numbers[mask]
)
```

server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -0,0 +1,220 @@
...
+        prefixer=account_info.prefixer,
+        fresh=False,
+        with_jira=JIRAEntityToFetch.NOTHING,
+    )
+
+    prs_numbers = await fetch_prs_numbers(
+        pr_facts[PullRequestFacts.f.node_id].values, account_info.meta_ids, mdb,
+    )
+
+    prefix_logical_repo = account_info.prefixer.prefix_logical_repo
```
vmarkovtsev, line 201:
```
with sentry_sdk.start_span(op="materialize models", description=str(len(pr_facts))):
    ...
```

vmarkovtsev (changes requested):

vmarkovtsev (approved):

PR #3076, opened 2022-11-28
[DEV-5347] search prs order by metrics

gaetano-guerriero:


server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -220,3 +240,137 @@ async def _search_pr_digests(
...
+        order_by_metrics = _OrderByMetrics.build(pr_facts, search_filter, order_by)
+    else:
+        order_by_metrics = None
+
+    order_keys: list[npt.NDArray] = []
+    keep_mask = np.full((len(pr_facts),), True, bool)
+
+    for expr in order_by:
+        if expr.field in _OrderByMetrics.FIELDS:
+            assert order_by_metrics is not None
```
vmarkovtsev, line 265:
```suggestion
    order_keys: list[npt.NDArray] = []
    keep_mask = np.full((len(pr_facts),), True, bool)
    order_by_metrics = None

    for expr in order_by:
        if expr.field in _OrderByMetrics.FIELDS:
            if order_by_metrics is None:
                order_by_metrics = _OrderByMetrics.build(pr_facts, search_filter, order_by)
```

server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -195,9 +211,13 @@ async def _search_pr_digests(
         with_jira=JIRAEntityToFetch.NOTHING,
     )
 
+    if order_by:
+        pr_facts = _apply_order_by(pr_facts, search_filter, order_by)
+
     prs_numbers = await fetch_prs_numbers(
         pr_facts[PullRequestFacts.f.node_id].values, account_info.meta_ids, mdb,
     )
+
```
vmarkovtsev, line 220:
```suggestion
```

server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -220,3 +240,137 @@ async def _search_pr_digests(
...
+        pr_facts: pd.DataFrame,
+        search_filter: _SearchPRsFilter,
+        order_by: Sequence[SearchPullRequestsOrderByExpression],
+    ) -> Optional[_OrderByMetrics]:
+        metrics = [expr.field for expr in order_by if expr.field in cls.FIELDS]
+        assert metrics
+        min_times = np.array(
+            [search_filter.time_from.replace(tzinfo=None)], dtype="datetime64[ns]",
+        )
+        max_times = np.array([search_filter.time_to.replace(tzinfo=None)], dtype="datetime64[ns]")
```
vmarkovtsev, line 300:
```suggestion
        min_times, max_times = (np.array(
            [t.replace(tzinfo=None)], dtype="datetime64[ns]",
        ) for t in (search_filter.time_from, search_filter.time_to))
```

server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -220,3 +240,137 @@ async def _search_pr_digests(
...
+        )
+        groups = np.full((1, len(pr_facts)), True, bool)
+        calc_ensemble(pr_facts, min_times, max_times, groups)
+        return cls(calc_ensemble)
+
+    def parse_expression(
+        self,
+        expr: SearchPullRequestsOrderByExpression,
+    ) -> tuple[npt.NDArray, npt.NDArray[int]]:
+        """Parse an expression and return a tuple with order keys and discard indexes.
```
vmarkovtsev, line 312: which expression? not clear...

server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -220,3 +240,137 @@ async def _search_pr_digests(
...
+        `order_keys` is an array with the same length as `pr_facts` with the sorting key for
+        this expression for each pr.
+        `discard` is an array with the indexes of element in `pr_facts` to drop from end result
+        due to expression.
+
+        """
+        # first calculator is taken, calculators using split() are not yet supported
+        calc = self._calc_ensemble[expr.field][0]
+        values = calc.peek[0]
+        assert len(values)
```
vmarkovtsev, line 323: Does not always happen, we can easily fetch 0 PRs according to the user's filters.

server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -220,3 +240,137 @@ async def _search_pr_digests(
...
+        if expr.direction == OrderByDirection.DESCENDING.value:
+            # with descending order I want a huge null if they should be first in the result
+            nullmax = expr.nulls_first
+        else:
+            # with ascending order I want a huge null if they should last in the result
+            nullmax = not expr.nulls_first
+
+        null_replacement: int | float | np.timedelta64
+
+        if ar.dtype == np.dtype("timedelta64[s]"):
```
vmarkovtsev, line 361: We should check against `Calculator.nan` like this:
```python
if calc.has_nan:
    nulls = ar != ar
else:
    nulls = ar == calc.nan
```

vmarkovtsev: Regarding the replacements, the whole approach should be different (and way simpler!):

```python
notnulls = ~nulls
notnull_indexes = np.flatnonzero(notnulls)[np.argsort(ar[notnull]s)]
null_indexes = np.flatnonzero(nulls)
# then decide depending on `nullmax` whether to set indexes = np.concatenate([notnull_indexes, null_indexes]) or vice versa
```

The goal is to avoid all the custom code.

vmarkovtsev (changes requested):

server/athenian/api/controllers/search_controller/search_prs.py
gaetano-guerriero: len(pr_facts) == 0 is also checked in `_apply_order_by`, anyway I'm removing this it was more for debugging

server/athenian/api/controllers/search_controller/search_prs.py
gaetano-guerriero: the `SearchPullRequestsOrderByExpression` passed as argument, added "order by expression" to be more clear

server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -220,3 +239,133 @@ async def _search_pr_digests(
...
+        if expr.field in _OrderByMetrics.FIELDS:
+            if order_by_metrics is None:
+                order_by_metrics = _OrderByMetrics.build(pr_facts, search_filter, order_by)
+            exp_order_keys, discard = order_by_metrics.parse_expression(expr)
+            order_keys.append(exp_order_keys)
+            keep_mask[discard] = False
+        else:
+            raise ValueError(f"Invalid order by field {expr.field}")
+
+    sorted_positions = np.lexsort(list(reversed(order_keys)))
```
vmarkovtsev, line 268:
```suggestion
    sorted_positions = np.lexsort(order_keys[::-1])
```
saves one copy

server/athenian/api/controllers/search_controller/search_prs.py
gaetano-guerriero: I've replaced `nulls` building with the suggested code.
Regarding replacements, I understand this code generates indexes to order by a single expression,  can it be adapted for multiple expressions ?
My approach generates a list of sorting keys and pass it to `lexsort` so I need to convert nulls to some value to control their position for each sorting keys.
The ugly part is ofc the switch on calc.dtype to get replacements

vmarkovtsev: You should ditch lexsort in favor of several regular `np.sort`-s. The performance will be similar if not better because lexsort is slow. Even it becomes worse, it's still worth because the current code is too complex and tricky.

vmarkovtsev: To elaborate, `np.sort(kind="stable")` for each column in the opposite direction with nulls appended or prepended.

vmarkovtsev: I mean this:
```python
state = np.arange(len())
for expr in reversed(expressions):
    ar = ...  # column
    ar = ar[state]
    if calc.has_nan:
        nulls = ar != ar
    else:
        nulls = ar == calc.nan
    state_notnull = state[~nulls][np.argsort(ar[~nulls], kind="stable")]
    state_parts = [state[nulls], state_notnull]
    if expr.nulls_last:
        state_parts = state_parts[::-1]
    state = np.concatenate(state_parts)
```

gaetano-guerriero: nice, this is indeed simpler

server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -220,3 +236,97 @@ async def _search_pr_digests(
...
+            for t in (search_filter.time_from, search_filter.time_to)
+        )
+        calc_ensemble = PullRequestMetricCalculatorEnsemble(
+            *metrics, quantiles=(0, 1), quantile_stride=DEFAULT_QUANTILE_STRIDE,
+        )
+        groups = np.full((1, len(pr_facts)), True, bool)
+        calc_ensemble(pr_facts, min_times, max_times, groups)
+        return cls(calc_ensemble)
+
+    def parse_expression(
```
vmarkovtsev, line 297: I'd rather rename this to `parse_and_apply_expression` since we are doing quite some extra work besides parsing.

server/athenian/api/controllers/search_controller/search_prs.py
```diff
@@ -220,3 +236,97 @@ async def _search_pr_digests(
...
+        calc = self._calc_ensemble[expr.field][0]
+        values = calc.peek[0][ordered_indexes]
+
+        if calc.has_nan:
+            nulls = values != values
+        else:
+            nulls = values == calc.nan
+
+        if len(nulls) and expr.exclude_nulls:
+            discard = nulls
```
vmarkovtsev, line 318: Note: in this case, the array dtype will be boolean. Did you mean: `np.flatnonzero(nulls)`?

vmarkovtsev (approved):

PR #3102, opened 2022-12-07
[DEV-5401] Validate account query parameter

gaetano-guerriero:


server/athenian/api/auth.py
```diff
@@ -747,3 +709,76 @@ def ensure_non_default_user(request: AthenianWebRequest):
...
+
+    async def in_query(self, request: EspecificoRequest) -> None:
+        """Validate the account in the request query parameters."""
+        accounts = request.query.get("account")
+        if not accounts:
+            return
+        try:
+            account = int(accounts[0])
+        except ValueError:
+            return
```
vmarkovtsev, line 760:
```suggestion
        try:
            account = int(request.query["account"][0])
        except (ValueError, KeyError):
            return
```

server/athenian/api/auth.py
```diff
@@ -747,3 +709,76 @@ def ensure_non_default_user(request: AthenianWebRequest):
...
+
+        with sentry_sdk.configure_scope() as scope:
+            scope.set_tag("account", account)
```
vmarkovtsev, line 763: Let's extract this to a function, we use this code in both validators.

vmarkovtsev (changes requested):

vmarkovtsev (approved):

PR #3120, opened 2022-12-13
[DEV-5365] pr cycle time alias pr lead time

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3120)


vmarkovtsev (approved):
Is this s backward-incompatible change that requires a webapp update? In that case, let's not merge it until the webapp deploys to production. For the sake of playing safe :pray:

gaetano-guerriero:
This is a backward incompatible change in the sense that semantic for `pr-cycle-*` metrics is changed, but we want this change in the webapp, i.e. we want to use the old `pr-lead-time` where we were using the old `pr-cycle-time`. 
https://athenian.slack.com/archives/CRE79AHLH/p1670848455396879?thread_ts=1670838448.085069&cid=CRE79AHLH
So this can be shipped with no changes on the webapp.

PR #3121, opened 2022-12-13
[DEV-5508] Allow default user in match_identities controller

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3121)


server/athenian/api/controllers/integrations_controller.py
```diff
@@ -27,7 +26,6 @@
 from athenian.api.response import ResponseError, model_response
 
 
-@disable_default_user
```
vmarkovtsev, line 30: I think that our fear was in DoS-ing the backend here.

vmarkovtsev (approved):

PR #3122, opened 2022-12-13
[DEV-5399] spec metric params validation

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3122)

### DIFF
--- a/server/athenian/api/controllers/goal_controller.py
+++ b/server/athenian/api/controllers/goal_controller.py
@@ -13,6 +13,7 @@ from athenian.api.align.goals.dbaccess import (
     delete_goal as db_delete_goal,
     delete_goal_template_from_db,
     dump_goal_repositories,
+    fetch_goal,
     fetch_goal_account,
     fetch_team_goals,
     get_goal_template_from_db,
@@ -57,6 +58,7 @@ from athenian.api.models.web import (
     GoalTemplateUpdateRequest,
     GoalUpdateRequest,
     InvalidRequestError,
+    PullRequestMetricID,
 )
 from athenian.api.request import AthenianWebRequest
 from athenian.api.response import ResponseError, model_response
@@ -305,7 +307,8 @@ async def update_goal(request: AthenianWebRequest, id: int, body: dict) -> web.R
             account = await fetch_goal_account(id, sdb_conn)
             if not await request_user_belongs_to_account(request, account):
                 raise GoalNotFoundError(id)
-            update_info = await _parse_update_request(account, request, update_req)
+            goal = await fetch_goal(account, id, sdb_conn)
+            update_info = await _parse_update_request(goal, request, update_req)
             await replace_team_goals(account, id, update_info.team_goals, sdb_conn)
             kw = {
                 Goal.repositories.name: update_info.repositories,
@@ -389,10 +392,10 @@ class _GoalUpdateInfo:
 
 @sentry_span
 async def _parse_update_request(
-    account_id: int,
+    goal: Goal,
     request: AthenianWebRequest,
     update_req: GoalUpdateRequest,
-):
+) -> _GoalUpdateInfo:
     duplicates = []
     seen = set()
     team_goals = []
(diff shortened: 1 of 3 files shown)


server/athenian/api/application.py
```diff
@@ -931,6 +931,11 @@ def add_api(self, specification: str, ref_resolver_store: Optional[dict] = None,
 
         self.app["route_spec"] = route_spec
 
+        # collect all schemas from the spec, so that they can be used for custom validation
```
vmarkovtsev, line 934: Do we expect `schemas_spec` to sometimes exist, and sometimes not?

server/athenian/api/controllers/goal_controller.py
```diff
@@ -420,6 +425,19 @@ async def _parse_update_request(
...
+    team_goals_pseudo_instance = {
+        "metric": goal[Goal.metric.name],
+        "team_goals": [tg.to_dict() for tg in update_req.team_goals],
+    }
+    try:
+        team_goals_validator.validate(team_goals_pseudo_instance)
+    except jsonschema.ValidationError as e:
+        raise ResponseError(
+            InvalidRequestError(".team_goals", f'Invalid team goals: "{e.message}"'),
+        )
```
vmarkovtsev, line 439: To me, this is an indicator that we are misusing OpenAPI. These checks should happen automatically. Are we underdefining the spec?

Look:
- We first deserialize JSON to dicts
- Then validate them over OpenAPI spec
- Then make the request object, validating them further
- Then we convert a part of the request back to dicts
- Then we validate those dicts over another OpenAPI spec

This is the only such place and it is not worth the trouble. We should hardcode all the checks here without feeling shame.

vmarkovtsev (changes requested):

server/athenian/api/controllers/goal_controller.py
gaetano-guerriero: The problem here is that the spec cannot express the restriction.
The restriction is about validating a piece of the request (metric_params) with something that it's not in the GoalUpdateRequest, but in the target object on the DB.
By creating a new object with can reuse the existing schema for the GoalCreateRequest

alternatives are:
A. write some procedural code to find the metric type and then manually call the correct subschema for metric_params
B. write completely procedural code for the validation of metric_params against metric
C. decide that  the ~80 lines to do B are not worth the effort and sacrifice metric_params validation for update_goal

server/athenian/api/application.py
gaetano-guerriero: fist added specification creates the `schemas_spec` dict in the app, following ones will update the same dict; just the same as `route_spec`

gaetano-guerriero: removed now

server/athenian/api/controllers/goal_controller.py
gaetano-guerriero: implemented with procedural code, PTAL

server/athenian/api/controllers/goal_controller.py
```diff
@@ -429,3 +436,40 @@ async def _parse_update_request(
...
+                    InvalidRequestError(".team_goals", f'Invalid team goals: "{tg}"'),
+                )
+
+
+def _check_team_goals_threshold_type(
+    metric: str,
+    update_req: GoalUpdateRequest,
+    types: type | tuple[type, ...],
+) -> None:
+    for tg in [tg for tg in update_req.team_goals if tg.metric_params]:
```
vmarkovtsev, line 472: too many `tg`-s :smile: 
```suggestion
    for tg in update_req.team_goals:
        if not tg.metric_params:
            continue
```

server/athenian/api/controllers/goal_controller.py
```diff
@@ -429,3 +436,40 @@ async def _parse_update_request(
         jira_priorities=jira_priorities,
         jira_issue_types=jira_issue_types,
     )
+
+
```
vmarkovtsev, line 440: IMO this looks nicer unless we'll have to update this frequently. Probably, we'll move them somewhere up where the metrics live, even.

vmarkovtsev (approved):
I think this is better than messing with the spec. Let's see how it will feel to maintain...

PR #3166, opened 2022-12-22
[DEV-5582] refactor db repositories access

gaetano-guerriero:
Moved parse/dump/resolved goal repositories functions outside of goals module
Also use `prefixer.dereference_repositories(references)` in `dereference_db_repositories` to avoid code repetition

This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3166)

### DIFF
--- a/server/athenian/api/internal/repos.py
+++ b/server/athenian/api/internal/repos.py
@@ -1,0 +1,51 @@
+from __future__ import annotations
+
+import logging
+from typing import Optional, Sequence
+
+from athenian.api import metadata
+from athenian.api.internal.prefixer import Prefixer, RepositoryName, RepositoryReference
+from athenian.api.internal.settings import LogicalRepositorySettings
+
+
+def dereference_db_repositories(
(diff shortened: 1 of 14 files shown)


server/athenian/api/internal/repos.py
```diff
@@ -0,0 +1,51 @@
...
+def parse_db_repositories(
+    val: Optional[list[tuple[int, str | None]]],
+) -> Optional[list[RepositoryReference]]:
+    """Parse the raw value in DB repositories JSON column as list of `RepositoryReference`."""
+    if val is None:
+        return val
+    return [
+        RepositoryReference("github.com", repo_id, logical_name or "")
+        for repo_id, logical_name in val
+    ]
```
vmarkovtsev, line 41: Well, we should plan changing all the places in the DB to include the "github.com" prefix. This will reduce the pain from gitlab. Similarly to the reposets.

vmarkovtsev (approved):

PR #3214, opened 2023-01-03
[DEV-5639] batch calc physical include logical repos

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3214)

### DIFF
--- a/server/athenian/api/internal/settings.py
+++ b/server/athenian/api/internal/settings.py
@@ -638,6 +638,42 @@ class LogicalRepositorySettings:
             return False
         return logical_repo.unprefixed in physical_repo_settings.logical_repositories
 
+    def augment_with_logical_repos(self, repos: Iterable[str]) -> list[str]:
+        """Augment repos requested by user with logical repositories.
+
+        For each physical repo its logical components are added to the list when no
+        logical component is already explicitly present in the list.
+        Exploded physical repos are left in the result.
+
+        Example:
+        Given repository "athenianco/r0" with "athenianco/r0/l0" and
+        "athenianco/r0/l1" logical components, and "athenianco/r1" with no logical
+        components:
+
+        ```python
+        >>> logical_settings.augment_with_logical_repos(["athenianco/r0"])
+        ["athenianco/r0", "athenianco/r0/l0", "athenianco/r0/l1"]
+        >>> logical_settings.augment_with_logical_repos(["athenianco/r0", "athenianco/r0/l0"])
+        ["athenianco/r0", "athenianco/r0/l0"]
+        >>> logical_settings.augment_with_logical_repos(["athenianco/r1", "athenianco/r0/l0"])
+        ["athenianco/r1", "athenianco/r0/l0"]
+        ```
+        """
+        names = [RepositoryName.from_unprefixed(r) for r in repos]
+        referred_physical = {n.unprefixed_physical for n in names if n.is_logical}
+
+        exploded = []
+        for name in names:
+            exploded.append(unprefixed := name.unprefixed)
+            if (not name.is_logical) and name.unprefixed not in referred_physical:
+                try:
+                    logical_repos = self.prs(name.unprefixed).logical_repositories
+                except KeyError:
+                    continue
+
+                exploded.extend(r for r in logical_repos if r != unprefixed)
+        return exploded
+
 
 class Settings:
     """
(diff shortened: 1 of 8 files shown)


server/athenian/api/internal/settings.py
```diff
@@ -638,6 +638,27 @@ def repo_exists(self, logical_repo: RepositoryName) -> bool:
...
+        names = [RepositoryName.from_unprefixed(r) for r in repos]
+        referred_physical = {n.unprefixed_physical for n in names if n.is_logical}
+
+        exploded = []
+        for name in names:
+            exploded.append(unprefixed := name.unprefixed)
+            if (not name.is_logical) and name.unprefixed not in referred_physical:
+                try:
+                    logical_repos = self.prs(name.unprefixed).logical_repositories
+                except KeyError:
```
vmarkovtsev, line 656: We shouldn't handle a `KeyError` here. If we catch it, we did a poor job validating the repositories beforehand. We should loudly crash here and make the calling code satisfy the contract.

vmarkovtsev (changes requested):
Just one nitpick

server/athenian/api/internal/settings.py
gaetano-guerriero: At least in unit tests repositories with no logical components are not found in `self._prs` so `KeyError` happens also for valid physical repositories, which should be left untouched. I can replace this with a `and name.unprefixed in self._prs` though

server/athenian/api/internal/settings.py
```diff
@@ -638,6 +638,27 @@ def repo_exists(self, logical_repo: RepositoryName) -> bool:
             return False
         return logical_repo.unprefixed in physical_repo_settings.logical_repositories
 
+    def explode_user_repos(self, repos: Iterable[str]) -> list[str]:
```
vmarkovtsev, line 641:
```suggestion
    def augment_with_logical_repos(self, repos: Iterable[str]) -> list[str]:
```

server/athenian/api/internal/settings.py
```diff
@@ -638,6 +638,27 @@ def repo_exists(self, logical_repo: RepositoryName) -> bool:
...
+        """Explode the list of repositories requested by user.
+
+        Each physical repo is exploded with its logical components if none of them is in the list.
+        Exploded physical repos are left in the result.
+        """
+        names = [RepositoryName.from_unprefixed(r) for r in repos]
+        referred_physical = {n.unprefixed_physical for n in names if n.is_logical}
+
+        exploded = []
+        for name in names:
```
vmarkovtsev, line 651:
```suggestion
        for name in referred_physical:
```
because we don't have to do anything with physical repositories without logical occurrences. This explains why you catch `KeyError` below.

server/athenian/api/internal/settings.py
```diff
@@ -638,6 +638,27 @@ def repo_exists(self, logical_repo: RepositoryName) -> bool:
             return False
         return logical_repo.unprefixed in physical_repo_settings.logical_repositories
 
+    def explode_user_repos(self, repos: Iterable[str]) -> list[str]:
+        """Explode the list of repositories requested by user.
+
+        Each physical repo is exploded with its logical components if none of them is in the list.
```
vmarkovtsev, line 644:
```suggestion
        We insert all logical components of a repo if at least one logical component appears in `repos`.
```
(that's my understanding of the logic)

vmarkovtsev (changes requested):

server/athenian/api/internal/settings.py
gaetano-guerriero: That's the opposite, if no logical components is in the list the user wants to refer everything so we add all logical components. 
https://athenian.slack.com/archives/CQU6FR09Z/p1672665741875189?thread_ts=1672664721.132099&cid=CQU6FR09Z
Given repos `athenianco/r0` and `athenianco/r0/l0` `athenianco/r0/l1`logical components:
`[athenianco/r0] => [atheniaco/r0, athenianco/r0/l0, athenianco/r0/l1]`
`[atheniaco/r0, athenianco/r0/l0] => [atheniaco/r0, athenianco/r0/l0]`

gaetano-guerriero: I've rephrased the docstring

PR #3226, opened 2023-01-05
[DEV-5684] Handle the edge case with review request without review in an open PR

vmarkovtsev:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3226)

### DIFF
--- a/server/tests/controllers/miners/github/test_pull_request.py
+++ b/server/tests/controllers/miners/github/test_pull_request.py
@@ -21,6 +21,7 @@ from athenian.api.internal.miners.github.precomputed_prs import (
     store_open_pull_request_facts,
 )
 from athenian.api.internal.miners.github.pull_request import PullRequestFactsMiner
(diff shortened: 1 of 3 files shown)


server/tests/controllers/miners/github/test_pull_request.py
```diff
@@ -2017,3 +2018,95 @@ async def test_mine_pull_requests_event_releases(
     facts2 = await calc(*args)
     facts2.sort_values(PullRequestFacts.f.created, inplace=True, ignore_index=True)
     assert_frame_equal(facts1.iloc[1:], facts2.iloc[1:])
+
+
+def test_pull_request_review_request_without_review_open():
+    miner = PullRequestFactsMiner(set())
+    pr = MinedPullRequest(
```
vmarkovtsev, line 2025: @gaetano-guerriero do we already have a factory for this? It would be really handy...

PR #3249, opened 2023-01-13
[DEV-5772] Fix grouping with one participants groups in batch_calc

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3249)

### DIFF
--- a/server/athenian/api/internal/features/entries.py
+++ b/server/athenian/api/internal/features/entries.py
@@ -376,7 +376,7 @@ class MetricEntriesCalculator:
         )
         lines_grouper = partial(group_prs_by_lines, lines)
         repo_grouper = partial(group_by_repo, PullRequest.repository_full_name.name, repositories)
-        with_grouper = partial(group_prs_by_participants, participants)
+        with_grouper = partial(group_prs_by_participants, participants, True)
         dedupe_mask = calculate_logical_prs_duplication_mask(
             df_facts, release_settings, logical_settings,
         )
@@ -482,7 +482,9 @@ class MetricEntriesCalculator:
                     ],
                     df_facts,
                 ),
-                group_prs_by_participants([t.participants for t in request.teams], df_facts),
+                group_prs_by_participants(
+                    [t.participants for t in request.teams], False, df_facts,
+                ),
                 group_pr_facts_by_jira(jira_grouping, df_facts),
             ]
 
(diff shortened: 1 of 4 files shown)


server/athenian/api/internal/features/entries.py
```diff
@@ -376,7 +377,12 @@ async def calc_pull_request_metrics_line_github(
         )
         lines_grouper = partial(group_prs_by_lines, lines)
         repo_grouper = partial(group_by_repo, PullRequest.repository_full_name.name, repositories)
-        with_grouper = partial(group_prs_by_participants, participants)
+        # as optimization, if len(participants) == 1, we've already filtered in SQL
+        # so don't have to re-check
+        if len(participants) == 1:
```
gaetano-guerriero, line 382: imho it's simpler to have the optimization outside, else group_prs_by_participants contract is more complex

vmarkovtsev: Let's move it inside. I was there, and it's hell once the number of these optimizations grow. It is trivial to move `if` inside `group_prs_by_participants` and, honestly, it's their business to work correctly. Besides, we have similar `if`-s in other groupers.

vmarkovtsev (changes requested):

server/athenian/api/internal/features/entries.py
gaetano-guerriero: If we keep it  inside I need to add a boolean flag `skip_single_group_optimization` to `group_prs_by_participants`, is this better ?

gaetano-guerriero: This is because in batch_calc_prs... to SQL we passed multiple participant groups, but then for a single request we group by a single participant group

server/athenian/api/internal/features/github/pull_request_metrics.py
```diff
@@ -87,19 +87,24 @@ def group_prs_by_lines(lines: Sequence[int], items: pd.DataFrame) -> List[np.nda
...
+    if `participants` contains a single group.
     """
     # FIXME: participants here can also be List[Dict[PRParticipationKind, Set[str]]]
 
-    # if len(participants) == 1, we've already filtered in SQL so don't have to re-check
     # if there are no participant groups also we select everything and don't re-check
-    if len(participants) < 2:
+    # if len(participants) == 1 and items only contains specified participants we also select
+    # everything
+    if not participants or (items_only_contain_specified_participants and len(participants) == 1):
```
vmarkovtsev, line 107:
```suggestion
    if len(participants) == 0 or (items_only_contain_specified_participants and len(participants) == 1):
```
if `participants` is a numpy array, it will crash

vmarkovtsev (changes requested):

vmarkovtsev (approved):

PR #3252, opened 2023-01-16
[DEV-5731] lazy prefixer proxy

gaetano-guerriero:
To be used as:
```
async def controller():
   prefixer_proxy  =LazyPrefixerProxy()
   awaith gather(a(prefixer_proxy), b(prefixer_proxy))

async def a(prefixer_proxy):
   if  cond_a:
      prefixer = prefixer_proxy()

async def b(prefixer_proxy):
   if  cond_b:
      prefixer = prefixer_proxy()
```

This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3252)

### DIFF
--- a/server/athenian/api/controllers/dashboard_controller.py
+++ b/server/athenian/api/controllers/dashboard_controller.py
@@ -14,18 +14,17 @@ from athenian.api.internal.dashboard import (
     reorder_dashboard_charts,
 )
 from athenian.api.internal.jira import parse_request_issue_types, parse_request_priorities
-from athenian.api.internal.prefixer import Prefixer
-from athenian.api.internal.repos import dump_db_repositories
+from athenian.api.internal.prefixer import LazyPrefixerProxy, Prefixer
+from athenian.api.internal.repos import parse_request_repositories
 from athenian.api.internal.team import get_team_from_db
 from athenian.api.models.state.models import DashboardChart, Team, TeamDashboard
 from athenian.api.models.web import (
     CreatedIdentifier,
     DashboardChartCreateRequest,
     DashboardUpdateRequest,
-    InvalidRequestError,
 )
 from athenian.api.request import AthenianWebRequest, model_from_body
-from athenian.api.response import ResponseError, model_response
+from athenian.api.response import model_response
 
 
 async def get_dashboard(
@@ -66,7 +65,8 @@ async def create_dashboard_chart(
     dashboard = await _get_request_dashboard(request, team_id, dashboard_id)
     create_request = model_from_body(DashboardChartCreateRequest, body)
 
-    extra_values = await _parse_request_chart_filters(create_request, dashboard.account, request)
+    prefixer_proxy = LazyPrefixerProxy(request, dashboard.account)
+    extra_values = await _parse_request_chart_filters(create_request, prefixer_proxy)
 
     async with request.sdb.connection() as sdb_conn:
         async with sdb_conn.transaction():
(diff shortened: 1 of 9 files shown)


vmarkovtsev:
@gaetano-guerriero We had exactly this code in the past: to lazily load the Prefixer asynchronously. As I remember, even under the same name! :smile: After several months of working with that proxy, I concluded that it was a bad idea and removed it. The reasons were:
- Higher complexity of the source code.
- More difficult to trace the performance. It could take a slot of more critical stuff, sometimes.
- All the places in the code needed the materialized Prefixer anyway.

So I suggest not repeating my mistake here. We will always have an async setup in each controller to load all the stuff: release settings, logical settings, etc. and the prefixer fits there very well.

server/athenian/api/controllers/dashboard_controller.py
```diff
@@ -66,7 +65,8 @@ async def create_dashboard_chart(
     dashboard = await _get_request_dashboard(request, team_id, dashboard_id)
```
vmarkovtsev, line 65: for example, we should load the prefixer at this line, in the very first async call

server/athenian/api/controllers/goal_controller.py
```diff
@@ -137,7 +137,7 @@ async def create_goal_template(request: AthenianWebRequest, body: dict) -> web.R
     await get_user_account_status_from_request(request, create_request.account)
 
     repositories = await parse_request_repositories(
-        create_request.repositories, request, create_request.account,
+        create_request.repositories, LazyPrefixerProxy(request, create_request.account),
```
vmarkovtsev, line 140: Also, if we remove the proxy, we will follow the same conventions that exist in the rest of the codebase, so it should help future programmers

vmarkovtsev (changes requested):

server/requirements.txt
```diff
@@ -1,4 +1,5 @@
 aiohttp[speedups]==3.8.3
+async-lru==1.0.3
```
vmarkovtsev, line 2: We will also avoid using this external package

server/athenian/api/internal/prefixer.py
```diff
@@ -298,3 +299,20 @@ def __repr__(self) -> str:
...
+
+class LazyPrefixerProxy:
+    """When called return an instance of Prefixer, building it only when needed."""
+
+    def __init__(self, request, account):
+        """Init the `LazyPrefixerProxy`."""
+
+        @alru_cache
+        async def get_prefixer():
+            return await Prefixer.from_request(request, account)
```
vmarkovtsev, line 312: Finally, this is likely not working as you'd expect. When you return an async function, it doesn't actually start to execute! You must wrap it into `asyncio.Task` to let it execute along with the rest of the code.

Also, we already have the memcached. It takes very few milliseconds to load the cached prefixed. So we won't win much here.

vmarkovtsev (changes requested):

server/athenian/api/controllers/goal_controller.py
gaetano-guerriero: Looking at the example, without proxy you can:

1 - always build prefixer, you do useless work; imagine cond_a or cond_b is True 10% of the time:
```
async def controller():
   prefixer = Prefixer()
   awaith gather(a(prefixer), b(prefixer))
async def a(prefixer):
   if  cond_a:
        prefixer.do_something()
async def b(prefixer):
   if  cond_b:
          prefixer.do_something()
```

2 - check conditions also in main function, propagate an optional prefixer, code gets complex, 
you repeat condition check, controller() has to mess with logic it shouldn't know, a() and b() has to mess with 
the assumption that the resource they need is there when they need it

```
async def controller():
   if cond_a or cond_b:
        maybe_prefixer = Prefixer()
   else:
        maybe_prefixer = None
   awaith gather(a(maybe_prefixer), b(maybe_prefixer))
async def a(prefixer):
   if  cond_a:
         assert prefixer is not None
        prefixer.do_something()
async def b(prefixer):
   if  cond_b:
          assert prefixer is not None
          prefixer.do_something()
```

3 - build prefixer twice, this is still better than 1 -  if cond_a or cond_b is False 10% of the time
```
async def controller():
   awaith gather(a(), b())
async def a():
   if  cond_a:
        prefixer = Prefixer()
        prefixer.do_something()
async def b():
   if  cond_b:
          prefixer = Prefixer()
          prefixer.do_something()
```

server/athenian/api/internal/prefixer.py
gaetano-guerriero: :thinking: yes I don't want to execute it, it's lazy after all

server/athenian/api/controllers/goal_controller.py
gaetano-guerriero: So I understand you suggest going for 1 - and trust memcached

PR #3265, opened 2023-01-17
[DEV-5726] privileged default user

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3265)

### DIFF
--- a/server/athenian/api/auth.py
+++ b/server/athenian/api/auth.py
@@ -687,23 +687,23 @@ class AthenianAioHttpSecurityHandlerFactory(especifico.security.AioHttpSecurityH
 
 
 def disable_default_user(func):
-    """Decorate an endpoint handler to raise 403 if the user is the default one."""
+    """Decorate an endpoint handler to raise 403 if the user is the default one and is not god."""
 
     async def wrapped_disable_default_user(
         request: AthenianWebRequest,
         *args,
         **kwargs,
     ) -> aiohttp.web.Response:
-        ensure_non_default_user(request)
+        forbid_unprivileged_default_user(request)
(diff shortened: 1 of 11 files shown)


server/athenian/api/auth.py
```diff
@@ -687,23 +687,23 @@ async def wrapper(request: EspecificoRequest):
...
 
     wraps(wrapped_disable_default_user, func)
     return wrapped_disable_default_user
 
 
-def ensure_non_default_user(request: AthenianWebRequest):
-    """Raise exception if the user on the request is the default user."""
-    if request.is_default_user:
+def forbid_unprivileged_default_user(request: AthenianWebRequest):
+    """Raise exception if the request user is the default user and has not god privileges."""
```
vmarkovtsev, line 705:
```suggestion
    """Raise exception if the request user is the default user and has no god privileges."""
```

server/athenian/api/auth.py
```diff
@@ -687,23 +687,23 @@ async def wrapper(request: EspecificoRequest):
...
+    if request.is_default_user and getattr(request, "god_id", None) is None:
```
vmarkovtsev, line 706:
```suggestion
    if request.is_default_user and not hasattr(request, "god_id"):
```
(shorter, the same as in internal/account.py `get_user_account_status_from_request`)

vmarkovtsev (approved):
Question: did we consider setting `--force-user` in e2e tests? This way, we should bypass Auth0 and all the requests will happen on behalf of that user.

gaetano-guerriero:
The would be great, we could use another user, I will check if it plays well with demo mode.

PR #3285, opened 2023-01-25
[DEV-5733] calc metrics prs jiragroups

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3285)


gaetano-guerriero (commented):
> *[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3285)* status: 1 of 6 files reviewed, 3 unresolved discussions (waiting on @vmarkovtsev)

___
*[`server/athenian/api/controllers/histograms_controller.py` line 106 at r1](https://reviewable.io/reviews/athenianco/athenian-api/3285#-NMi2HwA7j5wzzZQhsti:-NMiJOQn3gzTayLdGRDS:b-e4hv27) ([raw file](https://github.com/athenianco/athenian-api/blob/39135e92dc9a13d5e9a2f3894e74c41961ee20e7/server/athenian/api/controllers/histograms_controller.py#L106)):*

What if there are no jiragroups, shall we write `filter_prs.jiragroups[0] if filter_prs.jiragroups else [filter_prs.jira]`? Or are we changing the API spec backward-incompatible?
</blockquote></details>

here `filter_prs` is an instance of `FilterPRs`, there `jiragroups` will always contain something. The parsing `ForSetPullRequest => FilterPRs` is in `compile_filter_prs`
___
*[`server/athenian/api/controllers/metrics_controller.py` line 541 at r1](https://reviewable.io/reviews/athenianco/athenian-api/3285#-NMi33Sw2eUR5Jnj1MA4:-NMiOEvm5mv2gY79maE9:b-smuqmt) ([raw file](https://github.com/athenianco/athenian-api/blob/39135e92dc9a13d5e9a2f3894e74c41961ee20e7/server/athenian/api/controllers/metrics_controller.py#L541)):*

How about
</blockquote></details>

Done.
Right,
I was thinking that since they produce the same result in output we could only use one filter.
But `calculate_for_set_metrics` needs a list of the same length to map back the request to the "for" field in the response.
"for" field in the response will be wrong anyway, it will say "for": {"jira": something but the Jira part was ignored.
___
*[`server/athenian/api/models/web/for_set_pull_requests.py` line 35 at r1](https://reviewable.io/reviews/athenianco/athenian-api/3285#-NMi3jy44UHpGAUheEQ9:-NMiKCNeDLrfPnZ6dtbk:b-896fix) ([raw file](https://github.com/athenianco/athenian-api/blob/39135e92dc9a13d5e9a2f3894e74c41961ee20e7/server/athenian/api/models/web/for_set_pull_requests.py#L35)):*
> ```Python
>                 return self.copy()
>         if self.jiragroups is None or index > len(self.jiragroups):
>             raise IndexError(f"jiragroup {index} doesn't exist")
> ```

Done.

vmarkovtsev (approved):
> *[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3285)* status: 1 of 6 files reviewed, all discussions resolved

PR #3322, opened 2023-02-01
[DEV-5911] Optimize group by repos in batch metrics calculation

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3322)

### DIFF
--- a/server/athenian/api/internal/features/entries.py
+++ b/server/athenian/api/internal/features/entries.py
@@ -486,24 +486,17 @@ class MetricEntriesCalculator:
             df_facts, release_settings, logical_settings,
         )
 
+        repo_grouping = _BatchRepoGrouping(all_repositories, df_facts, logical_settings)
         results = []
         for request in requests:
             jira_grouping = convert_jira_filters_to_grouping(t.jira_filter for t in request.teams)
             group_by = [
-                group_by_repo(
-                    PullRequest.repository_full_name.name,
-                    [
-                        logical_settings.augment_with_logical_repos(t.repositories)
-                        for t in request.teams
-                    ],
-                    df_facts,
-                ),
+                repo_grouping.get_groups([t.repositories for t in request.teams]),
                 group_prs_by_participants(
                     [t.participants for t in request.teams], False, df_facts,
                 ),
                 group_pr_facts_by_jira(jira_grouping, df_facts),
             ]
-
             groups = _intersect_items_groups(len(request.teams), len(df_facts), *group_by)
             groups = deduplicate_groups(
                 groups,
@@ -923,22 +916,13 @@ class MetricEntriesCalculator:
             with_jira=_get_jira_entities_to_fetch(jira_filters, ()),
         )
 
+        repo_grouping = _BatchRepoGrouping(all_repositories, df_facts, logical_settings)
         results = []
         for request in requests:
             jira_grouping = convert_jira_filters_to_grouping(t.jira_filter for t in request.teams)
             group_by = [
-                group_by_repo(
-                    Release.repository_full_name.name,
-                    [
-                        logical_settings.augment_with_logical_repos(t.repositories)
-                        for t in request.teams
-                    ],
-                    df_facts,
-                ),
-                group_releases_by_participants(
-                    [t.participants for t in request.teams],
-                    df_facts,
-                ),
+                repo_grouping.get_groups([t.repositories for t in request.teams]),
+                group_releases_by_participants([t.participants for t in request.teams], df_facts),
                 group_release_facts_by_jira(jira_grouping, df_facts),
             ]
             groups = _intersect_items_groups(len(request.teams), len(df_facts), *group_by)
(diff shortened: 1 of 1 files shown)


vmarkovtsev (changes requested):
> *[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3322)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @gaetano-guerriero)

___
*[`server/athenian/api/internal/features/entries.py` line 495 at r1](https://reviewable.io/reviews/athenianco/athenian-api/3322#-NNBCgkXELsazaS2fiqn:-NNBCgkXELsazaS2fiqo:bhpj053) ([raw file](https://github.com/athenianco/athenian-api/blob/a8a2ee595b09d2c929833dd544caa044929d2719/server/athenian/api/internal/features/entries.py#L495)):*
> ```Python
>             repo_groups = repo_grouping.get_groups([t.repositories for t in request.teams])
>             group_by = [
>                 repo_groups,
> ```

We don't have to cache them AFAIS, so let's save a line



_Code quote:_
```Python
repo_grouping.get_groups([t.repositories for t in request.teams])
```
___
*[`server/athenian/api/internal/features/entries.py` line 1508 at r1](https://reviewable.io/reviews/athenianco/athenian-api/3322#-NNBDDWrE3rxP2Xf4gKH:-NNBDDWrE3rxP2Xf4gKI:b-64kj7v) ([raw file](https://github.com/athenianco/athenian-api/blob/a8a2ee595b09d2c929833dd544caa044929d2719/server/athenian/api/internal/features/entries.py#L1508)):*
> ```Python
>         # it's faster to just compare the lengths
>         if all(len(r) == len(self._all_repositories) for r in request_repos):
>             return [np.arange(len(self._df_facts))] * len(request_repos)
> ```

Regarding this. The way the internals work is we do `len(request_repos)` more work as needed. This in turn blows up the performance on e.g. acc 186. We should have as few groups as possible. It's very worth deduplicating the groups and dereferencing in the final result.

gaetano-guerriero (commented):
> *[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3322)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @vmarkovtsev)

___
*[`server/athenian/api/internal/features/entries.py` line 495 at r1](https://reviewable.io/reviews/athenianco/athenian-api/3322#-NNBCgkXELsazaS2fiqn:-NNBEQpS4_ziehMteLfO:b-896fix) ([raw file](https://github.com/athenianco/athenian-api/blob/a8a2ee595b09d2c929833dd544caa044929d2719/server/athenian/api/internal/features/entries.py#L495)):*

We don't have to cache them AFAIS, so let's save a line
</blockquote></details>

Done.
___
*[`server/athenian/api/internal/features/entries.py` line 1508 at r1](https://reviewable.io/reviews/athenianco/athenian-api/3322#-NNBDDWrE3rxP2Xf4gKH:-NNBG7a64W_WvUtppgfz:b-5h8wc) ([raw file](https://github.com/athenianco/athenian-api/blob/a8a2ee595b09d2c929833dd544caa044929d2719/server/athenian/api/internal/features/entries.py#L1508)):*

Regarding this. The way the internals work is we do `len(request_repos)` more work as needed. This in turn blows up the performance on e.g. acc 186. We should have as few groups as possible. It's very worth deduplicating the groups and dereferencing in the final result.
</blockquote></details>

With this change we compute the same groups but faster since `group_by_repo` is not called when not needed. Basically it won't be needed for each goals not having a repositories filter.

About reducing the number of groups, is this something you propose to do later after this change, or something you propose to do instead of this change ?




Also, this change on the frontend https://athenianco.atlassian.net/browse/DEV-5916?atlOrigin=eyJpIjoiYWQ5MDhmYzc2NTdlNGZmNzg0NzRlMTAwM2I0NzE1NDAiLCJwIjoiamlyYS1zbGFjay1pbnQifQ will greatly reduce requested teams and so group.

vmarkovtsev (approved):
> *[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3322)* status: 0 of 1 files reviewed, all discussions resolved

___
*[`server/athenian/api/internal/features/entries.py` line 1508 at r1](https://reviewable.io/reviews/athenianco/athenian-api/3322#-NNBDDWrE3rxP2Xf4gKH:-NNGVa0W1VJ8uTi-uFe6:b-oq7xzp) ([raw file](https://github.com/athenianco/athenian-api/blob/a8a2ee595b09d2c929833dd544caa044929d2719/server/athenian/api/internal/features/entries.py#L1508)):*
<details><summary><i>Previously, gaetano-guerriero wrote…</i></summary><blockquote>

With this change we compute the same groups but faster since `group_by_repo` is not called when not needed. Basically it won't be needed for each goals not having a repositories filter.

About reducing the number of groups, is this something you propose to do later after this change, or something you propose to do instead of this change ?




Also, this change on the frontend https://athenianco.atlassian.net/browse/DEV-5916?atlOrigin=eyJpIjoiYWQ5MDhmYzc2NTdlNGZmNzg0NzRlMTAwM2I0NzE1NDAiLCJwIjoiamlyYS1zbGFjay1pbnQifQ will greatly reduce requested teams and so group.
</blockquote></details>

Let's do this in a separate PR, you are right. It's already a big improvement.

PR #3376, opened 2023-02-17
[DEV-5736] metrics prs jira labels grouping

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3376)


vmarkovtsev (changes requested):
> Reviewed 8 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3376)* status: 8 of 17 files reviewed, 2 unresolved discussions (waiting on @gaetano-guerriero)

___
*[`server/athenian/api/internal/features/entries.py` line 1614 at r1](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOTl9pQ2VZnlmH1gX6R:-NOTl9pQ2VZnlmH1gX6S:b-j99e72) ([raw file](https://github.com/athenianco/athenian-api/blob/08451d1ad74ee0e062a78f534635202b32b50404/server/athenian/api/internal/features/entries.py#L1614)):*
> ```Python
>                 else:
>                     issue_types = None
>                 # jira_filter.labels.exclude are ignored for grouping
> ```

Note: we should write about it in the OpenAPI spec.
___
*[`server/athenian/api/internal/features/metric_calculator.py` line 1126 at r1](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOTlO5772i9WaxHuOCK:-NOTlO5772i9WaxHuOCL:b-qxhgjx) ([raw file](https://github.com/athenianco/athenian-api/blob/08451d1ad74ee0e062a78f534635202b32b50404/server/athenian/api/internal/features/metric_calculator.py#L1126)):*
> ```Python
>     def __bool__(self) -> bool:
>         """Get whether the group definition is empty.  An empty group does not do any grouping."""
>         return any(getattr(self, f.name) is not None for f in dataclasses.fields(self))
> ```

Nope, this is slow as hell. dataclasses.fields, dataclasses.as_dict, etc. should be banned, they are implemented so slow that it's not even funny.
___
*[`server/athenian/api/internal/miners/jira/issue.py` line 954 at r1](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOTm8ZF3_cwBJbbcOLZ:-NOTm8ZF3_cwBJbbcOL_:b-k8urdt) ([raw file](https://github.com/athenianco/athenian-api/blob/08451d1ad74ee0e062a78f534635202b32b50404/server/athenian/api/internal/miners/jira/issue.py#L954)):*
> ```Python
>         """Fetch the mapping from PR node IDs to JIRA issue IDs."""
>         nprji = NodePullRequestJiraIssues
>         node_id_cond, _ = column_values_condition(nprji.node_id, prs)
> ```

Yeah, we've got 100500 places where we write this. Good point. We would add "progressive_in()" as another method to the columns, similarly to how we did it with `in_any_values()`. That way, we don't have to import a magic function.

gaetano-guerriero (commented):
> *[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3376)* status: 6 of 21 files reviewed, 2 unresolved discussions (waiting on @vmarkovtsev)

___
*[`server/athenian/api/internal/features/metric_calculator.py` line 1126 at r1](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOTlO5772i9WaxHuOCK:-NOja42c8oouLBcBZQpO:b-896fix) ([raw file](https://github.com/athenianco/athenian-api/blob/08451d1ad74ee0e062a78f534635202b32b50404/server/athenian/api/internal/features/metric_calculator.py#L1126)):*

Nope, this is slow as hell. dataclasses.fields, dataclasses.as_dict, etc. should be banned, they are implemented so slow that it's not even funny.
</blockquote></details>

Done.
___
*[`server/athenian/api/internal/miners/jira/issue.py` line 954 at r1](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOTm8ZF3_cwBJbbcOLZ:-NOjmbRu3DPbsDscia07:b-896fix) ([raw file](https://github.com/athenianco/athenian-api/blob/08451d1ad74ee0e062a78f534635202b32b50404/server/athenian/api/internal/miners/jira/issue.py#L954)):*

Yeah, we've got 100500 places where we write this. Good point. We would add "progressive_in()" as another method to the columns, similarly to how we did it with `in_any_values()`. That way, we don't have to import a magic function.
</blockquote></details>

Done.

vmarkovtsev (changes requested):
> Reviewed 1 of 17 files at r1, 1 of 10 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3376)* status: 8 of 21 files reviewed, 9 unresolved discussions (waiting on @gaetano-guerriero)

___
*[`server/athenian/api/db.py` line 506 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnBDDy8PndR64KG0WF:-NOnBDDy8PndR64KG0WG:banhds9) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/db.py#L506)):*
> ```Python
>     cond = column.progressive_in(values)
>     if len(values) > 100:
>         hints: tuple[str] = (f"Rows({column.table.name} *VALUES* #{len(values)})",)
> ```

We cannot do this by default. There are several known exclusions when automatic hints do not work at best and make the planner crazy at worst:
- unions
- pre-filtering, when VALUES is a minor filter
- negations
- the table is a view

Therefore, IMO, hints should always be written manually for each case, considering all the random factors. I am afraid that leaving this function will lead to erroneous hints that will complicate further optimizations - i.e., everyone will think that the query is already optimized. I see we've got one place where we are using this function, so we'd better remove it.
___
*[`server/athenian/api/internal/miners/github/pull_request.py` line 3071 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnFobb4N395ESAzTgG:-NOnFobb4N395ESAzTgH:b-v3vyh8) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/internal/miners/github/pull_request.py#L3071)):*
> ```Python
>                 else:
>                     if col is Issue.labels:
>                         value = np.concatenate(value, dtype=dtype)
> ```

To avoid silly bugs in the future



_Code quote (i):_
```Python
value = np.concatenate(value, dtype=dtype)
```



_Code snippet (ii):_
```Python
value = np.concatenate(value, dtype=dtype, casting="unsafe")
```
___
*[`server/athenian/api/internal/miners/jira/issue.py` line 968 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnGL5h2YgPOsufBXNl:-NOnGL5h2YgPOsufBXNm:brquf12) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/internal/miners/jira/issue.py#L968)):*
> ```Python
>                 ),
>             )
>             .where(nprji.node_id.progressive_in(prs), nprji.node_acc.in_(meta_ids)),
> ```

It's obvious to me that we should hint this. Let's see, there is only one condition and we fetch mapped PRs. We can rely on at least 50% PRs mapped. So `Rows({nprji.__tablename__} *VALUES* {len(prs) // 2})`. However, we've got a JOIN, Postgres likes to move (and will move!) `*VALUES*` outside, leading to bad performance. So we must define `Leading((({nprji.__tablename__} *VALUES*) {Issue.__tablename__}))`

You should see now why automatic hinting is quite useless and even harmful.
___
*[`server/athenian/api/internal/miners/jira/issue.py` line 1187 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnHZfw8L6Iv97JneU7:-NOnHZfw8L6Iv97JneU8:bp1b2r9) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/internal/miners/jira/issue.py#L1187)):*
> ```Python
>         issues[[Issue.acc_id.name, Issue.components.name]]
>         .groupby(Issue.acc_id.name, sort=False)
>         .aggregate(lambda s: set(flatten(s)))
> ```

Production code: no groupby, no aggregate, no flatten with Python objects, for goodness sake. Pandas sucks.
Do we need to group components by acc_id? `order = np.argsort(issues[Issue.acc_id.name].values)` and `np.unique(issues[Issue.acc_id.name].values[order], return_counts=True)`, etc.



_Code quote:_
```Python
        .groupby(Issue.acc_id.name, sort=False)
        .aggregate(lambda s: set(flatten(s)))
```
___
*[`server/athenian/api/internal/miners/jira/issue.py` line 1194 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnIRE33Crstsr1D2Ra:-NOnIRE33Crstsr1D2Rb:b-vv7wbz) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/internal/miners/jira/issue.py#L1194)):*
> ```Python
>     )
>     rows = await mdb.fetch_all(
>         sql.select(Component.acc_id, Component.id, Component.name).where(sql.or_(*conditions)),
> ```

This is a terrible performance, we should never-ever execute OR in SQL except for very few exclusions. We should rewrite to a `UNION` of individual queries by each component condition.

I see we do `r[2].lower()`. Let's move it straight to SQL: `select(Component.name.lower())`.
___
*[`server/athenian/api/internal/miners/jira/issue.py` line 1201 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnJWDBCOMbKSZcTgQb:-NOnJWDBCOMbKSZcTgQc:bxqbx0f) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/internal/miners/jira/issue.py#L1201)):*
> ```Python
>     labels_col = issues[Issue.labels.name].values
>     for i, (acc_id, row_labels, row_components) in enumerate(
>         zip(issues[Issue.acc_id.name].values, labels_col, issues[Issue.components.name].values),
> ```

This is potentially very slow. Someone like account 186 is going to have 100k issues and we will bottleneck here.
Try to find a vectorized numpy way of doing the same thing.
___
*[`server/athenian/api/models/__init__.py` line 151 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnJmGSDQWhvqIGLUVM:-NOnJmGSDQWhvqIGLUVN:b710wr3) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/models/__init__.py#L151)):*
> ```Python
> 
> def _progressive_in(self: ColumnOperators, other: Sized):
>     if len(other) > 100:
> ```

Suggestion: move 100 to the default value of `threshold` kwarg. For future fine-tuning,

vmarkovtsev (changes requested):
> *[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3376)* status: 8 of 21 files reviewed, 9 unresolved discussions (waiting on @gaetano-guerriero)

___
*[`server/athenian/api/internal/miners/jira/issue.py` line 1187 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnHZfw8L6Iv97JneU7:-NOnKv9y11F9HYnHK3Cs:b-jd4fut) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/internal/miners/jira/issue.py#L1187)):*

Production code: no groupby, no aggregate, no flatten with Python objects, for goodness sake. Pandas sucks.
Do we need to group components by acc_id? `order = np.argsort(issues[Issue.acc_id.name].values)` and `np.unique(issues[Issue.acc_id.name].values[order], return_counts=True)`, etc.
</blockquote></details>

One second. We always, always have the same account ID for everything. We don't need to group by account ID.

gaetano-guerriero (commented):
> *[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3376)* status: 8 of 21 files reviewed, 8 unresolved discussions (waiting on @vmarkovtsev)

___
*[`server/athenian/api/db.py` line 506 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnBDDy8PndR64KG0WF:-NOna83FEYpG7CTruHdW:b-896fix) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/db.py#L506)):*

We cannot do this by default. There are several known exclusions when automatic hints do not work at best and make the planner crazy at worst:
- unions
- pre-filtering, when VALUES is a minor filter
- negations
- the table is a view

Therefore, IMO, hints should always be written manually for each case, considering all the random factors. I am afraid that leaving this function will lead to erroneous hints that will complicate further optimizations - i.e., everyone will think that the query is already optimized. I see we've got one place where we are using this function, so we'd better remove it.
</blockquote></details>

Done.
___
*[`server/athenian/api/internal/miners/github/pull_request.py` line 3071 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnFobb4N395ESAzTgG:-NOnL8Hj9x6as2_nLIZd:b-896fix) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/internal/miners/github/pull_request.py#L3071)):*

To avoid silly bugs in the future
</blockquote></details>

Done.
___
*[`server/athenian/api/internal/miners/jira/issue.py` line 968 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnGL5h2YgPOsufBXNl:-NOnV-Zl2y3XyzDxGrig:b333q6a) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/internal/miners/jira/issue.py#L968)):*

It's obvious to me that we should hint this. Let's see, there is only one condition and we fetch mapped PRs. We can rely on at least 50% PRs mapped. So `Rows({nprji.__tablename__} *VALUES* {len(prs) // 2})`. However, we've got a JOIN, Postgres likes to move (and will move!) `*VALUES*` outside, leading to bad performance. So we must define `Leading((({nprji.__tablename__} *VALUES*) {Issue.__tablename__}))`

You should see now why automatic hinting is quite useless and even harmful.
</blockquote></details>

Cannot reproduce Postgres moving *VALUES* as a later step,
anyway adding the hint as requested

Also with this 
https://www.postgresql.org/docs/14/release-14.html#id-1.11.6.12.5.3.6
probably ANY VALUES optimization will be useless.
It should be for version 14 but I already get identical plan with IN ( / ANY VALUES ( in cloud sql
___
*[`server/athenian/api/internal/miners/jira/issue.py` line 1187 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnHZfw8L6Iv97JneU7:-NOnLSyB1LP0ElqHu1Qd:b-c1elv7) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/internal/miners/jira/issue.py#L1187)):*

One second. We always, always have the same account ID for everything. We don't need to group by account ID.
</blockquote></details>

This is 2020 code I just extracted the functionality to use it twice
https://github.com/athenianco/athenian-api/blob/master/server/athenian/api/internal/miners/github/pull_request.py#L1178.

Anyway should we optimize this now ? Maybe asserting just one jira account id ?
I guess it never surfaced while analyzing performances since it's easy to improve.
___
*[`server/athenian/api/internal/miners/jira/issue.py` line 1194 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnIRE33Crstsr1D2Ra:-NOnLlVb5VOnC2wVvPCJ:bfjpff7) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/internal/miners/jira/issue.py#L1194)):*

This is a terrible performance, we should never-ever execute OR in SQL except for very few exclusions. We should rewrite to a `UNION` of individual queries by each component condition.

I see we do `r[2].lower()`. Let's move it straight to SQL: `select(Component.name.lower())`.
</blockquote></details>

One clause for each jira account, so no OR in practice
About lower in SQL well that was obvious , I was expecting you didn't do it for some other reasons
___
*[`server/athenian/api/internal/miners/jira/issue.py` line 1201 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnJWDBCOMbKSZcTgQb:-NOna4Xr2tcfQKJzI95m:bm1f55q) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/internal/miners/jira/issue.py#L1201)):*

This is potentially very slow. Someone like account 186 is going to have 100k issues and we will bottleneck here.
Try to find a vectorized numpy way of doing the same thing.
</blockquote></details>

As I said this code was just moved
___
*[`server/athenian/api/models/__init__.py` line 151 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnJmGSDQWhvqIGLUVM:-NOna-u6DLiDnU3C32wm:bb6un1j) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/models/__init__.py#L151)):*

Suggestion: move 100 to the default value of `threshold` kwarg. For future fine-tuning,
</blockquote></details>

Done

vmarkovtsev (approved):
> Reviewed 1 of 10 files at r2, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3376)* status: 11 of 21 files reviewed, all discussions resolved

___
*[`server/athenian/api/internal/miners/jira/issue.py` line 968 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnGL5h2YgPOsufBXNl:-NOtHMqI6CuTktarxKJP:b-zh4k13) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/internal/miners/jira/issue.py#L968)):*
<details><summary><i>Previously, gaetano-guerriero wrote…</i></summary><blockquote>

Cannot reproduce Postgres moving *VALUES* as a later step,
anyway adding the hint as requested

Also with this 
https://www.postgresql.org/docs/14/release-14.html#id-1.11.6.12.5.3.6
probably ANY VALUES optimization will be useless.
It should be for version 14 but I already get identical plan with IN ( / ANY VALUES ( in cloud sql
</blockquote></details>

It's very cool that you've noticed this Gaetano! I've planned the upgrade to 14. Regarding the reproduction, yes, this is something that can only reproduce on real battle queries.
We have NOT IN queries, too, and they added the support for hashed NOT IN in 15.
There is only one advantage left then: VALUES gets parsed faster than IN.
___
*[`server/athenian/api/internal/miners/jira/issue.py` line 1187 at r3](https://reviewable.io/reviews/athenianco/athenian-api/3376#-NOnHZfw8L6Iv97JneU7:-NOtJaUBApbXT0X9PpaJ:b-4b7zre) ([raw file](https://github.com/athenianco/athenian-api/blob/8172aabfe38ea9987b8f7b95e46f8bf5ad9a61ed/server/athenian/api/internal/miners/jira/issue.py#L1187)):*
<details><summary><i>Previously, gaetano-guerriero wrote…</i></summary><blockquote>

This is 2020 code I just extracted the functionality to use it twice
https://github.com/athenianco/athenian-api/blob/master/server/athenian/api/internal/miners/github/pull_request.py#L1178.

Anyway should we optimize this now ? Maybe asserting just one jira account id ?
I guess it never surfaced while analyzing performances since it's easy to improve.
</blockquote></details>

Yep, I checked the code myself and realized I was the author haha. So you see, I coded the "normal" way in 2020. Lots of lessons learned since...

Indeed, it's better to leave for the next PR.

vmarkovtsev (approved):
> Reviewed 1 of 17 files at r1, 6 of 10 files at r2, 3 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3376)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gaetano-guerriero)

PR #3410, opened 2023-02-24
[DEV-6077] Set all PRs released by specific node ID as release ignored

vmarkovtsev:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3410)


vmarkovtsev:
@gaetano-guerriero note: this reworks the original logic of outlier first releases. Our problem was that we didn't precompute all the PRs in the release and pr facts (why? IDK) so we failed to hide some PRs. The new way is more reliable: we don't depend on the release facts and always take the current state of `done_pull_request_facts`.

PR #3430, opened 2023-03-01
[DEV-6072] fetch jira issues by keys

gaetano-guerriero:


server/athenian/api/internal/miners/jira/issue.py
```diff
@@ -435,15 +434,87 @@ async def _fetch_jira_issues(
...
+        Issue.title,
+        Issue.reporter_display_name,
+        Issue.assignee_display_name,
+        Issue.url,
+        Issue.priority_name,
+        Issue.status,
+        Issue.type_id,
+        Issue.project_id,
+        Issue.comments_count,
+        Issue.story_points,
```
vmarkovtsev, line 485: Since this is a low-level API, we should consider doing the same as in `_fetch_jira_issues`: have two column sets, the obligatory columns that we need to map to PRs, for example, and `extra_columns`.

server/athenian/api/internal/miners/jira/issue.py
```diff
@@ -571,14 +642,27 @@ async def _fetch_jira_issues(
     issues[ISSUE_PRS_RELEASED] = released
     issues[ISSUE_PRS_COUNT] = prs_count
     issues[ISSUE_PR_IDS] = issue_prs
-    issues[resolved_colname] = issues[resolved_colname].astype(issues[created_colname].dtype)
-    if (negative := issues[resolved_colname].values < issues[created_colname].values).any():
+    ISSUE_RESOLVED = AthenianIssue.resolved.name
+    ISSUE_CREATED = Issue.created.name
+
+    issues[ISSUE_RESOLVED] = issues[ISSUE_RESOLVED].astype(issues[ISSUE_CREATED].dtype)
```
vmarkovtsev, line 648: This is an old code, I think we no longer need the casts.

server/athenian/api/internal/miners/jira/issue.py
```diff
@@ -888,6 +960,34 @@ def hint_epics(q):
...
+    mdb: Database,
+    cache: Optional[aiomcache.Client],
+) -> pd.DataFrame:
+    where = [
+        Issue.acc_id == jira_config.acc_id,
+        Issue.project_id.in_(jira_config.projects),
+        Issue.is_deleted.is_(False),
+        Issue.key.progressive_in(keys),
+    ]
+    issue_status_join = sa.join(
```
vmarkovtsev, line 977: It's better to do outerjoin here. Otherwise, the planner can do smth nasty.

server/athenian/api/internal/miners/jira/issue.py
```diff
@@ -888,6 +960,34 @@ def hint_epics(q):
...
+        Issue,
+        Status,
+        sql.and_(Issue.status_id == Status.id, Issue.acc_id == Status.acc_id),
+    )
+    issue_athenian_issue_join = sa.outerjoin(
+        issue_status_join,
+        AthenianIssue,
+        sql.and_(Issue.acc_id == AthenianIssue.acc_id, Issue.id == AthenianIssue.id),
+    )
+    stmt = sa.select(*columns).select_from(issue_athenian_issue_join).where(*where)
```
vmarkovtsev, line 987: Since we do joins, the planner can place `*VALUES*` at the end of the plan, so we should hint with `Leading({Issue.__tablename__}, *VALUES*)`

server/athenian/api/internal/miners/jira/issue.py
vmarkovtsev: btw, is it to get the status category?

vmarkovtsev (changes requested):

server/athenian/api/internal/miners/jira/issue.py
gaetano-guerriero: It's needed by `_validate_and_clean_issues` which I was not calling, now I do

server/athenian/api/internal/miners/jira/issue.py
```diff
@@ -888,6 +950,36 @@ def hint_epics(q):
...
+    stmt = stmt.with_statement_hint(f"Leading({Issue.__tablename__}, *VALUES*)")
```
vmarkovtsev, line 978:
```suggestion
    if len(keys) > 100:
        stmt = stmt.with_statement_hint(f"Leading({Issue.__tablename__} *VALUES*)")
```
- condition
- removed the comma `,`

vmarkovtsev (changes requested):

vmarkovtsev (approved):

PR #3458, opened 2023-03-09
Avoid undefined call to __builtin_clzl

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3458)

### DIFF
--- a/server/athenian/api/internal/features/github/check_run_metrics_accelerated.pyx
+++ b/server/athenian/api/internal/features/github/check_run_metrics_accelerated.pyx
@@ -27,6 +27,12 @@ import numpy as np
 # __builtin_clzl is a compiler built-in that counts the number of leading zeros
 cdef extern int __builtin_clzl(unsigned long)
 
+cdef int _leading_zero_bits(unsigned long n):
+   if n == 0:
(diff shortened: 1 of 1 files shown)


vmarkovtsev (approved):

server/athenian/api/internal/features/github/check_run_metrics_accelerated.pyx
```diff
@@ -52,8 +52,13 @@ def calculate_interval_intersections(starts: np.ndarray,
     time_offset = 64 - __builtin_clzl(max_intervals - 1)
```
vmarkovtsev, line 52: Shall we do the same check here? I can easily image `max_intervals` to be equal to 1...

vmarkovtsev (changes requested):

server/athenian/api/internal/features/github/check_run_metrics_accelerated.pyx
gaetano-guerriero: right, done

vmarkovtsev (changes requested):

vmarkovtsev (approved):

PR #3473, opened 2023-03-17
[DEV-6215] Fix missing release_settings in filter_jira_stuff

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3473)

### DIFF
--- a/server/athenian/api/controllers/jira_controller/common.py
+++ b/server/athenian/api/controllers/jira_controller/common.py
@@ -57,18 +57,14 @@ class AccountInfo:
     account: int
     meta_ids: tuple[int, ...]
     jira_conf: JIRAConfig
-    branches: pd.DataFrame | None
-    default_branches: dict[str, str] | None
-    release_settings: ReleaseSettings | None
-    logical_settings: LogicalRepositorySettings | None
+    branches: pd.DataFrame
+    default_branches: dict[str, str]
+    release_settings: ReleaseSettings
+    logical_settings: LogicalRepositorySettings
     prefixer: Prefixer
 
 
-async def collect_account_info(
-    account: int,
-    request: AthenianWebRequest,
-    with_branches_and_settings: bool,
-) -> AccountInfo:
+async def collect_account_info(account: int, request: AthenianWebRequest) -> AccountInfo:
     """Collect the AccountInfo from the request."""
     sdb, mdb, pdb, cache = request.sdb, request.mdb, request.pdb, request.cache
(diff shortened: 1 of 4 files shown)


server/athenian/api/controllers/jira_controller/common.py
```diff
@@ -79,20 +75,17 @@ async def collect_account_info(
...
-        branches = release_settings = logical_settings = None
-        default_branches = {}
+
+    settings = Settings.from_request(request, account, prefixer)
+    (branches, default_branches), logical_settings = await gather(
+        BranchMiner.load_branches(
+            repos, prefixer, account, meta_ids, mdb, pdb, cache, strip=True,
+        ),
+        settings.list_logical_repositories(repos),
+        op="sdb/branches and releases",
```
vmarkovtsev, line 85:
```suggestion
        op="sdb/branches, logical settings",
```

server/athenian/api/controllers/jira_controller/common.py
```diff
@@ -79,20 +75,17 @@ async def collect_account_info(
...
+    )
+    repos = logical_settings.append_logical_prs(repos)
+    release_settings = await settings.list_release_matches(repos)
```
vmarkovtsev, line 88: We should call `settings.list_release_matches()` inside `gather()`, we don´t care about which repos to list

vmarkovtsev (changes requested):

server/athenian/api/controllers/jira_controller/common.py
gaetano-guerriero: should we also call `settings.list_logical_repositories` with no repos ?

server/athenian/api/controllers/jira_controller/common.py
```diff
@@ -79,20 +75,16 @@ async def collect_account_info(
...
-        branches = release_settings = logical_settings = None
-        default_branches = {}
+
+    settings = Settings.from_request(request, account, prefixer)
+    (branches, default_branches), logical_settings, release_settings = await gather(
+        BranchMiner.load_branches(
+            repos, prefixer, account, meta_ids, mdb, pdb, cache, strip=True,
+        ),
+        settings.list_logical_repositories(repos),
+        settings.list_release_matches(),
```
vmarkovtsev, line 85: @gaetano-guerriero I think that `settings.list_release_matches(repos)` will automatically include the logical repos. We can change the code, it seems

PR #3534, opened 2023-04-03
[DEV-6335] hint queries jira issues

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3534)

### DIFF
--- a/server/athenian/api/controllers/jira_controller/common.py
+++ b/server/athenian/api/controllers/jira_controller/common.py
@@ -222,13 +222,19 @@ async def fetch_issues_prs(
     log = logging.getLogger(f"{metadata.__package__}.fetch_issues_prs")
 
     async def fetch_prs_and_dependent_tasks():
-        prs_df = await read_sql_query(
+        prs_query = (
             sa.select(PullRequest)
             .where(
                 PullRequest.acc_id.in_(account_info.meta_ids),
-                PullRequest.node_id.in_(pr_ids),
+                PullRequest.node_id.progressive_in(pr_ids),
             )
-            .order_by(PullRequest.node_id.name),
+            .order_by(PullRequest.node_id.name)
+        )
+        if len(pr_ids) > 100:
+            # repo and pr are alias used in api_pull_requests view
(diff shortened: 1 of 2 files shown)


server/athenian/api/controllers/jira_controller/common.py
```diff
@@ -222,13 +222,19 @@ async def fetch_issues_prs(
...
                 PullRequest.acc_id.in_(account_info.meta_ids),
-                PullRequest.node_id.in_(pr_ids),
+                PullRequest.node_id.progressive_in(pr_ids),
             )
-            .order_by(PullRequest.node_id.name),
+            .order_by(PullRequest.node_id.name)
+        )
+        if len(pr_ids) > 100:
+            # repo and pr are alias used in api_pull_requests view
+            prs_query = prs_query.with_statement_hint("HashJoin(repo pr)")
```
vmarkovtsev, line 235: This is normally fixed by row count hints. In general, try to prefer row count hints versus direct join algorithm hints. Because, the problem is usually deeper than just a few hash joins.

vmarkovtsev: I changed this in #3561

server/athenian/api/hacks/show_logged_query.py
```diff
@@ -0,0 +1,78 @@
...
+
+def _get_query_body(uuid: str) -> str:
+    args = [
+        "gcloud",
+        "logging",
+        "read",
+        "--freshness=200d",
+        "--limit=1",
+        "--format=value(jsonPayload.msg)",
+        f'"{uuid}"',
```
vmarkovtsev, line 32: This is too weak and thus will cost us a lot of money. We should add the following filters:
```
resource.labels.container_name="api-server"
resource.labels.cluster_name="production-cluster"
```
plus `--freshness` should be reworked. We would point at an exact timestamp (we know it from Sentry) and set the window of ±5min.

vmarkovtsev: https://athenianco.atlassian.net/browse/DEV-6406

PR #3535, opened 2023-04-03
[DEV-6335] Add hint in _fill_issues_with_mapped_prs_info query

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3535)

### DIFF
--- a/server/athenian/api/internal/miners/jira/issue.py
+++ b/server/athenian/api/internal/miners/jira/issue.py
@@ -596,6 +596,16 @@ async def _fill_issues_with_mapped_prs_info(
         f"Leading({NodePullRequestJiraIssues.__tablename__} *VALUES* "
         f"{NodePullRequest.__tablename__})",
     )
+    if len(issues.index.values) > 300:
+        # with many issues sometimes our Leading hint confuses planner,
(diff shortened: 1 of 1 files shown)


server/athenian/api/internal/miners/jira/issue.py
```diff
@@ -596,6 +596,16 @@ async def _fill_issues_with_mapped_prs_info(
         f"Leading({NodePullRequestJiraIssues.__tablename__} *VALUES* "
         f"{NodePullRequest.__tablename__})",
     )
+    if len(issues.index.values) > 300:
+        # with many issues sometimes our Leading hint confuses planner,
+        # a sequential scan is done on NodePullRequestJiraIssues
+        # forcing Parallel (which would be probably done anyway with a good plan) avoids this
+        for hint in (
+            f"HashJoin({NodePullRequestJiraIssues.__tablename__} *VALUES)",
```
vmarkovtsev, line 604: Note: this can also happen if we've got more than one meta id. In that case, should use a UNION

gaetano-guerriero: We already use the union for multiple meta ids, or do you mean that we should always have those hints if we using a union ?

PR #3547, opened 2023-04-04
[DEV-6356] added table installation_progress

Ildyakov:


Ildyakov:
working with https://github.com/athenianco/cloud-common/pull/138
and https://github.com/athenianco/metadata/pull/1586

server/athenian/api/models/state/versions/5b3dc49a9d7b_installation_progress.py
```diff
@@ -0,0 +1,65 @@
...
+            sa.TIMESTAMP(timezone=True),
+        ),
+        sa.Column(
+            "precompute_completed",
+            sa.TIMESTAMP(timezone=True),
+        ),
+        sa.Column(
+            "current_status",
+            sa.Text(),
+        ),
```
lwsanty, line 59: can all these timestamps be null?

server/athenian/api/models/state/versions/5b3dc49a9d7b_installation_progress.py
```diff
@@ -0,0 +1,65 @@
...
+        "installation_progress",
+        sa.Column(
+            "account_id",
+            sa.Integer(),
+            sa.ForeignKey("accounts.id", name="fk_installation_progress_account"),
+            primary_key=True,
+        ),
+        sa.Column(
+            "account_created",
+            sa.TIMESTAMP(timezone=True),
```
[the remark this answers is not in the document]
Ildyakov: probably we don`t need an existing accounts at all. the point of this feature is just for tracking new installations

server/athenian/api/models/state/versions/5b3dc49a9d7b_installation_progress.py
Ildyakov: yep and they will be

server/athenian/api/models/state/models.py
```diff
@@ -595,7 +595,7 @@ class DashboardChart(create_time_mixin(created_at=True, updated_at=True), Base):
 
 
 class DashboardChartGroupBy(create_time_mixin(created_at=True, updated_at=True), Base):
-    """The group by configured for a dashboard chart."""
+    """Tracking installation progress for github accounts"""
```
dennwc, line 598: I think this comment is misplaced.

server/athenian/api/models/state/models.py
```diff
@@ -640,3 +640,48 @@ class DashboardChartGroupBy(create_time_mixin(created_at=True, updated_at=True),
...
+
+
+class InstallationProgress(create_time_mixin(created_at=False, updated_at=False), Base):
+    """The group by configured for a dashboard chart."""
+
+    __tablename__ = "installation_progress"
+
+    account_id = Column(
+        Integer(),
+        ForeignKey("accounts.id", name="fk_installation_progress_account"),
```
dennwc, line 652: This FK assumes that there must be an Athenian account associated with the record. This is not the case for GitHub installations: they usually start without Athenian account. So we won't be able to insert most of the records here, unless we force Athenian account creation from MD side.

server/athenian/api/models/state/versions/5b3dc49a9d7b_installation_progress.py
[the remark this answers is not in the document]
dennwc: Agree with Olek: we should create records for existing accounts. This is where the `current_status` will help - we can set it, but leave timestamps empty. Maybe we can update the table later to reflect actual installations that happened in the past.

dennwc (changes requested):

server/athenian/api/models/state/models.py
Ildyakov: ok removed

server/athenian/api/models/state/models.py
Ildyakov: yeah! a bit mistaken

server/athenian/api/models/state/models.py
dennwc: So what's your plan for the `account_id` column? Is it still an Athenian account ID or GitHub account ID?

Ildyakov: this have to be a Github ID as it is in the Event

server/athenian/api/models/state/models.py
```diff
@@ -640,3 +640,47 @@ class DashboardChartGroupBy(create_time_mixin(created_at=True, updated_at=True),
...
     DashboardChartGroupBy.jira_labels,
 )
+
+
+class InstallationProgress(create_time_mixin(created_at=False, updated_at=False), Base):
+    """Tracking installation progress for github accounts"""
+
+    __tablename__ = "installation_progress"
+
+    account_id = Column(
```
vmarkovtsev, line 650: Could you please format every field in one line for better readability and consistency with the surrounding models :pray: E.g.
```suggestion
    account_id = Column(Integer(), primary_key=True)
```

Also, note that `primary_key=True` implies `nullable=False` under the hood.

server/athenian/api/models/state/models.py
```diff
@@ -640,3 +640,47 @@ class DashboardChartGroupBy(create_time_mixin(created_at=True, updated_at=True),
...
+        Integer(),
+        nullable=False,
+        primary_key=True,
+    )
+    account_created = Column(
+        "account_created",
```
vmarkovtsev, line 656: We don't have to set the name for each column, luckily. This is enough:
```python
account_created = Column(TIMESTAMP(timezone=True))
```

server/athenian/api/models/state/versions/5b3dc49a9d7b_installation_progress.py
```diff
@@ -0,0 +1,63 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "installation_progress",
+        sa.Column(
+            "account_id",
+            sa.Integer(),
+            nullable=False,
```
vmarkovtsev, line 24:
```suggestion
```

vmarkovtsev (changes requested):

vmarkovtsev:
@Ildyakov merge conflicts

vmarkovtsev:
I think that there is a new clean PR: https://github.com/athenianco/athenian-api/pull/3564

PR #3564, opened 2023-04-11
[DEV-6356] installation progress

Ildyakov:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3564)

### DIFF
--- a/server/athenian/api/models/state/versions/5b3dc49a9d7b_installation_progress.py
+++ b/server/athenian/api/models/state/versions/5b3dc49a9d7b_installation_progress.py
@@ -22,7 +22,6 @@ def upgrade():
 
 
 def downgrade():
-
     with op.batch_alter_table("repository_sets") as bop:
         bop.drop_column("precompute_started")
         bop.drop_column("precompute_finished")


server/athenian/api/models/state/models.py
```diff
@@ -640,3 +640,19 @@ class DashboardChartGroupBy(create_time_mixin(created_at=True, updated_at=True),
...
+    __tablename__ = "installation_progress"
+
+    github_account_id = Column(Integer(), primary_key=True)
+    account_created = Column(TIMESTAMP(timezone=True))
+    fetch_started = Column(TIMESTAMP(timezone=True))
+    fetch_completed = Column(TIMESTAMP(timezone=True))
+    consistency_started = Column(TIMESTAMP(timezone=True))
+    consistency_completed = Column(TIMESTAMP(timezone=True))
+    precompute_started = Column(TIMESTAMP(timezone=True))
+    precompute_completed = Column(TIMESTAMP(timezone=True))
```
vmarkovtsev, line 657: This is  reposet and Athenian account specific. We have to move those timestamps to the `repository_sets` table.

server/athenian/api/models/state/models.py
```diff
@@ -640,3 +640,19 @@ class DashboardChartGroupBy(create_time_mixin(created_at=True, updated_at=True),
...
+    current_status = Column(Text())
```
vmarkovtsev, line 658: Let's delete this *for now*. I need us to create something absolutely minimal and ASAP. If we make the status, we will have to think what statuses can we have, when to update and what, consider different edge cases, etc.

server/athenian/api/models/state/models.py
```diff
@@ -640,3 +640,19 @@ class DashboardChartGroupBy(create_time_mixin(created_at=True, updated_at=True),
...
+
+
+class InstallationProgress(create_time_mixin(created_at=False, updated_at=False), Base):
+    """Tracking installation progress for github accounts"""
+
+    __tablename__ = "installation_progress"
+
+    github_account_id = Column(Integer(), primary_key=True)
+    account_created = Column(TIMESTAMP(timezone=True))
+    fetch_started = Column(TIMESTAMP(timezone=True))
```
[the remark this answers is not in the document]
Ildyakov: more so we don`t have a "account_created" event. yet.
and for a case of re-fetch they will be different.
so for now we can cut this field I think. yes

server/athenian/api/models/state/models.py
```diff
@@ -640,3 +640,19 @@ class DashboardChartGroupBy(create_time_mixin(created_at=True, updated_at=True),
     DashboardChartGroupBy.jira_issue_types,
     DashboardChartGroupBy.jira_labels,
 )
+
+
+class InstallationProgress(create_time_mixin(created_at=False, updated_at=False), Base):
```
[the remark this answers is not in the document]
Ildyakov: agreed

server/athenian/api/models/state/models.py
Ildyakov: so this force us to have some key in the state DB for github_account_ids.
create a table for comparing github_account_id and athenian_account_id needed.

server/athenian/api/models/state/models.py
Ildyakov: noone is forcing us to use all the columns of the table from the begining.
we can create a table "as is" and add the features to the pipeline and API when we will be ready

server/athenian/api/models/state/models.py
vmarkovtsev: You probably mean `account_github_accounts`?

server/athenian/api/models/state/models.py
vmarkovtsev: Let's migrate to additional fields in the future. To avoid seduction.

dennwc: Sorry, but how one text field is making it harder? We already have a PR to MD that uses it.

server/athenian/api/models/state/models.py
Ildyakov: if it is contains both this IDs - yes it is

server/athenian/api/models/state/models.py
Ildyakov: addition - PR to cloud-common is ready too. and if we cut the columns it will slow us to implement changes instead of getting some features faster.
and if we delete precompute timestamps from this table we will use all the rest immediately after releasing all the features that in process now

server/athenian/api/models/state/models.py
vmarkovtsev: Then yes it is

server/athenian/api/models/state/models.py
```diff
@@ -100,6 +100,8 @@ class RepositorySet(
     )
     tracking_re = Column(Text(), nullable=False, default=".*", server_default=".*")
     precomputed = Column(Boolean(), nullable=False, default=False, server_default="false")
+    precompute_started = Column(TIMESTAMP(timezone=True))
+    precompute_completed = Column(TIMESTAMP(timezone=True))
```
vmarkovtsev, line 104:
```suggestion
    precompute_finished = Column(TIMESTAMP(timezone=True))
```
Minor syntax nitpick: start-finish; begin-end; initiate-complete. I know that people don't usually care, but this is just good education and conformance to the surrounding table conventions :)

vmarkovtsev (changes requested):

server/athenian/api/models/state/models.py
vmarkovtsev: Oh, and I see we have the same syntax to fix in `installation_progress` :pray:

server/athenian/api/models/state/models.py
dennwc: Yes, I'd still like to have that field for the bot.

Agree about precomputer, we will need a second table then?

Ildyakov: precompute- fields dropped. account_created field dropped. others (including status) is used by actual written logic in metadata. the point is that status can have a voluntary (apart of some statuses that would be fixed) value and this can be used for some specific messages like "paused", "delayed" etc.

dennwc (changes requested):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3564)* status: all files reviewed, 6 unresolved discussions (waiting on @Ildyakov and @vmarkovtsev)

___
*[`server/athenian/api/models/state/versions/5b3dc49a9d7b_installation_progress.py` line 20 at r2](https://reviewable.io/reviews/athenianco/athenian-api/3564#-NTDiaUA7Pc5tI1Xpwoi:-NTDiaUA7Pc5tI1Xpwoj:b7b22oe) ([raw file](https://github.com/athenianco/athenian-api/blob/fadd50c0a78a2c0023a877f16d420e491fcc787f/server/athenian/api/models/state/versions/5b3dc49a9d7b_installation_progress.py#L20)):*
> ```Python
> def upgrade():
>     op.create_table(
>         "installation_progress",
> ```

Since we are removing API-related things from the table anyway, does it make sense to move it to MD repo completely? It can be a set of added columns for `github.account`, for example. This way we can add a data migration as well (set all accounts to "done" status).

Ildyakov:
> > Reviewed 2 of 2 files at r2, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3564)_ status: all files reviewed, 6 unresolved discussions (waiting on @Ildyakov and @vmarkovtsev)
> 
> _[`server/athenian/api/models/state/versions/5b3dc49a9d7b_installation_progress.py` line 20 at r2](https://reviewable.io/reviews/athenianco/athenian-api/3564#-NTDiaUA7Pc5tI1Xpwoi:-NTDiaUA7Pc5tI1Xpwoj:b7b22oe) ([raw file](https://github.com/athenianco/athenian-api/blob/fadd50c0a78a2c0023a877f16d420e491fcc787f/server/athenian/api/models/state/versions/5b3dc49a9d7b_installation_progress.py#L20)):_
> 
> > ```python
> > def upgrade():
> >     op.create_table(
> >         "installation_progress",
> > ```
> 
> Since we are removing API-related things from the table anyway, does it make sense to move it to MD repo completely? It can be a set of added columns for `github.account`, for example. This way we can add a data migration as well (set all accounts to "done" status).

this is interesting. yes the reason to use state db was precomputing.
I think this is a proper solution to move the draft to metadata. WDYT @vmarkovtsev?

server/athenian/api/models/state/models.py
Ildyakov: @dennwc we have a field "precomputed bool" in the table "repository_sets"
so it would be right to add a couple new fields to it. and provide a logic to fill this fields from the precomputer directly (not with event handler).

vmarkovtsev: It makes sense to write to metadata :+1:

PR #3566, opened 2023-04-13
[DEV-6381] Preserve jira priorities and issue types order in charts a…

gaetano-guerriero:
…nd goals

- - -
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3566)

### DIFF
--- a/server/athenian/api/internal/jira.py
+++ b/server/athenian/api/internal/jira.py
@@ -513,18 +513,18 @@ def normalize_user_type(type_: str) -> str:
     return type_
 
 
-def parse_request_priorities(req_priorities: Optional[list[str]]) -> Optional[list[str]]:
+def parse_request_priorities(req_priorities: list[str] | None) -> list[str] | None:
     """Parse the raw Jira priorities received in a request."""
     if req_priorities is None:
         return None
-    return sorted({normalize_priority(p) for p in req_priorities})
+    return list({normalize_priority(p): None for p in req_priorities})  # unique keeping order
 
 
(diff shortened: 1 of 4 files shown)


server/athenian/api/internal/jira.py
```diff
@@ -513,18 +513,20 @@ def normalize_user_type(type_: str) -> str:
...
 
 
-def parse_request_issue_types(req_issue_types: Optional[list[str]]) -> Optional[list[str]]:
+def parse_request_issue_types(req_issue_types: list[str] | None) -> list[str] | None:
     """Parse the raw Jira issue types received in a request."""
     if req_issue_types is None:
         return None
-    return sorted({normalize_issue_type(t) for t in req_issue_types})
+    normalized = [normalize_issue_type(t) for t in req_issue_types]
+    return list({t: None for t in normalized})
```
vmarkovtsev, line 529:
```suggestion
    return list({(normalize_issue_type(t) for t in req_issue_types})
```

gaetano-guerriero: set step will forget the req_issue_types order here

vmarkovtsev:
```python
return list({normalize_issue_type(t): None for t in req_issue_types})
```

vmarkovtsev: otherwise, we create a list, then a dict, then a list.

vmarkovtsev (approved):

PR #3569, opened 2023-04-13
[DEV-6407] jira issue acknowledge time field

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3569)


server/athenian/api/internal/miners/jira/issue.py
```diff
@@ -1344,6 +1344,28 @@ def resolve_resolved(
...
+    The acknowledge_time is `work_began` - `created`.
+    For the issues not yet started (so no work_began) it is `now()` - `created`.
+    `acknowledge_time` can never be less than 0.
+    """
+    res = np.empty(len(created), dtype="timedelta64[us]")
+    res = work_began - created
+    no_work_began = res != res
+    res[no_work_began] = now - created[no_work_began]
+    res[res < np.timedelta64(0, "us")] = np.timedelta64(0, "us")
+    return res
```
vmarkovtsev, line 1366: This duplicates `AcknowledgeTimeCalculator` and is actually not fully consistent with it. We'd rather try to refactor the calc by making a static subfunction.

server/athenian/api/controllers/jira_controller/__init__.py
```diff
@@ -458,6 +467,13 @@ async def _epic_flow(
                 epic.life_time = now - epic.created
                 if epic.work_began is not None:
                     epic.lead_time = now - epic.work_began
+
+            epic.acknowledge_time = resolve_acknowledge_time(
+                np.array([epic.created], dtype=np.datetime64),
+                np.array([epic.work_began], dtype=np.datetime64),
+                now,
+            )[0]
```
vmarkovtsev, line 475: Acc. 186 has 10k epics in 1 month and this will work too slow. We should try to vectorize over the df columns.

vmarkovtsev (changes requested):

server/athenian/api/internal/miners/jira/issue.py
gaetano-guerriero: They start from a different definition of "effective_work_began", so let's try first to  uniform this:

- for `jira-acknowledge-time` metric
`effective_work_began = fmin(work_began, prs_began) AND status == TODO`

- for `jira-lead-time` metric
`effective_work_began = fmin(work_began, prs_began)`

- for /jira/filter output
`effective_work_began = fmin(work_began, prs_began) AND work_began is not NaT`

I can use third definition for `jira-acknowledge-time` but that would change result of the metric when status is todo and work_began is filled.

vmarkovtsev: It is `status != TODO`: `acknowledged[status == TODO] = None` in the code. It is equivalent to checking `work_began == work_began` so the definitions are the same. Let's not touch the lead time though.

vmarkovtsev (approved):

PR #3576, opened 2023-04-17
[DEV-6383] deploy change failure metrics

gaetano-guerriero:
This change is [Reviewable](https://reviewable.io/reviews/athenianco/athenian-api/3576)


server/tests/controllers/metrics_controller/test_calc_metrics_deployments.py
```diff
@@ -395,3 +406,67 @@ async def test_deployment_metrics_empty_granularities(self, sample_deployments):
...
+                DeploymentMetricID.DEP_CHANGE_FAILURE_COUNT,
+                DeploymentMetricID.DEP_CHANGE_FAILURE_RATIO,
+            ],
+        )
+        DeplFact = DeploymentNotificationFactory
+        DeplCompFact = DeployedComponentFactory
+        DeplLabelFact = DeployedLabelFactory
+        await models_insert(
+            rdb,
+            DeplFact(name="d1", finished_at=dt(2005, 1, 10), started_at=dt(2005, 1, 1)),
```
vmarkovtsev, line 428: Can we please use either "Dep" or "Deployment" prefix. Not blocking, just for the future note. Two reasons:
- "l" sucks to read anyway
- We use "dep" in the main code, it is more widespread. This must be the same anti-logic as shortening "repository" to "repo" instead of "rep".

vmarkovtsev (approved):

